PR #9, opened 2017-09-08
composer: drop dependency on doctrine/* packages [closes #7]

### LINKED ISSUE #7: Hard dependency on doctrine/orm makes in impossible to use only with doctrine/common
_As the title says._

Maybe the package should be split to phpstan-doctrine-collections and phpstan-doctrine-orm?

JanTvrdik:


ondrejmirtes:
No, sorry, I feel like there need to be constraints in case something work differently in new (breaking) versions of Doctrine.

JanTvrdik:
I understand the problem it is trying to solve but it is extremely bad solution.

1) It is a lie. The library claims to require files it does not actually require and by design MUST NOT require (depending on the files would mean it's not doing static analysis). It would be better to use `conflict` (with inverse version range) instead of `require`.

2) It breaks the entire analysis when phpstan is installed correctly the way tools are supposed to be installed (i.e. not in project vendor directory, for example with usage of [composer-bin-plugin plugin](https://github.com/bamarni/composer-bin-plugin)), because it installs different version of Doctrine than you application is using and **uses the wrong version of doctrine for analysis**.

This is what phpstan installation looks on darujme.cz

image

Without manually writing the `replace` section and thus manually killing broken `require` from `phpstan/phpstan-doctrine` it would not work at all.

ondrejmirtes:
OK, I will take some time and try to reconsider this.

hkdobrev:
👍 on this. @JanTvrdik Perhaps you could resolve the conflicts and add the conflict section you've mentioned? Thank you!

hkdobrev:
@ondrejmirtes You could use the `conflicts` section to list ranges PHPStan could conflict with without requiring the actual Doctrine packages.

ondrejmirtes:
Superseded by #29.

PR #12, opened 2017-09-24
Add tests for existing extensions

vhenzl:
Unit tests for `DoctrineSelectableClassReflectionExtension`, `DoctrineSelectableDynamicReturnTypeExtension` and `EntityManagerFindDynamicReturnTypeExtension`.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,4 +1,6 @@
 parameters:
 	ignoreErrors:
 		- '#but returns PHPUnit_Framework_MockObject_MockObject#'
+		- '#PHPUnit_Framework_MockObject_MockObject given#'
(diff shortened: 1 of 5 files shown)


ondrejmirtes:
Thanks a lot, I really appreciate this! I merged this into master with some changes:

1) I had to update some tests because getReference/find/etc. now return `mixed` by default. It means that anything can be called on it and that PHPStan does not report "Cannot call method X() on object."
2) I like asserting type descriptions (strings) rather than asserting instances of Type objects. I consider them an implementation detail, but string type descriptions are considered public-facing.
3) I added a commit that integrates the new `phpstan-phpunit` extension which correctly handles mocks: https://github.com/phpstan/phpstan-doctrine/commit/6ad0f3d624b673087579b61dbf797850ad8f409c

PR #13, opened 2017-09-24
Update PHPUnit and its configuration

vhenzl:
PHPUnit 6.3 and configuration similar to the one in phpstan.


composer.json
```diff
@@ -19,7 +19,7 @@
 		"consistence/coding-standard": "~0.13.0",
 		"jakub-onderka/php-parallel-lint": "^0.9.2",
 		"phing/phing": "^2.16.0",
-		"phpunit/phpunit": "^5.7",
+		"phpunit/phpunit": "^6.3",
```
carusogabriel: We already are in PHPUnit 7, so this PR can be closed :)

PR #15, opened 2017-11-17
support ObjectManager interface

JanTvrdik:


ondrejmirtes:
`getPartialReference` and `getReference` are defined later in EntityManager, but I guess it does not matter here.

PR #16, opened 2017-12-01
Dynamic return type for EntityManager#merge

stesie:
Dynamically determine the return type of `EntityManager#merge`.

The merge function takes a single argument, the entity object  to re-attach with the EM. It passes back a managed copy of the entity, hence the return type is type of first argument.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -22,6 +22,10 @@ services:
 		tags:
 			- phpstan.broker.dynamicMethodReturnTypeExtension
 	-
+		class: PHPStan\Type\Doctrine\EntityManagerMergeDynamicReturnTypeExtension
+		tags:
+			- phpstan.broker.dynamicMethodReturnTypeExtension
+	-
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Hi, thank you for the patch, but this really cannot be in `EntityManagerGetRepositoryDynamicReturnTypeExtension.php`. Please create and register a separate extension :)

stesie:
oh my, ... I haven't noticed the `GetRepository` part in the class name, I've just looked for the `EntityManager` being selected in `getClass()` and looked no further.

Sorry for that, will follow up within shortly

stesie:
here we go :-)

ondrejmirtes:
I just realized that depending on ObjectManager would be better, see https://github.com/phpstan/phpstan-doctrine/pull/15 😊

stesie:
good point, since it's declared on the interface.

fixed :)

ondrejmirtes:
Thank you 😊

PR #29, opened 2018-05-28
Replaced some requirements with conflicts

lookyman:


hkdobrev:
🙌 Thanks! This would have been quite helpful a few weeks ago when we did `composer require --update-with-all-dependencies phpstan/phpstan-doctrine ^0.9` and it unintentionally updated Doctrine to 2.6 which resulted in some undocumented behaviour changes.

PR #17, opened 2017-12-04
Add support for getRepository() call on ObjectManager interface

pauci:
In our code we mostly typehint entity manager as `EntityManagerInterface` or even `ObjectManager` interface. This change should add support for them as well as for doctrine ODM's `DocumentManager`.

PR #34, opened 2018-08-06
Only override repository return type for entity managers

alcaeus:
The ObjectManager interface is shared across multiple projects, so assuming every object manager returns an entity repository is wrong and breaks using the extension with MongoDB ODM.

In addition to that, I would suggest renaming the generic extensions to `ObjectManager...` since they affect all object managers, not just an `EntityManager`. I wasn't sure if that would be considered a BC break so I left the refactoring out. Would you be open to renaming the classes?

### DIFF
--- a/src/Type/Doctrine/EntityManagerGetRepositoryDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/EntityManagerGetRepositoryDynamicReturnTypeExtension.php
@@ -23,7 +23,7 @@ class EntityManagerGetRepositoryDynamicReturnTypeExtension implements \PHPStan\T
 
 	public function getClass(): string
 	{
-		return 'Doctrine\Common\Persistence\ObjectManager';
+		return 'Doctrine\ORM\EntityManagerInterface';
 	}
 
 	public function isMethodSupported(MethodReflection $methodReflection): bool


ondrejmirtes:
Hi, the fact that this stops working is a bigger BC break than the extension renaming 😊

Anyway, this goes against #17. It would be better to send a PR that has ObjectManager returning ObjectRepository and EntityManagerInterface returning EntityRepository.

alcaeus:
> It would be better to send a PR that has ObjectManager returning ObjectRepository and EntityManagerInterface returning EntityRepository.

`ObjectManager` already returns `ObjectRepository`: https://github.com/doctrine/persistence/blob/af1ec238659a83e320f03e0e454e200f689b4b97/lib/Doctrine/Common/Persistence/ObjectManager.php#L108.

ondrejmirtes:
But I mean here, in the context of the extension, equivalent of this type: https://github.com/phpstan/phpstan-doctrine/blob/master/src/Type/Doctrine/EntityRepositoryType.php

alcaeus:
Forgive my ignorance, but why? Wouldn't PHPStan normally read it from the Docblock, so no additional type handling is necessary?

alcaeus:
Now I got it - I thought you wanted an equivalent of `EntityManagerGetRepositoryDynamicReturnTypeExtension` for `ObjectManager` - I'll create an `ObjectRepositoryDynamicReturnTypeExtension` to handle `ObjectRepository` return types 👍

ondrejmirtes:
You will need to create:

* ObjectRepositoryDynamicReturnTypeExtension
* ObjectManagerGetRepositoryDynamicReturnTypeExtension
* ObjectRepositoryType

PR #37, opened 2018-08-18
Add integration tests for ORM features

lcobucci:
@ondrejmirtes there's also a small fix which I can move to another PR if needed.

### DIFF
--- a/src/Type/Doctrine/EntityManagerFindDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/EntityManagerFindDynamicReturnTypeExtension.php
@@ -43,12 +43,7 @@ class EntityManagerFindDynamicReturnTypeExtension implements \PHPStan\Type\Dynam
 			return $mixedType;
 		}
 
-		$type = new ObjectType($argType->getValue());
-		if ($methodReflection->getName() === 'find') {
-			$type = TypeCombinator::addNull($type);
-		}
-
-		return $type;
+		return TypeCombinator::addNull(new ObjectType($argType->getValue()));
 	}
 
 }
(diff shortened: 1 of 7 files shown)


src/Type/Doctrine/EntityManagerFindDynamicReturnTypeExtension.php
```diff
@@ -43,12 +43,7 @@ public function getTypeFromMethodCall(
 			return $mixedType;
 		}
 
-		$type = new ObjectType($argType->getValue());
-		if ($methodReflection->getName() === 'find') {
-			$type = TypeCombinator::addNull($type);
-		}
-
-		return $type;
+		return TypeCombinator::addNull(new ObjectType($argType->getValue()));
```
alcaeus: This fix may be valid for ORM, but it's invalid for MongoDB ODM: `getReference` and `getPartialReference` don't return `null` in ODM. That said, it's best to further split this extension since both methods in question aren't part of the `ObjectManager` interface.

lookyman:
LGTM 😊

ondrejmirtes:
Hi, this is fine, going to merge this, but integration tests should also inspect some negative cases - actually verify that the extension doesn't just return `mixed` and thus turning off all checks, but that it actually returns the right thing. So there should be some negative checks generating error messages as well.

lcobucci:
@ondrejmirtes I think that we missid what @alcaeus said?

alcaeus:
Yep, not even as much as a reply to the highlighted issue. @ondrejmirtes the change to th extension is invalid, as is the original assumption that the ObjectManager interface provides a getReference method.

Issue #35, opened 2018-08-06
Call to an undefined method ... ObjectRepository::findOneByShortName().


tlorens:
Are magic methods not supported or is there some configuration option(s) I missed?

ondrejmirtes:
Nope, they are not supported right now.

tlorens:
Yeah, I found another post on it.. I'll just add it to the ignore list. .Thanks!

ankitpp:
Error - Call to an undefined method object::something();

In this particular instance method was called after Instantiating an object from child class. Child class extends parent class but error was still thrown. 

**Parent** 
abstract class Duhh 
{
   public function something(){
    *Definition of Function in parent*
   }
}


**Child**
class DuhhChild extends Duhh
{
  
}

**Use of Method In Some Controller that throws PHP Stan Error - Call to an undefined method object::something();**

use DuhhChild;
class SomeController extends random
{
  $something = new DuhhChild;
  $something->something()
}


inorder to work around this error simple define a docBlock that explains that variable as an Instance of Child Class above the variable that you are using to Instantiate the object. Example;

use DuhhChild;
class SomeController extends random
{
 /** @var DuhhChild $something */
  $something = new DuhhChild;
  $something->something()
}

ondrejmirtes:
@ankitpp This is completely unrelated to this issue and even to phpstan-doctrine, please open a new one in https://github.com/phpstan/phpstan/issues/new/choose

PR #41, opened 2018-09-26
Add support for magic find* entity repository methods #35

jackbentley:


keelerm84:
I believe this PR would also fix some issues I'm having. If the tests were passing, would this likely get merged?

jackbentley:
Thanks for the comment 😄 I've fixed the tests - was just CS issues.

ondrejmirtes:
Hi, this PR does not take full advantage of static analyser - ideally, it should know which exact `findBy` methods exist. It should be possible once the entity metadata are available from other PRs.

jackbentley:
I agree, @ondrejmirtes. I wanted to implement that behaviour but wasn't able to easily achieve it. Hopefully, that's something that can be improved when the API becomes available?

ondrejmirtes:
Hi, this will need to be reworked in light of #49 - it will allow you to find out which `findBy` methods are really there based on the entity :)

ondrejmirtes:
Thanks! I'll implement the checks as a rule.

ondrejmirtes:
Implemented the rest: https://github.com/phpstan/phpstan-doctrine/commit/5bf58a32e0c056ee9abaca114e6109ee4ed0aa79

PR #46, opened 2019-02-12
ManagerRegistry::getRepository returns correct EntityRepository

mcfedr:
Similar to how ObjectManager:: getRepository works, but for ManagerRegistry

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -32,3 +32,9 @@ services:
 		class: PHPStan\Type\Doctrine\ObjectRepositoryDynamicReturnTypeExtension
 		tags:
 			- phpstan.broker.dynamicMethodReturnTypeExtension
+	-
+		class: PHPStan\Type\Doctrine\ManagerRegistryGetRepositoryDynamicReturnTypeExtension
+		arguments:
+			repositoryClass: %doctrine.repositoryClass%
+		tags:
+			- phpstan.broker.dynamicMethodReturnTypeExtension
--- a/src/Type/Doctrine/GetRepositoryDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/GetRepositoryDynamicReturnTypeExtension.php
@@ -1,0 +1,48 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Doctrine;
+
+use PhpParser\Node\Expr\MethodCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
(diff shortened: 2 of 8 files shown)


ondrejmirtes:
Hi, please also add some error on purpose to the example file so that it can be seen it’s the right type detected. It will generate a JSON file you also need to commit. Thanks.

tests/DoctrineIntegration/Persistence/data/managerRegistryDynamicReturn.php
```diff
@@ -0,0 +1,43 @@
...
+    private $managerRegistry;
+
+    public function __construct(ManagerRegistry $managerRegistry)
+    {
+        $this->managerRegistry = $managerRegistry;
+    }
+
+    public function findDynamicType(): void
+    {
+        $test = $this->managerRegistry->getRepository(MyEntity::class)->createQueryBuilder('e');
```
mcfedr: I think its a bit more obvious now, `createQueryBuilder` is only on `EntityRepository`, not `ObjectRepository` that would normally be the type

ondrejmirtes: This isn't what I meant. I didn't want `phpstan-negative.neon` that doesn't have the extension. Rather, I wanted to verify that the extension works. So I'd like to have only a single integration "levels" test that verifies that `ManagerRegistry::getRepository()` actually returns EntityRepository, by calling something nonexistent on the EntityRepository object.

So besides verifying that "Call to an undefined method Doctrine\\Common\\Persistence\\ObjectRepository::createQueryBuilder()." does NOT GET reported, I'd also like to know that `$this->managerRegistry->getRepository(MyEntity::class)->nonexistent()` GETS reported.

mcfedr: Ok, I think I see what you mean, so you want to prove its an entity manager because the error will say its an entity manager, and they dont have function x.

One moment :)

mcfedr: the negative option didnt seem great, but i didnt understand what you where asking...

ondrejmirtes: I want to see that `$this->managerRegistry->getRepository(MyEntity::class)` returns the right thing and not just `mixed` (that silents all errors).

mcfedr: Its done now.

ondrejmirtes:
Thanks, perfect!

PR #39, opened 2018-08-18
Add integration tests for ODM features

lcobucci:
Which kicks-off the support for #26 

Depends on #38

PR #47, opened 2019-02-12
Resolve custom repository classes

mcfedr:
I've taken #18 and fixed it up and rebased on master

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,13 @@
 	"require": {
 		"php": "~7.1",
 		"phpstan/phpstan": "^0.11",
-		"nikic/php-parser": "^4.0"
+		"nikic/php-parser": "^4.0",
+		"doctrine/orm": "^2.5",
+		"doctrine/common": "^2.7",
+		"doctrine/annotations": "^1.5",
+		"doctrine/persistence": "^1.1",
+		"doctrine/dbal": "^2.5",
+		"doctrine/event-manager": "^1.0"
 	},
 	"require-dev": {
 		"consistence/coding-standard": "^3.0.1",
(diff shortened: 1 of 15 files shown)


composer.json
```diff
@@ -12,7 +12,13 @@
...
 		"php": "~7.1",
 		"phpstan/phpstan": "^0.11",
-		"nikic/php-parser": "^4.0"
+		"nikic/php-parser": "^4.0",
+		"doctrine/orm": "^2.5",
+		"doctrine/common": "^2.7",
+		"doctrine/annotations": "^1.5",
+		"doctrine/persistence": "^1.1",
+		"doctrine/dbal": "^2.5",
+		"doctrine/event-manager": "^1.0"
```
mcfedr: I'm quite unconvinced of the need for this - composer-require-checker demands this, but actually if you dont configure 'mapping' you wont need any of this, and if you do, you should have it installed anyway

mcfedr:
Will make sense to rebase this on top of #46 because the code moves into the abstract class

ondrejmirtes:
Hi, the work done by @lcobucci in #40 seems like a more solid base to work on.

I really don't like how `src/DoctrineClassMetadataProvider.php` constructor looks like - it violates the principle "don't do any work in constructor", for example people without the `pdo_sqlite` extension would be out of luck.

composer.json
ondrejmirtes: I don't understand why any of the changes in composer.json were needed.

ondrejmirtes:
Unfortunately there's some unresolved discussion in #39 and #40 is based on that.

mcfedr:
Makes sense, #48 fixes the issues in #39.

If you merge that ill see about fixing up #40 as well.

composer.json
mcfedr: Its because of the code that loads/creates the entity manager, but if we use the approach in #40 this wont be needed.

PR #48, opened 2019-02-12
Add integration tests for ODM features

mcfedr:
Same as #39 but without extra composer deps

Had to leave 'config.mongodb' in composer json so that it would allow installing doctrine-odm

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,12 +25,15 @@
 		"slevomat/coding-standard": "^4.5.2",
 		"doctrine/common": "^2.7",
 		"doctrine/orm": "^2.5",
-		"doctrine/collections": "^1.0"
+		"doctrine/collections": "^1.0",
+		"doctrine/mongodb-odm": "^1.2",
+		"alcaeus/mongo-php-adapter": "^1.1"
 	},
 	"conflict": {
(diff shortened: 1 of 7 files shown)


mcfedr:
The travis job has failed because of some travis issue, but you can see in the php 7.1 job that it passes.

ondrejmirtes:
Please rebase on top of master, it conflicts now.

composer.json
```diff
@@ -39,5 +41,10 @@
 	},
 	"autoload-dev": {
 		"classmap": ["tests/"]
+	},
+	"config": {
+		"platform": {
+			"ext-mongo": "1.6.16"
```
ondrejmirtes: What's this for?

mcfedr: `doctrine/mongodb-odm` requires `ext-mongo`, this tells composer that its installed, even though it isnt, and allows odm to be installed

mcfedr: The other option is to run `composer install` with `--ignore-platform-reqs`, if you think thats better @ondrejmirtes ?

ondrejmirtes: We can leave this in :)

ondrejmirtes:
Thanks!

PR #56, opened 2019-02-20
Do not error if a non-field findByX method exists

### LINKED ISSUE #52: issue with findOneByX
In some repositories I have some custom functions named like `findOneByX`

Example:

```php
class ReviewRepository
{
    public function findOneByReviewConfigAndId(...) { }
}
```

Seems that is not working properly:

```
 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   someClass.php                                                                                                                                             
 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  209    Call to method X\ReviewRepository<X\Entity\Review>::findOneByReviewConfigAndId() - entity X\Entity\Review does not have a field named  
         $reviewConfigAndId.                                                                                                                                                                         
 ------ --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

roverwolf:
Do not generate an error if a custom findByX (or findOneByX, countByX) method exists but X is not a field in the entity class.

This should fix #52

### DIFF
--- a/src/Rules/Doctrine/ORM/MagicRepositoryMethodCallRule.php
+++ b/src/Rules/Doctrine/ORM/MagicRepositoryMethodCallRule.php
@@ -78,6 +78,11 @@ class MagicRepositoryMethodCallRule implements Rule
 			return [];
 		}
 
+		$repositoryReflectionClass = new \ReflectionClass($calledOnType->getClassName());
+		if ($repositoryReflectionClass->hasMethod($methodName)) {
+			return [];
+		}
+
 		return [sprintf(
 			'Call to method %s::%s() - entity %s does not have a field named $%s.',
 			$calledOnType->describe(VerbosityLevel::typeOnly()),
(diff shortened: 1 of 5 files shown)


ondrejmirtes:
Hi, this isn't sufficient.

1) There's no test that reproduces the problem.
2) PHPStan's reflection objects should be used - you can obtain them by injecting `Broker` into the rule and calling `$this->broker->getClass()` and then ask for `hasNativeMethod`.
3) There are multiple places that need to be fixed (and all of those bugs need to be reproduced first with tests):

* https://github.com/phpstan/phpstan-doctrine/blob/master/src/Type/Doctrine/ObjectRepositoryDynamicReturnTypeExtension.php
* https://github.com/phpstan/phpstan-doctrine/blob/master/src/Rules/Doctrine/ORM/MagicRepositoryMethodCallRule.php

roverwolf:
I've added a test and changed to use PHPStan's Broker (sorry, I am new to PHPStan).

I have not seen any errors from PHPStan due to `ObjectRepositoryDynamicReturnTypeExtension`.  Does it need to change if we assume the rule stays the same where a "find" method should return the entity class (or null, or an array of them)?

ondrejmirtes:
I'll look at `ObjectRepositoryDynamicReturnTypeExtension` myself. Thank you!

PR #68, opened 2019-07-16
add php 7.4snapshot

andreybolonin:


ondrejmirtes:
Hi, PHP 7.4 isn't released yet and the build failed. Please make first sure that the build succeeds and then submit a PR.

PR #81, opened 2019-10-10
Tests: use PHPStan binary directly

mhujer:
On Windows `vendor/bin/phpstan` contains some code for Cygwin, which is not understood when ran directly with PHP binary


ondrejmirtes:
Makes sense, thank you :)

PR #84, opened 2019-10-15
Doctrine ORM entities cannot be final or contain final methods

mhujer:
See https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/architecture.html#entities

- Depends on https://github.com/phpstan/phpstan-doctrine/pull/83 which should be merged first
- I'm not sure about the rules naming convention - should the rule name contain "must not" or "cannot" or nothing?


ondrejmirtes:
Superseded by https://github.com/phpstan/phpstan-doctrine/pull/143 :)

PR #90, opened 2019-11-30
Build also on Windows

mhujer:


mhujer:
Surprisingly, even the setup from phpstan/phpstan-src does not work (and it fails there as well)

ondrejmirtes:
I hope it's just a temporary downtime ((503) Server Unavailable. Service Unavailable) and it will resolve itself.

mhujer:
It seems [they are having trouble at Chocolatey](https://status.chocolatey.org/), I will update the PR when they are resolved.

image

mhujer:
It seems to be working fine now.

ondrejmirtes:
Thank you 😊

PR #92, opened 2019-11-30
Ramsey UUID TypeDescriptor

mhujer:


src/Type/Doctrine/Descriptors/Ramsey/UuidBinaryOrderedTimeTypeDescriptor.php
```diff
@@ -0,0 +1,32 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Doctrine\Descriptors\Ramsey;
```
mhujer: What about putting third-party descriptors into respective subnamespaces?

src/Type/Doctrine/Descriptors/Ramsey/UuidBinaryOrderedTimeTypeDescriptor.php
```diff
@@ -0,0 +1,32 @@
...
+use PHPStan\Type\Doctrine\Descriptors\DoctrineTypeDescriptor;
+use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use Ramsey\Uuid\Doctrine\UuidBinaryOrderedTimeType;
+use Ramsey\Uuid\UuidInterface;
+
+class UuidBinaryOrderedTimeTypeDescriptor implements DoctrineTypeDescriptor
+{
+
+	public function getType(): string
```
mhujer: All classes are identical except for the `getType()` method

ondrejmirtes: Instead, inject the type into the constructor and register the same class as three different services :)

Also a test case would be nice, you can add one of the descriptors to EntityColumnRuleTest and add an entity field with UUID type there :)

lookyman:
I was thinking about creating a separate package for this. But considering the popularity of Ramsey's extension, I guess we can have this in here?

ondrejmirtes:
It doesn’t have any dependency nor downside, I want it here 😊

ondrejmirtes:
Thank you!

PR #96, opened 2019-12-17
Add support for php 7.4

snapshotpl:


ondrejmirtes:
PHP 7.4 is supported, but it's not that easy to get it to work in the build, definitely not a one-line change.

PR #100, opened 2019-12-21
README: ReflectionDescriptor should use 'factory' instead of 'class'

mhujer:
Otherwise there is a deprecation reported: Service '0': option 'class' should be changed to 'factory'

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,6 @@ services:
 
 	# in case you are using the ReflectionDescriptor
 	-
-		class: PHPStan\Type\Doctrine\Descriptors\ReflectionDescriptor('MyApp\MyCustomTypeName')
+		factory: PHPStan\Type\Doctrine\Descriptors\ReflectionDescriptor('MyApp\MyCustomTypeName')
 		tags: [phpstan.doctrine.typeDescriptor]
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
How is the deprecation reported?

mhujer:
```
> phpstan analyse -c phpstan.neon src public --level 8 --no-progress
PHP Deprecated:  Service '0': option 'class' should be changed to 'factory'. in phar://C:/dev/htdocs/foobar.cz/vendor/phpstan/phpstan/phpstan/vendor/nette/di/src/DI/Config/DefinitionSchema.php on line 81
Deprecated: Service '0': option 'class' should be changed to 'factory'. in phar://C:/dev/htdocs/foobar.cz/vendor/phpstan/phpstan/phpstan/vendor/nette/di/src/DI/Config/DefinitionSchema.php on line 81

... actual PHPStan report follows

```

ondrejmirtes:
Oh yeah, I've never actually seen this in real world but heard about it. Generally `class` isn't probably deprecated, only this specific usage.

ondrejmirtes:
Thank you :)

PR #104, opened 2020-01-10
GetRepositoryDynamicReturnTypeExtension: improved return type when argument is not recognized

pepakriz:
Previously this bug hasn't been detected:

```php
	/**
	 * @param class-string $entityName
	 */
	public function doSomethingWithEntity(string $entityName): void
	{
		$repository = $this->entityManager->getRepository($entityName);
		$repository->createQueryBuilder('a')->unknownMethod(); // bug
```

### DIFF
--- a/src/Type/Doctrine/GetRepositoryDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/GetRepositoryDynamicReturnTypeExtension.php
@@ -9,7 +9,6 @@ use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\Generic\GenericClassStringType;
 use PHPStan\Type\Generic\GenericObjectType;
-use PHPStan\Type\MixedType;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\Type;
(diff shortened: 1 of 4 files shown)


pepakriz:
@ondrejmirtes 
1) `Doctrine\Persistence\ObjectRepository<T>`
2) Ok, done

ondrejmirtes:
Unresolved `T` in this context is wrong - `ObjectRepository<T>`. You're passing a `class-string` so you should get `ObjectRepository<mixed>` or `ObjectRepository<object>` at best..

ondrejmirtes:
I believe you so for now I reverted the change released it as 0.12.9. The stubs should work for everyone again.

You'll be able to install dev-master and reproduce this problem so we can return to simplified and thinner stub files again when we fix this problem.

ondrejmirtes:
Oh, wrong thread.

pepakriz:
@ondrejmirtes ok, check the second commit.

ondrejmirtes:
Thank you! I modified the solution in your second commit to let PHPStan do the logic: https://github.com/phpstan/phpstan-doctrine/commit/5762a215be4a06206ee4cd62df24da3196ebdcc9

ondrejmirtes:
I'll be able to release this once #101 is solved.

PR #108, opened 2020-01-27
Support ORM's embeddables

malarzm:
Today at work I've realized there's currently no rules for ORM's embeddables so here's an attempt to change that :) I'll remove WIP once I'm done with my goals, for now I'm making a PR for early feedback (or to learn that somebody else is already working on this).

Goals:
- [x] Check if property's type agrees with ORM's knowledge
- [x] Check columns specified in embeddables just like for entities

### DIFF
--- a/src/Rules/Doctrine/ORM/EntityColumnRule.php
+++ b/src/Rules/Doctrine/ORM/EntityColumnRule.php
@@ -64,9 +64,6 @@ class EntityColumnRule implements Rule
 		}
 
 		$className = $class->getName();
-		if ($objectManager->getMetadataFactory()->isTransient($className)) {
-			return [];
-		}
 
 		try {
 			$metadata = $objectManager->getClassMetadata($className);
--- a/src/Rules/Doctrine/ORM/EntityEmbeddableRule.php
+++ b/src/Rules/Doctrine/ORM/EntityEmbeddableRule.php
@@ -1,0 +1,87 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Doctrine\ORM;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MissingPropertyFromReflectionException;
+use PHPStan\Rules\Rule;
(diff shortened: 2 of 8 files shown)


src/Rules/Doctrine/ORM/EntityColumnRule.php
```diff
@@ -86,6 +90,10 @@ public function processNode(Node $node, Scope $scope): array
 			return [];
 		}
 
+		if (isset($metadata->embeddedClasses[$propertyName])) {
+			return $this->checkEmbeddedObject($class, $property, $metadata, $propertyName);
```
malarzm: Shall I make a separate rule for this check?

src/Rules/Doctrine/ORM/EntityColumnRule.php
```diff
@@ -64,9 +68,6 @@ public function processNode(Node $node, Scope $scope): array
 		}
 
 		$className = $class->getName();
-		if ($objectManager->getMetadataFactory()->isTransient($className)) {
```
malarzm: I'm trying to figure out internally why the embeddables are considered transient, but removal of this check causes no test failures

malarzm: Also classes not managed by Doctrine will have an exception thrown a bit later while trying to get their metadata and will return early from the rule with no errors, so I believe the intent is preserved

malarzm: I've created an issue in ORM but I think we can continue without waiting for its resolution https://github.com/doctrine/orm/pull/8006

malarzm:
PR is now ready :)

malarzm:
Anybody? @lookyman @ondrejmirtes ?

ondrejmirtes:
Hi, sorry for keeping you waiting, I'll look into this when I have the time.

enumag:
Any news on this one? It would be helpful imo. :-)

kissifrot:
Looks like this PR has conflicts, but it would be helpful to have it merged :)

Issue #112, opened 2020-02-14
False positive on getRepository(static::class) in abstract class.


noemi-salaun:
- **phpstan/phpstan v0.12.10**
- **phpstan/phpstan-doctrine v0.12.9**

In our project we have a base abstract class extended by every entities that provide usefull static shortcut.

```php
abstract class BaseEntity {
    public static function getRepository(): EntityRepository
    {
        return DoctrineSingleton::getEntityManager()->getRepository(static::class);
    }

    // ...many others static and non-static methods.
}
```

It allows us to use it like :

```php
$posts = Post::getRepository()->findBy(['...']);
```

_DISCLAIMER : Our project use singleton before switching to dependency injection but many part of the project still use the old singleton._ 

So my issue is :

> The class 'BaseEntity' was not found in the chain configured namespaces...

And this kind of error cannot be ignored.

So I tried to add this class to a fake namespace in my `phpstan/object-manager.php`.

It works but I got a new error :

>  Class "BaseEntity" is not a valid entity or mapped super class

It's true, but it's ok as `BaseEntity` is an abstract class and `static::class` will never be resolve to `BaseEntity` class. And this error cannot be ignored either.

> Error message ... cannot be ignored, use excludes_analyse instead.

But I don't wan't to exclude all the file from inspection just for one line.

noemi-salaun:
I don't know what the solution can be. I haven't look at the implementation of these rules yet. Maybe try-catching the process so we can ignore these error with abstract class ? Because in some case we want to have this rule activated, if an abstract class is a mapped super class or the base entity of some inheritance schema...

I'll look inside the implementation tomorrow, so I can post more pertinents proposals and eventually make a PR.

GKFX:
This error also appears with generics set up like the following:
```php
/***
 * @phpstan-template T of BaseEntity
 * @phpstan-param class-string<T> $repository_name
 */
public function fn($repository_name)
{
    $repo = $this->em->getRepository($repository_name);
}
```
The result here should be that `$repo` is an ordinary entity repository returning objects of types inheriting from BaseEntity. Instead the "is not a valid entity or mapped super class." message is seen.

ondrejmirtes:
Well the BaseEntity should be a mapped superclass, otherwise you’re risking someone will extend it without making the child class an entity and you’ll get the same error at runtime.

GKFX:
[...]

PR #113, opened 2020-02-15
ObjectMetadataResolver ignore error on invalid entity for abstract class

noemi-salaun:
I go for the simplest solution: try-catching Doctrine mapping error if the class is abstract.

Fix issue #112


ondrejmirtes:
I believe this is already solved on master.

PR #120, opened 2020-04-18
Add Stub for DBAL\QueryBuilder

Khartir:
This adds a stub for `Doctrine\DBAL\QueryBuilder`.
The `QueryBuilder` has two possible states:

1. By default, or as the return type of `select()` or `addSelect()` the `execute()` method returns an instance of `Doctrine\DBAL\Statement`.
2. The instance returned from calling `insert()`, `update()` or `delete()` will return `int` on calling `execute()`.

I did not manage to set the default (without calling `select()` or `addSelect()`) to return a `Statement`. I don't know if this is possible with stubs. If this is only possible by writing a full type definition, I can look into it. On the other hand, calling `execute` without calling one of the other methods first seems like a very rare usecase, so maybe this isn't even necessary.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -22,6 +22,8 @@ parameters:
 		- stubs/ObjectManager.stub
 		- stubs/ObjectManagerDecorator.stub
 		- stubs/ObjectRepository.stub
+		- stubs/QueryBuilder.stub
+		- stubs/Statement.stub
 		- stubs/Persistence/ManagerRegistry.stub
 		- stubs/Persistence/ObjectManager.stub
 		- stubs/Persistence/ObjectManagerDecorator.stub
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Hi, I don't think that this is the right reason to add generics to `DBAL\QueryBuilder`, it wouldn't look nice in the code. We could solve `execute()` return type other way, like `QueryBuilder::getDQL()` is solved.

Feel free to poke around here https://github.com/phpstan/phpstan-doctrine/tree/master/src/Type/Doctrine/QueryBuilder and here https://github.com/phpstan/phpstan-doctrine/tree/master/src/Type/Doctrine/Query.

If you'd still like to have your QueryBuilder to be generic this way, you can include this stub in your own project. Thanks.

PR #123, opened 2020-04-24
Add RepositoryFactory stub

jeroennoten:
This change improves type checking when using a custom `RepositoryFactory`.


ondrejmirtes:
Thank you!

Issue #67, opened 2019-07-10
Support for Repositories as a Services (Symfony)


Kocal:
Hi, this feature request follows https://github.com/phpstan/phpstan-doctrine/issues/65#issuecomment-509661033 and https://github.com/phpstan/phpstan-doctrine/issues/65#issuecomment-509799147.

Since Symfony 4.0, it is possible to define [Repositories as as Service](https://symfonycasts.com/screencast/symfony4-upgrade/extras#repositories-as-a-service) by extending from `ServiceEntityRepository` like this:

```php
<?php

namespace App\Repository\User;

use App\Entity\User;
use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
use Symfony\Bridge\Doctrine\RegistryInterface;

/**
 * @method User|null find($id, $lockMode = null, $lockVersion = null)
 * @method User|null findOneBy(array $criteria, array $orderBy = null)
 * @method User[]    findAll()
 * @method User[]    findBy(array $criteria, array $orderBy = null, $limit = null, $offset = null)
 */
class UserRepository extends ServiceEntityRepository
{
    public function __construct(RegistryInterface $registry)
    {
        parent::__construct($registry, User::class);
    }
}
```

And we can accessing it directly from services constructor or constroller's actions with type hinting:

```php
<?php

namespace App\Controller;

use App\Repository\User\UserRepository;
use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
use Symfony\Component\Routing\Annotation\Route;

class HomeController extends AbstractController
{
    /**
     * @Route("/", name="home")
     */
    public function index(UserRepository $userRepository)
    {
        $user = $userRepository->find(123);
        $user->getFooBar();
    }
}
``` 

But there are some issues.

If we use this phpstan extension, calling whatever method on `$user` that does not exist on `App\Entity\User` will not fails.

If we don't use it, `$userRepository->find*` methods will works correctly (due to `@method` annotations, but they are needed for PHPStorm too) and we will see the error `Cannot call method getFooBar() on App\Entity\User|null.`

As said in the issue, we can use `/** @var App\Repository\UserRepository<App\Entity\User> */` to make phpstan works, but:
- this is verbose, we have to add it everytime we inject a repository as service dependency
- this is non sense, since the type is `UserRepository`, phpstan should be able to works with it

To make it work, this extension (or the Symfony extension?) should:
- Find classes that extends from `ServiceEntityRepository`
[...]

PR #129, opened 2020-05-29
Adds stub for ServiceEntityRepository

gianiaz:
This stub was missing, hope it helps :-)


extension.neon
```diff
@@ -22,6 +22,8 @@ parameters:
 		- stubs/ObjectManager.stub
 		- stubs/ObjectManagerDecorator.stub
 		- stubs/ObjectRepository.stub
+		- stubs/ObjectRepository.stub
```
ondrejmirtes, line 25: This stub no longer exists, probably wrong rebase.

gianiaz: Sorry, I think I've fixed

ondrejmirtes:
Thank you!

Tobion:
I assume this change makes phpstan complain

> Class Acme\MyRepository extends generic class Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository but does not specify its types: TEntityClass

@ondrejmirtes @gianiaz what annotation do I need to put on MyRepository to make phpstan know the "TEntityClass"?
I tried `@implements ServiceEntityRepository<MyEntity>` but that wasn't it.

Tobion:
Somewhat related to #67

ondrejmirtes:
You’re extending a class so you need `@extends`. Here’s the docs: https://phpstan.org/blog/generics-in-php-using-phpdocs

Tobion:
Thanks @ondrejmirtes

Kocal:
Following this PR and #67, is there a way to automatically define `TEntityClass` with the 2nd argument of `ServiceEntityRepository::__construct()` through a PHPStan extension?

This way users won't have to `@extends ServiceEntityRepository<MyClass>` manually.

Thanks!

ondrejmirtes:
@Kocal It's not feasible to do something like that in an extension, having `@extends` annotation is required. Already answered here: https://github.com/phpstan/phpstan-doctrine/issues/178

Kocal:
Alright, sorry for not seeing the response before.
Thanks!

PR #139, opened 2020-08-04
Fix empty DQL with QueryBuilder::setParameters()

julienfalque:
Since version 0.12.18 I have a weird error:

```
QueryBuilder: [Syntax Error] line 0, col -1: Error: Expected IdentificationVariable | ScalarExpression | AggregateExpression | FunctionDeclaration | PartialObjectExpression | "(" Subselect ")" | CaseExpression, got end of string.
DQL: SELECT
```

It is reported in code that uses the `QueryBuilder::setParameters()` method. I first replaced those calls with several `QueryBuilder::setParameter()` calls and the errors were not reported anymore, but it was an unsatisfactory solution.

I noticed that in that version, a new stub was added for the `QueryBuilder::setParameters()` method so I tried editing it. When replacing `@return static` with `@return self`, the errors were not reported anymore. I have no idea why this fixes the issue though so please let me know if there's a better way.

### DIFF
--- a/src/Type/Doctrine/QueryBuilder/QueryBuilderMethodDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/QueryBuilder/QueryBuilderMethodDynamicReturnTypeExtension.php
@@ -79,7 +79,7 @@ class QueryBuilderMethodDynamicReturnTypeExtension implements \PHPStan\Type\Dyna
 		if ($returnType instanceof MixedType) {
 			return false;
 		}
-		return $returnType->isSuperTypeOf(new ObjectType($this->getClass()))->yes();
+		return (new ObjectType($this->getClass()))->isSuperTypeOf($returnType)->yes();
 	}
 
 	public function getTypeFromMethodCall(
--- a/src/Type/Doctrine/QueryBuilder/QueryBuilderTypeSpecifyingExtension.php
+++ b/src/Type/Doctrine/QueryBuilder/QueryBuilderTypeSpecifyingExtension.php
@@ -65,7 +65,7 @@ class QueryBuilderTypeSpecifyingExtension implements MethodTypeSpecifyingExtensi
 		if ($returnType instanceof MixedType) {
 			return new SpecifiedTypes([]);
 		}
-		if (!$returnType->isSuperTypeOf(new ObjectType($this->getClass()))->yes()) {
+		if (!(new ObjectType($this->getClass()))->isSuperTypeOf($returnType)->yes()) {
 			return new SpecifiedTypes([]);
 		}
 
--- a/tests/DoctrineIntegration/ORM/data/queryBuilder.php
+++ b/tests/DoctrineIntegration/ORM/data/queryBuilder.php
@@ -54,4 +54,18 @@ class Foo
 		return $queryBuilder->getQuery();
 	}
 
+	public function usingMethodThatReturnStatic(): ?MyEntity
+	{
+		$queryBuilder = $this->entityManager->createQueryBuilder();
+
+		$queryBuilder
+			->select('e')
+			->from(MyEntity::class, 'e')
+			->where('e.id = :id')
+			->setParameters([
+				'id' => 123,
+			]);
+
+		return $queryBuilder->getQuery()->getOneOrNullResult();
+	}
 }


ondrejmirtes:
> I have no idea why this fixes the issue though so please let me know if there's a better way.

And I have no idea what causes the issue so I don't like this fix :)

julienfalque:
I know, I just wanted to start a discussion. I'll investigate to understand the issue as soon as I find some time to.

ondrejmirtes:
One of my guesses would be that because of the typehint, this is no longer true and the extension doesn't get called: https://github.com/phpstan/phpstan-doctrine/blob/78376cbf68e0c30c2ff16d6e38f3acf1fdd386b4/src/Type/Doctrine/QueryBuilder/QueryBuilderMethodDynamicReturnTypeExtension.php#L82

It has to be something like that...

julienfalque:
Thanks for the hint, I'll have a look :)

julienfalque:
After some investigations it looks like you were right: the result of the line you shown is not always the same when I replace `static` with `self` :)

I looked at what happens for the first call that differs.

When the stub contains `self`, `$returnType` is an instance of `PHPStan\Type\ObjectType` with its `$className` property set to `'Doctrine\ORM\QueryBuilder'`. Since `self` in the context of `Doctrine\ORM\QueryBuilder` is a super type of `Doctrine\ORM\QueryBuilder`, this returns `true`.

When the stub contains `static`, `$returnType` is an instance of `PHPStan\Type\StaticType` with its `$baseClass` property set to `'Doctrine\ORM\QueryBuilder'`. In such case the method returns `false`. This is caused by the following line in `StaticType::isSuperTypeOf()`:

https://github.com/phpstan/phpstan-src/blob/916d540133554c483783efd1e1e06821922c15df/src/Type/StaticType.php#L85

It creates a *"maybe ... and ..."* which is never a *"yes"*, so calling `yes()` returns `false`.

Shouldn't `static` as a return type always be considered a super type of `QueryBuilder` as long as the class it refers to is `QueryBuilder` or one of its subclasses though?

By the way, is there a specific reason to use `static` return type in the stub? Asking because [the actual classe uses `self`](https://github.com/doctrine/orm/blob/7e5fe79349cd09a9f482460f8af22a8a6d1373ca/lib/Doctrine/ORM/QueryBuilder.php#L571).

julienfalque:
@ondrejmirtes What do you think about my investigation? Am I in the correct direction? If so I'd like to give it a try :)

ondrejmirtes:
@julienfalque Your investigation is correct but the conclusion is not.

> Shouldn't static as a return type always be considered a super type of QueryBuilder

Only parent classes and interfaces implemented by QueryBuilder can be considered supertype of QueryBuilder. `static` is by definition either QueryBuilder itself or a child class, therefore `static` is subtype of `QueryBuilder`.

The linked condition in https://github.com/phpstan/phpstan-doctrine/blob/78376cbf68e0c30c2ff16d6e38f3acf1fdd386b4/src/Type/Doctrine/QueryBuilder/QueryBuilderMethodDynamicReturnTypeExtension.php#L82 simply needs to be changed. Maybe it's sufficient to revert the queried objects?

```php
		return (new ObjectType($this->getClass())->isSuperTypeOf($returnType)->yes();
```

The condition is there to check whether QueryBuilder method basically returns itself, so like this it'd make more sense but someone has to try whether some tests fail...

julienfalque:
> @julienfalque Your investigation is correct but the conclusion is not.

Whoops, I confused super type with sub type 😅️

I managed to add a failing test case. The error only occurs if I call `setParameters()` **and** I store the query builder in a variable **before** calling methods on it, i.e. the variable does not contain the result of the last method in the chain.

Reverting the condition looks better regarding the intention of the code but does not fix the issue. For now, the only way to make the test pass is to replace `@return static` with `@return self` in the `QueryBuilder` stub. I need to investigate more.

julienfalque:
Ok I finally fixed this. Turns out there was a second place where the supertype check had to be reversed :)

ondrejmirtes:
Great, gonna release it now :)

ondrejmirtes:
Thank you!

ondrejmirtes:
Released as 0.12.23.

julienfalque:
Thanks @ondrejmirtes!

ondrejmirtes:
Sorry, had to revert it, it breaks with a custom `queryBuilderClass`.

julienfalque:
Damn, what breaks exactly? Do you have a small reproducer I could use to try fixing it?

ondrejmirtes:
Try setting custom `queryBuilderClass` and use it in any way. For example:

```
queryBuilderClass: Foo # descendant of Doctrine\ORM\QueryBuilder
```

And the code:

```php
public function doFoo(): Foo
{
    return $this->entityManager->createQueryBuilder();
}
```

Will fail because `createQueryBuilder` will return `Doctrine\ORM\QueryBuilder` but should return `Foo`.

julienfalque:
See #159.

ossinkine:
Hey @julienfalque @ondrejmirtes, the issue still reproduced for me in this example
```php
class FooRepository extends ServiceEntityRepository
{
    protected function getQuery(array $categories): Query
    {
        $queryBuilder = $this->createQueryBuilder('foo');
        $queryBuilder
            ->join('foo.bar', 'bar')
            ->andWhere('CONTAINS(foo.categories, :categories) = true')
            ->andWhere('bar.enabled = :enabled')
            ->andWhere('foo.enabled = :enabled')
            ->addOrderBy('bar.slug', 'ASC')
            ->setParameters([
                'categories' => sprintf('{%s}', implode(', ', $categories)),
                'enabled' => true,
            ])
        ;

        return $queryBuilder->getQuery();
    }
}
```

julienfalque:
I still have the issue too. The test case I wrote is likely a bit different than the actual issue. I'll give it another try.

PR #140, opened 2020-08-05
Implement dynamic return type extension for EntityRepository::createQueryBuilder()

### LINKED ISSUE #66: Compatibiliy with EntityRepository::createQueryBuilder
Hello,
It seems that query builders built with `\Doctrine\ORM\EntityRepository::createQueryBuilder` are not validating DQL, only ones built with `\Doctrine\ORM\EntityManagerInterface::createQueryBuilder`.

I'm trying to analyse a Symfony 4 project where all queryBuilder are created within repository classes with `$this->createQueryBuilder`.
Am I missing something ?

Thanks

xificurk:
closes #66

### DIFF
--- a/src/Type/Doctrine/QueryBuilder/EntityRepositoryCreateQueryBuilderDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/QueryBuilder/EntityRepositoryCreateQueryBuilderDynamicReturnTypeExtension.php
@@ -2,16 +2,14 @@
 
 namespace PHPStan\Type\Doctrine\QueryBuilder;
 
-use Doctrine\Common\Persistence\ObjectRepository;
 use PhpParser\Node\Arg;
 use PhpParser\Node\Expr\MethodCall;
-use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Identifier;
 use PhpParser\Node\Scalar\String_;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\MethodReflection;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
-use PHPStan\Type\GenericTypeVariableResolver;
+use PHPStan\Type\Generic\GenericClassStringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeWithClassName;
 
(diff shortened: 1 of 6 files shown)


src/Type/Doctrine/QueryBuilder/EntityRepositoryCreateQueryBuilderDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,65 @@
...
+
+	public function getTypeFromMethodCall(
+		MethodReflection $methodReflection,
+		MethodCall $methodCall,
+		Scope $scope
+	): Type
+	{
+		$calledOnType = $scope->getType($methodCall->var);
+		$entityName = $this->extractEntityName($calledOnType);
+		$entityNameArg = new Arg($entityName === null ? new Variable('entityName') : new String_($entityName));
```
ondrejmirtes, line 39: I really don't get this logic - we're calling `createQueryBuilder($entityName)` here but what does the hardcoded nonexisting variable means? Also this case doesn't seem to be tested.

ondrejmirtes: I think that it should just return the default return type of the method if we don't know `TEntityClass`...

xificurk: The whole logic is basically "inlining" of the method. If the entity type cannot be statically resolved, then we substitute a dummy variable for it. This strategy recognizes the query builder as having dynamic parameters. This is IMO better than simply returning default return type of the method (which would produce `Could not analyse QueryBuilder with unknown beginning.` error).

ondrejmirtes: I still don't get it. The `Could not analyse QueryBuilder with unknown beginning.` is actually wanted in this case because it's true, right? I don't like much asking for `$scope->getType()` of definitely invalid code, or possibly even misleading code. If I set the `$entityName` variable to some non-sense before this call, it'd resolve it to a type based on that non-sense.

xificurk: > The `Could not analyse QueryBuilder with unknown beginning.` is actually wanted in this case because it's true, right? 

I don't think it is. I would expect that the following two lines will always result in the exactly same error:
```php
$entityManager->createQueryBuilder()->select($alias)->from($entityName, $alias, $indexBy)->getQuery();
$entityManager->getRepository($entityName)->createQueryBuilder($alias, $indexBy)->getQuery();
```
Which would not be the case, if the default return type of method is returned.

> If I set the `$entityName` variable to some non-sense before this call, it'd resolve it to a type based on that non-sense.

I've completely overlooked this, I understand now. I'll try to fix this, somehow.

xificurk: @ondrejmirtes I rewrote the logic and removed the dummy variable. Are you ok with this version?

ondrejmirtes:
Really nice, thank you!

PR #141, opened 2020-09-11
allow EntityRepository usage by interfaces

Marmelatze:
Allows the usage of EntityRepositories by interface name.
So EntityRepositories can be passed around that implement a common interface and return entities that have a common interface. 

Currently the extension throws a MappingException, as the interface can not be found by the doctrine ClassMetadataLoader

### DIFF
--- a/src/Reflection/Doctrine/EntityRepositoryClassReflectionExtension.php
+++ b/src/Reflection/Doctrine/EntityRepositoryClassReflectionExtension.php
@@ -66,6 +66,21 @@ class EntityRepositoryClassReflectionExtension implements \PHPStan\Reflection\Me
 		}
 
 		$fieldName = $this->classify($methodFieldName);
+		// ensure all entities implementing the interface have the field
+		if (interface_exists($entityClassType->getClassName())) {
+			foreach ($objectManager->getMetadataFactory()->getAllMetadata() as $metadata) {
+				$implements = class_implements($metadata->getName());
+				if (!isset($implements[$entityClassType->getClassName()])) {
+					continue;
+				}
+				if (!$metadata->hasField($fieldName) && $metadata->hasAssociation($fieldName)) {
+					return false;
+				}
+			}
+
+			return true;
+		}
+
 		$classMetadata = $objectManager->getClassMetadata($entityClassType->getClassName());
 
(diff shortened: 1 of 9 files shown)


ruudk:
@Marmelatze Could you rebase this? It would be great to have support like this merged.
@ondrejmirtes Is it something that can be accepted?

ondrejmirtes:
Can you please explain what's the current situation on phpstan-doctrine 1.2.6 and what kind of code pattern doesn't work and what's the error message from PHPStan?

Marmelatze:
For example:

```
<?php
interface SomethingInterface {
    
}

class Something implements SomethingInterface {
    
}

class Factory
{
    /**
     * @return class-string<SomethingInterface>
     */
    public function getSomethingClass(): string
    {
        return Something::class;
    }
}

$factory = new Factory();
$entityManager->getRepository($factory->getSomethingClass())->find(1);
```

Will result in:

```
Internal error: Internal error: Class 'SomethingInterface' does not exist in file xxx
     Run PHPStan with --debug option and post the stack trace to:
     https://github.com/phpstan/phpstan/issues/new?template=Bug_report.md
     Child process error (exit code 1):
```

With debug:

```
Uncaught Doctrine\Persistence\Mapping\MappingException: Class 'SomethingInterface' does not exist in vendor/doctrine/persistence/lib/Doctrine/Persistence/Mapping/MappingException.php:94
```

ondrejmirtes:
This exception might already be prevented in phpstan-doctrine 1.2.6. If it still throws for you, then try phpstan-doctrine dev-master. Thanks :)

Marmelatze:
It work with 1.2.6 but only if using the default repository methods (find, findBy). Custom repository methods are not recognized

ondrejmirtes:
@Marmelatze Have you set your base repository class per https://github.com/phpstan/phpstan-doctrine#configuration ?

Marmelatze:
Yes. Maybe my problem is somewhat special: I have some Entities that implement one interface. All these entities have a repository that share a common Interface. phpstan-doctrine will not recognize calls to `$entityManager->getRepository($interfacedEntity)->specialMethod()`. Here all repositories for entities that implement the entity interface also implement an interface that has `specialMethod`.

I worked around this by specifying a type alias in phpstan.neon that contains all entites that implement said interface.

ondrejmirtes:
Alright ✌️

PR #143, opened 2020-09-24
Check that doctrine entities are not final

enumag:
I create all my classes as final by default. This however can cause issues with Doctrine:

```
  [Doctrine\Common\Proxy\Exception\InvalidArgumentException]
  Unable to create a proxy for a final class "<entity class>".
```

What do you think about a rule to avoid final entities?

### DIFF
--- a/rules.neon
+++ b/rules.neon
@@ -33,3 +33,9 @@ services:
 			reportUnknownTypes: %doctrine.reportUnknownTypes%
 		tags:
 			- phpstan.rules.rule
+	-
+		class: PHPStan\Rules\Doctrine\ORM\EntityNotFinalRule
+
+conditionalTags:
+	PHPStan\Rules\Doctrine\ORM\EntityNotFinalRule:
+		phpstan.rules.rule: %featureToggles.bleedingEdge%
--- a/src/Rules/Doctrine/ORM/EntityNotFinalRule.php
+++ b/src/Rules/Doctrine/ORM/EntityNotFinalRule.php
@@ -1,0 +1,56 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Doctrine\ORM;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Rules\Rule;
+use PHPStan\Type\Doctrine\ObjectMetadataResolver;
+use function sprintf;
+
+/**
+ * @implements Rule<Node\Stmt\Class_>
+ */
+class EntityNotFinalRule implements Rule
+{
+
+	/** @var \PHPStan\Type\Doctrine\ObjectMetadataResolver */
+	private $objectMetadataResolver;
+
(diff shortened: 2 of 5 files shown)


src/Rules/Doctrine/ORM/EntityRule.php
```diff
@@ -0,0 +1,68 @@
...
+			return [];
+		}
+
+		$classMetadataInfo = 'Doctrine\ORM\Mapping\ClassMetadataInfo';
+		if (!$metadata instanceof $classMetadataInfo) {
+			return [];
+		}
+
+		$errors = [];
+		if ($node->isFinal()) {
```
ondrejmirtes, line 58: Ask this first, and only if it's final then continue with checking whether it's an entity etc.

ondrejmirtes (changes requested):
Yeah, I'm fine with this. Please add it as a bleeding edge rule similar to here: https://github.com/phpstan/phpstan-strict-rules/blob/334898a32217e4605e0f9cfa3d3fc3101bda26be/rules.neon#L67-L69

enumag:
@ondrejmirtes Updated according to your suggestions. Also added a test case.

enumag:
Hmm... there is already a similar pull request: https://github.com/phpstan/phpstan-doctrine/pull/84

Except I slightly disagree with the statement that entities shouldn't have final methods. I have a case where I'm actually using final method on purpose. But then again I could just disable that particular rule in my codebase.

rules.neon
```diff
@@ -33,3 +33,7 @@ services:
 			reportUnknownTypes: %doctrine.reportUnknownTypes%
 		tags:
 			- phpstan.rules.rule
+
+conditionalTags:
+	PHPStan\Rules\Doctrine\ORM\EntityRule:
```
ondrejmirtes, line 38: Please rename the rule to something like `FinalEntityClassRule` and also add it to `services` (without any tags).

I also guess this is the reason why the build failed, so please fix that.

ondrejmirtes (changes requested):
I'm fine with the rule in this PR as it is.

enumag:
Done. :-)

rules.neon
```diff
@@ -19,6 +19,7 @@ rules:
 	- PHPStan\Rules\Doctrine\ORM\DqlRule
 	- PHPStan\Rules\Doctrine\ORM\RepositoryMethodCallRule
 	- PHPStan\Rules\Doctrine\ORM\EntityRelationRule
+	- PHPStan\Rules\Doctrine\ORM\EntityNotFinalRule
```
ondrejmirtes, line 22: It has to be added to `services`. This way you're always registering the rule and the `conditionalTags` section has no effect.

enumag:
Should be okay now.

ondrejmirtes:
Thank you!

ondrejmirtes:
Tagged as 0.12.20.

mhujer:
@enumag thanks! 👍

PR #147, opened 2020-10-02
Update README.md

ThomasLandauer:
Fixing Kernel namespace, see https://github.com/phpstan/phpstan-symfony#console-command-analysis

BTW: I had to change the path in the first line too - should I make a PR for this?:
```php
require __DIR__.'/../config/bootstrap.php';
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ For example, in a Symfony project, `object-manager.php` would look something lik
 <?php
 
 require dirname(__DIR__).'/../config/bootstrap.php';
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you!

The first line is consistent with https://github.com/phpstan/phpstan-symfony so either both of them need updating, or neither does.

ThomasLandauer:
About the file name & location: I called my file `phpstan-console-application.php`, and put it in the top directory, so that it's next to `phpstan.neon`. So I'm wondering if this should be the suggested way? Or can this file be used for something else (besides phpstan)?

ondrejmirtes:
People usually put it in the tests directory so then the paths are correcr”t.

Issue #152, opened 2020-10-19
False-positive Cannot call method on


fluffycondor:
# Bug report

PHPStan 0.12.50

`Cannot call method update() on App\Entity\MyEntity|false` when it can't be false.

### Code snippet that reproduces the problem

```php
/** @var \Doctrine\Common\Collections\Collection<MyEntity> $collection */
if ($collection->isEmpty()) {
    $collection->add($new);
} else {
    $collection->first()->update($new); // Cannot call method update() on App\Entity\MyEntity|false
}
```
### Expected output

There should not be any error, but Cannot call method update() on App\Entity\MyEntity|false raised

ondrejmirtes:
This is not a bug but a well-known limitation described in the documentation: https://phpstan.org/writing-php-code/narrowing-types

This can be solved with a simple TypeSpecifyingExtension (https://phpstan.org/developing-extensions/type-specifying-extensions). When `isEmpty()` is truthy, `first()` and other methods should return `null`. When it's falsey, `null` should be removed from `first()` and other methods typehints.

fluffycondor:
@ondrejmirtes I've tried to write a TypeSpecifyingExtension for the case based on some examples I've found in phpstan/phpstan-symfony, but it doesn't work as I expected. Everything I got - PHPStan gives me a new error
`Call to method Doctrine\Common\Collections\Collection<int,App\Entity\MyEntity>::isEmpty() will always evaluate to false.`, and `Cannot call method on` errors are still there.

Could you please explain what am I doing wrong?

This is my code:
```php
<?php

declare(strict_types=1);

namespace App\PHPStan;

use PhpParser\Node\Expr\MethodCall;
use PHPStan\Analyser\Scope;
use PHPStan\Analyser\SpecifiedTypes;
use PHPStan\Analyser\TypeSpecifier;
use PHPStan\Analyser\TypeSpecifierAwareExtension;
use PHPStan\Analyser\TypeSpecifierContext;
use PHPStan\Broker\Broker;
use PHPStan\Reflection\MethodReflection;
use PHPStan\Reflection\ParametersAcceptorSelector;
use PHPStan\Type\BooleanType;
use PHPStan\Type\MethodTypeSpecifyingExtension;
use PHPStan\Type\TypeCombinator;

final class DoctrineCollectionFirstTypeSpecifyingExtension implements MethodTypeSpecifyingExtension, TypeSpecifierAwareExtension
{
[...]

PR #153, opened 2020-10-20
Specify types for Doctrine Collection first() method

fluffycondor:
An attempt to solve #152.
PHPStan gives me a new error
`Call to method Doctrine\Common\Collections\Collection<int,App\Entity\MyEntity>::isEmpty() will always evaluate to false.`, and `Cannot call method on` errors are still there.

### DIFF
--- a/src/Type/Doctrine/Collection/FirstTypeSpecifyingExtension.php
+++ b/src/Type/Doctrine/Collection/FirstTypeSpecifyingExtension.php
@@ -16,6 +16,7 @@ use PHPStan\Type\TypeCombinator;
 
 final class FirstTypeSpecifyingExtension implements MethodTypeSpecifyingExtension, TypeSpecifierAwareExtension
 {
+
 	private const COLLECTION_CLASS = 'Doctrine\Common\Collections\Collection';
 	private const IS_EMPTY_METHOD_NAME = 'isEmpty';
 	private const FIRST_METHOD_NAME = 'first';
@@ -32,13 +33,13 @@ final class FirstTypeSpecifyingExtension implements MethodTypeSpecifyingExtensio
 		MethodReflection $methodReflection,
 		MethodCall $node,
 		TypeSpecifierContext $context
-	): bool {
-		return
-			(
-				$methodReflection->getDeclaringClass()->getName() === self::COLLECTION_CLASS
-				|| $methodReflection->getDeclaringClass()->isSubclassOf(self::COLLECTION_CLASS)
-			)
-			&& $methodReflection->getName() === self::IS_EMPTY_METHOD_NAME;
+	): bool
+	{
+		return (
+			$methodReflection->getDeclaringClass()->getName() === self::COLLECTION_CLASS
+			|| $methodReflection->getDeclaringClass()->isSubclassOf(self::COLLECTION_CLASS)
+		)
+		&& $methodReflection->getName() === self::IS_EMPTY_METHOD_NAME;
 	}
 
 	public function specifyTypes(
@@ -46,7 +47,8 @@ final class FirstTypeSpecifyingExtension implements MethodTypeSpecifyingExtensio
 		MethodCall $node,
 		Scope $scope,
 		TypeSpecifierContext $context
-	): SpecifiedTypes {
+	): SpecifiedTypes
+	{
 		$classReflection = $methodReflection->getDeclaringClass();
 		$methodVariants = $classReflection->getNativeMethod(self::FIRST_METHOD_NAME)->getVariants();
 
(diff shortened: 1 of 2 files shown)


src/Type/Doctrine/Collection/FirstTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,65 @@
...
+		MethodCall $node,
+		Scope $scope,
+		TypeSpecifierContext $context
+	): SpecifiedTypes {
+		$classReflection = $this->broker->getClass(self::COLLECTION_CLASS);
+		$methodVariants = $classReflection->getNativeMethod(self::FIRST_METHOD_NAME)->getVariants();
+
+		return $this->typeSpecifier->create(
+			new MethodCall($node->var, self::FIRST_METHOD_NAME),
+			TypeCombinator::remove(new BooleanType(), ParametersAcceptorSelector::selectSingle($methodVariants)->getReturnType()),
```
ondrejmirtes, line 56: These arguments need to be switched. It's `public static function remove(Type $fromType, Type $typeToRemove): Type`.

src/Type/Doctrine/Collection/FirstTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,65 @@
...
+	{
+		return self::COLLECTION_CLASS;
+	}
+
+	public function isMethodSupported(
+		MethodReflection $methodReflection,
+		MethodCall $node,
+		TypeSpecifierContext $context
+	): bool {
+		return $methodReflection->getName() === self::IS_EMPTY_METHOD_NAME && $context->false();
```
ondrejmirtes, line 42: Please check declaring class as well. It should be Collection or subclass of Collection.

ondrejmirtes, line 42: We can check the truthy context too, so remove that.

src/Type/Doctrine/Collection/FirstTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	public function specifyTypes(
+		MethodReflection $methodReflection,
+		MethodCall $node,
+		Scope $scope,
+		TypeSpecifierContext $context
+	): SpecifiedTypes {
+		$classReflection = $this->broker->getClass(self::COLLECTION_CLASS);
```
ondrejmirtes, line 51: You don't need to fetch the class reflection, just call `$methodReflection->getDeclaringClass()`.

src/Type/Doctrine/Collection/FirstTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,65 @@
...
+use PHPStan\Type\MethodTypeSpecifyingExtension;
+use PHPStan\Type\TypeCombinator;
+
+final class FirstTypeSpecifyingExtension implements MethodTypeSpecifyingExtension, TypeSpecifierAwareExtension
+{
+	private const COLLECTION_CLASS = 'Doctrine\Common\Collections\Collection';
+	private const IS_EMPTY_METHOD_NAME = 'isEmpty';
+	private const FIRST_METHOD_NAME = 'first';
+
+	private Broker $broker;
```
ondrejmirtes, line 24: Don't use 7.4 code for that, this repo needs to support PHP 7.1+.

ondrejmirtes (changes requested):

src/Type/Doctrine/Collection/FirstTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,64 @@
...
+		MethodCall $node,
+		Scope $scope,
+		TypeSpecifierContext $context
+	): SpecifiedTypes {
+		$classReflection = $methodReflection->getDeclaringClass();
+		$methodVariants = $classReflection->getNativeMethod(self::FIRST_METHOD_NAME)->getVariants();
+
+		return $this->typeSpecifier->create(
+			new MethodCall($node->var, self::FIRST_METHOD_NAME),
+			TypeCombinator::remove(ParametersAcceptorSelector::selectSingle($methodVariants)->getReturnType(), new BooleanType()),
```
ondrejmirtes, line 55: Since we're executing this now for all contexts, there should be an if-else about `$context->truthy()` and `$context->falsey()`. In the former case, `isFirst` is going to return `false`, in the second case, the `false` should be removed.

src/Type/Doctrine/Collection/FirstTypeSpecifyingExtension.php
fluffycondor: I have removed the context checking, but now I can't figure out where should I check for isEmpty() result?

ondrejmirtes (changes requested):
Please also add an integration test about this scenario.

You can see here how these things are tested: https://github.com/phpstan/phpstan-doctrine/tree/master/tests/DoctrineIntegration/ORM

src/Type/Doctrine/Collection/FirstTypeSpecifyingExtension.php
fluffycondor: I have added a condition, but it doesn't have any effect for some reason. The errors "Cannot call method on" still remain.

ondrejmirtes:
Hi, I'm sorry, you'll have to debug it yourself, I can't dive into this right now.

ruudk:
Can be closed as https://github.com/phpstan/phpstan-doctrine/pull/172 is merged.

PR #159, opened 2020-11-25
Fix empty DQL with QueryBuilder::setParameters()

julienfalque:
Follow up of #139.


ondrejmirtes:
Thank you, it makes sense like this.

julienfalque:
Thanks :)

PR #165, opened 2020-12-21
Use https instead of http for Doctrine link in README

matks:
Every effort counts, right 😅  ?


matks:
(I know hacktoberfest is over, this is not OSS spam although it might look at it 😆 )

ondrejmirtes:
This is fine, thank you :)

Issue #117, opened 2020-02-26
[FEATURE] Allow functions chained on execute()


tomasnorre:
I would like to suggest a rule to have following functions allowed to have `fetch()`, `fetchAll()`, `fetchColumn()` and `rowCount()` chained after `execute()` without getting the errors

```
Cannot call method fetch() on Doctrine\DBAL\Driver\Statement|int.
Cannot call method fetchAll() on Doctrine\DBAL\Driver\Statement|int.
Cannot call method fetchColumn() on Doctrine\DBAL\Driver\Statement|int.
Cannot call method rowCount() on Doctrine\DBAL\Driver\Statement|int.
```

The `execute()` returns `int` when a `count('*')` is made instead of a `select('*')`, therefor with two possible return types PHPStan complains about this.

### Code Example

```
$queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable($this->tableName);
$queueRec = $queryBuilder
    ->select('*')
    ->from($this->tableName)
    ->where(
        $queryBuilder->expr()->eq('qid', $queryBuilder->createNamedParameter($queueId))
    )
    ->execute()
    ->fetch();
```

My problem currently, as stated on twitter, I'm quite new to PHPStan and would not know where to start to add this rule.

If someone can point me into the direction how to implement this, I would happy to have a look at this, and I don't see it as a error with the current implementation of Doctrine `select()` and `count()`.

Hope that someone can point me in the right direction.

ondrejmirtes:
Hi, this isn't a case for a new rule as you're just trying to prevent a false positive. This is a case for a dynamic return type extension (https://phpstan.org/developing-extensions/dynamic-return-type-extensions). You need one for `execute()` and walk up the `MethodCall` AST chain to see if `select` or `count` was called at the beginning...

It's easy to write the extension for this case:

```php
$queryBuilder
    ->select('*')
    ->from($this->tableName)
    ->where(
        $queryBuilder->expr()->eq('qid', $queryBuilder->createNamedParameter($queueId))
    )
    ->execute()
    ->fetch();
```

It's a little bit harder to write it for this case:

```php
$tmp = $queryBuilder
    ->select('*')
    ->from($this->tableName)
    ->where(
        $queryBuilder->expr()->eq('qid', $queryBuilder->createNamedParameter($queueId))
    );
$tmp->execute()->fetch();
```

But we can start supporting only the first one as that will cover most uses I think.

[...]

PR #167, opened 2021-01-08
Introduce a DynamicMethodReturnTypeExtension for the DBAL QueryBuilde…

mitelg:
…r execute method

fixes https://github.com/phpstan/phpstan-doctrine/issues/117

As mentioned in the issue above, this is a first try to provide a solution. 

I would like to write a test for this new extension, but I don't know how to start here. I already added a test class, which also contains a mock, which I would like to test.

### DIFF
--- a/src/Type/Doctrine/DBAL/QueryBuilder/QueryBuilderExecuteMethodExtension.php
+++ b/src/Type/Doctrine/DBAL/QueryBuilder/QueryBuilderExecuteMethodExtension.php
@@ -1,0 +1,57 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Doctrine\DBAL\QueryBuilder;
+
+use Doctrine\DBAL\Driver\ResultStatement;
+use Doctrine\DBAL\Query\QueryBuilder;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Identifier;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\DynamicMethodReturnTypeExtension;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\Type;
+
+class QueryBuilderExecuteMethodExtension implements DynamicMethodReturnTypeExtension
+{
+
+	public function getClass(): string
+	{
+		return QueryBuilder::class;
+	}
+
+	public function isMethodSupported(MethodReflection $methodReflection): bool
+	{
+		return $methodReflection->getName() === 'execute';
+	}
+
+	public function getTypeFromMethodCall(MethodReflection $methodReflection, MethodCall $methodCall, Scope $scope): Type
+	{
+		$defaultReturnType = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+
+		$queryBuilderType = new ObjectType(QueryBuilder::class);
+		$var = $methodCall->var;
+		while ($var instanceof MethodCall) {
+			$varType = $scope->getType($var->var);
+			if (!$queryBuilderType->isSuperTypeOf($varType)->yes()) {
+				return $defaultReturnType;
+			}
+
+			$nameObject = $var->name;
+			if (!($nameObject instanceof Identifier)) {
+				return $defaultReturnType;
+			}
+
+			$name = $nameObject->toString();
+			if ($name === 'select' || $name === 'addSelect') {
+				return new ObjectType(ResultStatement::class);
+			}
+
+			$var = $var->var;
+		}
+
+		return $defaultReturnType;
+	}
+
+}
--- a/tests/DoctrineIntegration/ORM/data/dbalQueryBuilderExecuteDynamicReturn.php
+++ b/tests/DoctrineIntegration/ORM/data/dbalQueryBuilderExecuteDynamicReturn.php
@@ -1,0 +1,89 @@
+<?php declare(strict_types=1);
+
+namespace PHPStan\DoctrineIntegration\ORM\DbalQueryBuilderExecuteDynamicReturn;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Query\QueryBuilder;
+
+class Example
+{
+	/** @var Connection */
+	private $connection;
+
+	public function __construct(Connection $connection)
+	{
+		$this->connection = $connection;
+	}
+
+	/**
+	 * @return mixed[]
+	 */
+	public function testCaseOne(int $userId): array
+	{
+		return $this->connection->createQueryBuilder()
+			->select('*')
+			->from('user')
+			->where('user.id = :id')
+			->setParameter('id', $userId)
+			->execute()
+			->fetchAll();
+	}
+
+	/**
+	 * @return mixed[]
+	 */
+	public function testCaseTwo(int $userId): array
+	{
+		$qb = $this->connection->createQueryBuilder();
+		$qb->select('*');
+		$qb->from('user');
+		$qb->where('user.id = :id');
+		$qb->setParameter('id', $userId);
+
+		return $qb->execute()->fetchAll();
+	}
+
+	/**
+	 * @return mixed[]
+	 */
+	public function testCaseThree(?int $userId = null): array
+	{
+		$qb = $this->connection->createQueryBuilder();
+		$qb->select('*');
+		$qb->from('user');
+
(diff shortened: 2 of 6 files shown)


ondrejmirtes:
Hi, thanks. Extensions aren't tested like this. You simply write some code under https://github.com/phpstan/phpstan-doctrine/tree/master/tests/DoctrineIntegration/ORM, verify that it has a false positive without the extension, and that the extension solves the problem and no error is reported.

It's enough to create a new .php file under data/, add it to EntityManagerIntegrationTest.php::dataTopics() and run EntityManagerIntegrationTest.php. It will update the JSON files with expectations.

tests/DoctrineIntegration/ORM/data/dbalQueryBuilderExecuteDynamicReturn.php
```diff
@@ -0,0 +1,89 @@
...
+
+	/**
+	 * @return mixed[]
+	 */
+	private function testCaseFourB(QueryBuilder $qb, int $userId): array
+	{
+		$qb->where('user.id = :id');
+		$qb->setParameter('id', $userId);
+
+		return $qb->execute()->fetchAll();
```
mitelg, line 87: this is the only case, that does not work till now. but I guess, this is okay for the first implementation

mitelg:
hey @ondrejmirtes 
thanks for the hints. I implemented a test case successfully and it looks quite good imo :blush:

mitelg:
how could this be fixed? the tests are failing on lowest dependencies

```
-        "message": "Cannot call method fetchAll() on Doctrine\\DBAL\\Driver\\ResultStatement|int.",
+        "message": "Cannot call method fetchAll() on Doctrine\\DBAL\\Driver\\Statement|int.",
```

ondrejmirtes:
@mitelg Two possible solutions:

1) Figure out why it's failing and write the code such as it works on both highest and lowest dependencies.
2) Bump the lowest version in require-dev so that it works without change in code.

mitelg:
I bumped the lowest version of the doctrin/dbal requirement. so now, everything should be fine :+1:

.gitignore
```diff
@@ -1,2 +1,3 @@
 /composer.lock
 /vendor
+/.idea
```
ondrejmirtes, line 3: Not needed, you should ignore this on a global level in your OS Git settings.

src/Type/Doctrine/DBAL/QueryBuilder/QueryBuilderExecuteMethodExtension.php
```diff
@@ -0,0 +1,90 @@
...
+
+	public function getTypeFromMethodCall(MethodReflection $methodReflection, MethodCall $methodCall, Scope $scope): Type
+	{
+		$defaultReturnType = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+
+		$var = $methodCall->var;
+		while ($var instanceof MethodCall) {
+			$nameObject = $var->name;
+			if (!($nameObject instanceof Identifier)) {
+				continue;
```
ondrejmirtes, line 41: This will lead to infinite loop.

src/Type/Doctrine/DBAL/QueryBuilder/QueryBuilderExecuteMethodExtension.php
```diff
@@ -0,0 +1,90 @@
...
+			$name = $nameObject->toString();
+			if ($name === 'select' || $name === 'addSelect') {
+				return new ObjectType(ResultStatement::class);
+			}
+
+			$var = $var->var;
+		}
+
+		// Get class method node
+		/** @var Variable $classMethodNode */
```
ondrejmirtes, line 53: This is most certainly wrong, it can be any `Expr`.

ondrejmirtes (changes requested):
Hi, what I'd like more as the first step is an extension that would only handle the `testCaseOne` case. There we can simply traverse over the MethodCall chain to figure out what's happening.

Trying to track the calls through variable assignments like this isn't very sophisticated and it's inconsistent with what phpstan-doctrine does with ORM's QueryBuilder. (It introduces a custom QueryBuilderType that extends ObjectType and it simply relies on already implemented PHPStan internals to do the job of type inference.)

So I'd like to merge and release the simple version that handles only `testCaseOne` and then we can make it more advanced with the same approach this extension already uses in other scenarios.

Thank you.

src/Type/Doctrine/DBAL/QueryBuilder/QueryBuilderExecuteMethodExtension.php
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	public function getTypeFromMethodCall(MethodReflection $methodReflection, MethodCall $methodCall, Scope $scope): Type
+	{
+		$defaultReturnType = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+
+		$name = '';
+		$var = $methodCall->var;
+		while ($var instanceof MethodCall) {
+			$nameObject = $var->name;
```
mitelg, line 36: this could be `Identifier|Expr` so I need to check on the correct type to get the name string. I cannot simply cast them to string. without this if below, PHPStan would complain :wink:

mitelg:
hey @ondrejmirtes 

I reduced the code in `QueryBuilderExecuteMethodExtension` to only support the first test case. I left the other test cases, to make use of them in the future.

Unfortunately I'm still in the beginning of understanding the internals of PHPStan and how things are done. If you could guide me, I will take care of the other cases in another PR

src/Type/Doctrine/DBAL/QueryBuilder/QueryBuilderExecuteMethodExtension.php
```diff
@@ -0,0 +1,51 @@
...
+			if ($nameObject instanceof Identifier) {
+				$name = $nameObject->toString();
+			}
+
+			if ($name === 'select' || $name === 'addSelect') {
```
ondrejmirtes, line 41: This `if` should be inside the `if ($nameObject instanceof Identifier) {`.

src/Type/Doctrine/DBAL/QueryBuilder/QueryBuilderExecuteMethodExtension.php
ondrejmirtes: Yes, this is fine, in the else branch I guess we could return `$defaultReturnType` instead of continuing with the loop.

ondrejmirtes (changes requested):
Yes, this is much simpler and manageable and we're mostly done here :)

src/Type/Doctrine/DBAL/QueryBuilder/QueryBuilderExecuteMethodExtension.php
mitelg: I used an early return. hope I got that right :blush:

src/Type/Doctrine/DBAL/QueryBuilder/QueryBuilderExecuteMethodExtension.php
```diff
@@ -0,0 +1,51 @@
...
+		return $methodReflection->getName() === 'execute';
+	}
+
+	public function getTypeFromMethodCall(MethodReflection $methodReflection, MethodCall $methodCall, Scope $scope): Type
+	{
+		$defaultReturnType = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+
+		$var = $methodCall->var;
+		while ($var instanceof MethodCall) {
+			$nameObject = $var->name;
```
ondrejmirtes, line 35: One more thing here - we're not checking that `$var->var` is still `Doctrine\DBAL\Query\QueryBuilder`. You might have this situation. The best way to do that is early return in case:

```php
$queryBuilderType = new ObjectType(QueryBuilder::class); // this line should be above the while loop
$varType = $scope->getType($var->var);
if (!$queryBuilderType->isSuperTypeOf($varType)->yes()) {
    return $defaultReturnType;
}
```

Once this is resolved, I'll merge and release it :)

ondrejmirtes (changes requested):

src/Type/Doctrine/DBAL/QueryBuilder/QueryBuilderExecuteMethodExtension.php
mitelg: done :heavy_check_mark:

ondrejmirtes:
Perfect, thank you!

PR #173, opened 2021-02-09
Narrow type of Collection::last() when using Collection::isEmpty()

ruudk:
Now that we have #172 , we should do the same for `last()`.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -313,6 +313,6 @@ services:
 
 	# Doctrine Collection
 	-
-		class: PHPStan\Type\Doctrine\Collection\FirstTypeSpecifyingExtension
+		class: PHPStan\Type\Doctrine\Collection\IsEmptyTypeSpecifyingExtension
 		tags:
 			- phpstan.typeSpecifier.methodTypeSpecifyingExtension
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Thank you :)

ruudk:
⚡ Would be great to have this tagged too 🙏  Thanks!

ondrejmirtes:
Tagged 👌

VincentLanglet:
Hi @ruudk, would is be possible to do the same with the count value ?

image

ruudk:
Not sure if that's possible TBH. /cc @ondrejmirtes

ondrejmirtes:
Just assert that last() is not null instead...

VincentLanglet:
> Just assert that last() is not null instead...

I know, and that's what 'im doing.
But the same argument could have be use for `isEmpty`, and still this great improvement was made.

If I assert that the count is > 0, I could avoid to add another assert that last/first is not false.

Would it be implementable ?

ondrejmirtes:
Nope, the TypeSpecifyingExtension API doesn’t allow it.

PR #175, opened 2021-02-22
Use @template-covariant instead of @template in EntityRepository

jeroennoten:
Using `@template-covariant` instead of `@template` on `EntityRepository` allows for more flexible use cases. Example: https://github.com/phpstan/phpstan/issues/4582

### DIFF
--- a/stubs/Collections/Collection.stub
+++ b/stubs/Collections/Collection.stub
@@ -8,7 +8,7 @@ use IteratorAggregate;
 
 /**
  * @template TKey
- * @template T
+ * @template-covariant T
  * @extends IteratorAggregate<TKey, T>
  * @extends ArrayAccess<TKey|null, T>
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
I worry that such flexibility isn't possible because of the error you're getting in `matching` method.

ondrejmirtes:
(Doctrine's Collection is mutable.)

jeroennoten:
It should be possible if we also change the `Collection`s `ArrayAccess`s annotations to `@template-covariant`, or not?

ondrejmirtes:
Doctrine's Collection is mutable so it cannot have `@template-covariant`. What's PHPStan communicating to you in https://github.com/phpstan/phpstan/issues/4582 is that you can't really do that for type safety reasons.

PR #183, opened 2021-04-23
Adding object-manager.php for Symfony 5

ThomasLandauer:
The code is taken from https://github.com/phpstan/phpstan-symfony/issues/141#issuecomment-817485623

Question:

> ... allows DQL validation when enabled

How can you enable this?

PR #184, opened 2021-04-24
Clarifiying DQL validation

ThomasLandauer:
See https://github.com/phpstan/phpstan-doctrine/pull/183#issuecomment-826062440

I think this is more important than the repository thing, so I moved it in front. Do you agree?


ondrejmirtes:
Thank you!

PR #189, opened 2021-05-20
Add testcase for entities that are annotated final

DanielBadura:
I just update the plugin from `0.12.33` to `0.12.34` and after this i got several messages about making my entities not final because of how doctrine works, which is right. But my entities are not final on language level. I would expect that this message is not displayed if i annotate the entity to be final. So i added a testcase based on my expected behavior. If it is working right now as expected i would change the test accordingly :)

Why does this happen now? I dont know, i looked into the diff between the two releases and did not find which changes caused this. This feature is bleeding edge since `0.12.20` it should have happened before i guess.

PS: i did not run the testcase locally wanted the workflow to do the work 😆

### DIFF
--- a/src/Rules/Doctrine/ORM/EntityNotFinalRule.php
+++ b/src/Rules/Doctrine/ORM/EntityNotFinalRule.php
@@ -34,7 +34,7 @@ class EntityNotFinalRule implements Rule
 		if ($classReflection === null) {
 			throw new \PHPStan\ShouldNotHappenException();
 		}
-		if (!$classReflection->isFinal()) {
+		if (!$classReflection->isFinalByKeyword()) {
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
This line probably needs to be fixed to: `$classReflection->isFinalByKeyword()`

https://github.com/phpstan/phpstan-doctrine/blob/9d798be80c14aa2ce870edbc71d91504b474d634/src/Rules/Doctrine/ORM/EntityNotFinalRule.php#L37

DanielBadura:
This worked, thanks @ondrejmirtes 👍

ondrejmirtes:
Thanks! Gonna release it right away 👍

PR #208, opened 2021-08-31
Add allowNullablePropertyForRequiredField option

VincentLanglet:

PR #214, opened 2021-09-15
Pass allowNullablePropertyForRequiredField to EntityRelationRule

VincentLanglet:


VincentLanglet:
Hi @ondrejmirtes, this is a follow up of the option introduced in https://github.com/phpstan/phpstan-doctrine/pull/208
The option is working properly for properties but to me the same behavior is expected for OtO, OtM, ...

ondrejmirtes:
Is this solving a real-world problem when using entities as form DTOs?

VincentLanglet:
> Is this solving a real-world problem when using entities as form DTOs?

Sure, for instance I have those errors in my work project.
The option introduced solved 95% of the "property might be null, but database is not",
but I still have some errors with this message for OtO/MtO. This PR should fix my issue.

ondrejmirtes:
Thank you!

VincentLanglet:
Thanks, would a release possible ? :)

ondrejmirtes:
I've already opened 1.0-dev, the release is a few weeks out.

VincentLanglet:
> I've already opened 1.0-dev, the release is a few weeks out.

I see, it will be 1.0 for every extensions too ; didn't think about it

PR #218, opened 2021-10-06
Allow to use both ORM and ODM

VincentLanglet:
Might solve https://github.com/phpstan/phpstan-doctrine/issues/53#issuecomment-936667144
I need to check tomorrow.

### DIFF
--- a/src/Type/Doctrine/ObjectMetadataResolver.php
+++ b/src/Type/Doctrine/ObjectMetadataResolver.php
@@ -110,14 +110,18 @@ final class ObjectMetadataResolver
 		if ($metadata instanceof $ormMetadataClass) {
 			/** @var \Doctrine\ORM\Mapping\ClassMetadata $ormMetadata */
 			$ormMetadata = $metadata;
-			return $ormMetadata->customRepositoryClassName ?? $this->getResolvedRepositoryClass();
+			return $ormMetadata->customRepositoryClassName
+				?? $this->repositoryClass
+				?? 'Doctrine\ORM\EntityRepository';
 		}
 
 		$odmMetadataClass = 'Doctrine\ODM\MongoDB\Mapping\ClassMetadata';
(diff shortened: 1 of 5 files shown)


VincentLanglet:
> I need a test here so I can understand what are the ramifications.

I added one. Don't hesitate if you need more explanation or the test to be presented differently, because it wasn't easy to me to reproduce my issue.

If you implement your own ObjectManager in order to support both ORM and ODM depending on the class of the entity, you need the repository class to also depends on the entities. The object manager is correctly finding the classmetadata, so if the entity/document has a custom repository there is no issues. But in case there is not, prior to these change the default value was always the same repository and it could depend on the classmetadata.

I think it won't impact the perf of basics users since `??` has basically no cost, and it works for me (and for the tests).

VincentLanglet:
Friendly ping @ondrejmirtes. Do you have time to take a look ? This would help me to migrate on phpstan 1 :)

ondrejmirtes:
@VincentLanglet  My top priority right now is PHP 8.1 support (https://phpstan.org/blog/plan-to-support-php-8-1). The main action is now happening in https://github.com/ondrejmirtes/BetterReflection/tree/ng and https://github.com/phpstan/phpstan-src/tree/ng-better-reflection. Once this is finished and released, the operations will return to normal eventually. I now have 187 emails in my inbox worthy of my attention (all related to PHPStan), so please stay tuned.

ondrejmirtes:
I'm gonna fix it myself, stay tuned.

VincentLanglet:
Just saw that `getResolvedRepositoryClass` is only returning DocumentRepository when an objectManager was provided
https://github.com/phpstan/phpstan-doctrine/blob/master/src/Type/Doctrine/ObjectMetadataResolver.php#L163-L167 which is currently the blocking point.

Feel free to ping me if you want that I test your branch

Issue #219, opened 2021-10-07
Calling `QueryBuilder::update` is updating the type of the vairable or `$this`.


VincentLanglet:
```
\PHPStan\dumpType($this);
$this->getEntityManager()->getRepository(WaitingList::class)->createQueryBuilder('w')->update();
\PHPStan\dumpType($this);
```
give me
```
263    Dumped type: $this(App\Core\Workflow\WorkflowProductStoreEventSubscriber)     
265    Dumped type: Doctrine\ORM\QueryBuilder
```

In the same way
```
$repo = $this->getEntityManager()->getRepository(WaitingList::class);
$repo->createQueryBuilder('w')->update();
\PHPStan\dumpType($repo);
```
is giving `Dumped type: Doctrine\ORM\QueryBuilder` instead of a repository.

If I manually remove the `QueryBuilderTypeSpecifyingExtension`, it fix my issue.

PR #220, opened 2021-10-07
Do not specify a variable which is not a query builder

VincentLanglet:
Solve https://github.com/phpstan/phpstan-doctrine/issues/219


ondrejmirtes:
Next release is going to be out after PHPStan 1.0 is released which is still a few weeks away.

ondrejmirtes:
> Does requiring the dev-master branch of phpstan and the extensions could be stable enough to enjoy the recent bugfix without having too much issues ?

I can't recommend that. What you should do instead is to use your own fork of phpstan-doctrine with the fix. And then switch back to the original once it's released.

Issue #233, opened 2021-11-23
Cannot load object manager


mbrodala:
I tried to [load the object manager as documented](https://github.com/phpstan/phpstan-doctrine#configuration) to fix https://github.com/phpstan/phpstan-doctrine/issues/229 but this doesn't seem to work.

My `phpstan.neon`:

```neon
parameters:
  level: 5
  paths:
    - src
  doctrine:
    objectManagerLoader: tests/doctrine-object-manager.php
```

The `tests/doctrine-object-manager.php`:

```php
<?php

declare(strict_types=1);

use App\Kernel;
use Symfony\Component\Dotenv\Dotenv;

require __DIR__ . '/../vendor/autoload.php';

(new Dotenv())->bootEnv(__DIR__ . '/../.env');

$kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
$kernel->boot();

return $kernel->getContainer()->get('doctrine')->getManager();
```

Result:

```
$ phpstan
Note: Using configuration file /.../phpstan.neon.
 87/87 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 -- ----------------------------------------------------------------------------------------------------------------------------------------------------------- 
     Error                                                                                                                                                      
 -- ----------------------------------------------------------------------------------------------------------------------------------------------------------- 
     Reached internal errors count limit of 50, exiting...                                                                                                      
     Internal error: Internal error: Object manager could not be loaded in file /.../src/.../<file>.php                                          
     Run PHPStan with --debug option and post the stack trace to:                                                                                               
     https://github.com/phpstan/phpstan/issues/new?template=Bug_report.md                                                                                       
     ...  many more ...                                                                                   
     Child process error (exit code 1):                                                                                                                         
     Child process error (exit code 1):                                                                                                                         
 -- ----------------------------------------------------------------------------------------------------------------------------------------------------------- 

                                                                                                                        
 [ERROR] Found 60 errors                                                                                                
```

The suggested debug run:

```
$ phpstan --debug
Note: Using configuration file /.../phpstan.neon.
/.../src/.../<file>.php
Uncaught PHPStan\ShouldNotHappenException: Object manager could not be loaded in /.../vendor/phpstan/phpstan-doctrine/src/Type/Doctrine/ObjectMetadataResolver.php:67
[...]

PR #234, opened 2021-11-24
Use explicit errors for invalid object manager loader file

mbrodala:
Clearly indicate if the file exists or is not readable, also mention the file path.

See #233

### DIFF
--- a/src/Type/Doctrine/ObjectMetadataResolver.php
+++ b/src/Type/Doctrine/ObjectMetadataResolver.php
@@ -4,7 +4,7 @@ namespace PHPStan\Type\Doctrine;
 
 use Doctrine\Persistence\ObjectManager;
 use PHPStan\Reflection\ReflectionProvider;
-use function file_exists;
+use function is_file;
 use function is_readable;
 
 final class ObjectMetadataResolver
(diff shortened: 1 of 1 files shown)


src/Type/Doctrine/ObjectMetadataResolver.php
```diff
@@ -60,11 +60,18 @@ public function getObjectManager(): ?ObjectManager
 
 	private function loadObjectManager(string $objectManagerLoader): ?ObjectManager
 	{
-		if (
-			!file_exists($objectManagerLoader)
-			|| !is_readable($objectManagerLoader)
-		) {
-			throw new \PHPStan\ShouldNotHappenException('Object manager could not be loaded');
+		if (!file_exists($objectManagerLoader)) {
```
ondrejmirtes, line 63: Please use is_file() here instead.

ondrejmirtes (changes requested):
Also please fix the build failure.

src/Type/Doctrine/ObjectMetadataResolver.php
mbrodala: I think behavior changes should be done separately. Here I mostly move code around without functional changes.

ondrejmirtes: I asked you to make a change. It will leave the code in a better state. It can be a separate commit.

mbrodala: OK, done.

ondrejmirtes:
Build still failing

mbrodala:
> Build still failing

Yes, should be fixed now. There are too many unusual conventions/expectations here.

ondrejmirtes:
Thank you!

PR #238, opened 2021-12-16
Discover repository class from `#[Entity]` attribute on PHP 8

ruudk:
By doing it like this, we don't need to instantiate the object manager at all which makes PHPStan run faster and more reliable. Very often I had to manually clear my Symfony cache and re-run PHPStan.

Unfortunately, BetterReflection doesn't support attributes yet so therefore we need to use native `ReflectionClass`.

See https://github.com/phpstan/phpstan/discussions/5863#discussioncomment-1823297

### DIFF
--- a/attribute-errors.neon
+++ b/attribute-errors.neon
@@ -1,0 +1,6 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Call to an undefined method ReflectionClass::getAttributes\\(\\)\\.$#"
+			count: 1
+			path: src/Type/Doctrine/ObjectMetadataResolver.php
--- a/ignore-by-php-version.neon.php
+++ b/ignore-by-php-version.neon.php
@@ -1,0 +1,15 @@
+<?php declare(strict_types = 1);
+
+use PHPStan\DependencyInjection\NeonAdapter;
+
+$adapter = new NeonAdapter();
+
+$config = [];
+
+if (PHP_VERSION_ID < 80000) {
+	$config = array_merge_recursive($config, $adapter->load(__DIR__ . '/attribute-errors.neon'));
+}
+
+$config['parameters']['phpVersion'] = PHP_VERSION_ID;
+
+return $config;
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,6 +1,7 @@
 includes:
 	- extension.neon
 	- rules.neon
+	- ignore-by-php-version.neon.php
 	- vendor/phpstan/phpstan-strict-rules/rules.neon
 	- vendor/phpstan/phpstan-phpunit/extension.neon
 	- vendor/phpstan/phpstan-phpunit/rules.neon
--- a/src/Type/Doctrine/ObjectMetadataResolver.php
+++ b/src/Type/Doctrine/ObjectMetadataResolver.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Type\Doctrine;
 
+use Doctrine\ORM\Mapping\Entity;
 use Doctrine\Persistence\ObjectManager;
 use PHPStan\Reflection\ReflectionProvider;
 use function is_file;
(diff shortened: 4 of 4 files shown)


ruudk:
Hmm, I don't understand why on PHP 8.0 it fails.

ondrejmirtes:
Please try the nice way via ClassReflection::getNativeReflection() and require `phpstan/phpstan: "^1.3"` here.

ondrejmirtes:
The fact that PHPStan reports an error like that is weird and I should look into it, but it might still work correctly. Hard to tell without a test.

ruudk:
For some reason I cannot install the dependencies for this project anymore:
```

  Problem 1
    - Root composer.json requires doctrine/mongodb-odm ^1.3 || ^2.1 -> satisfiable by doctrine/mongodb-odm[1.3.0-RC1, ..., 1.3.x-dev, 2.1.0, ..., 2.3.x-dev].
    - doctrine/orm[2.9.1, ..., 2.11.x-dev] require doctrine/common ^3.0.3 -> satisfiable by doctrine/common[3.0.3, ..., 3.2.x-dev].
    - Conclusion: don't install doctrine/common 3.2.0 (conflict analysis result)
    - Root composer.json requires doctrine/orm ^2.9.1 -> satisfiable by doctrine/orm[2.9.1, ..., 2.11.x-dev].

```

Weird that it just works on CI.

ondrejmirtes:
Yeah, you need `--ignore-platform-reqs` if you don't have the mongodb PHP extension installed locally.

ondrejmirtes:
I really don't know, it works in the playground: https://phpstan.org/r/fa2489cc-dfd7-4a61-8865-c29e83a8a2a6

ruudk:
```
\PHPStan\dumpType($this->reflectionProvider->getClass($className));
		\PHPStan\dumpType($this->reflectionProvider->getClass($className)->getNativeReflection());
		\PHPStan\dumpType($this->reflectionProvider->getClass($className)->getNativeReflection()->getAttributes());
```
```
  101    Dumped type: PHPStan\Reflection\ClassReflection
  102    Dumped type: ReflectionClass
  103    Call to an undefined method ReflectionClass::getAttributes().
```

🤷

ondrejmirtes:
I know! It's caused by `config.platform.php` here: https://github.com/phpstan/phpstan-doctrine/blob/04cace0bfd8d72cd5af13d90180d15fcb459d1ca/composer.json#L36-L41

ruudk:
I removed the php platform and now phpunit fails. Shouldn't PHPunit be installed differently on PHP < 7.3?

I fixed the name of the mongodb extension, as it was wrong.

ruudk:
Oke, so the only problem left seems to be a bug in PHPStan, am I right?

https://phpstan.org/r/a34ea8cd-2986-4bbf-82e2-a1041cb4a6de

If so, I can report it on phpstan/phpstan

ruudk:
@ondrejmirtes I hoped that PHPStan 1.3 would solve the problem with attributes, as the release notes mention this:
> This brings plenty of bugfixes, for example PHP 8.0 Attributes can now be read in static reflection context

but I still cannot solve the problem.

Any idea how to proceed?

ondrejmirtes:
What PHPStan 1.3 fixed was an internal PHP error when accessing `ClassReflection::getNativeReflection()->getAttributes()` in some cases.

Are you trying to solve the problem on PHP 7.x from this example? https://phpstan.org/r/a34ea8cd-2986-4bbf-82e2-a1041cb4a6de This is usually solved by having different PHPStan configs for different PHP versions (when a project is analysed on different PHP cersions).

One notable example is PHPStan itself: https://github.com/phpstan/phpstan-src/blob/master/build/ignore-by-php-version.neon.php

This file is included in `phpstan.neon`: https://github.com/phpstan/phpstan-src/blob/b614f4af90b26e24d653b9d11a27679ee9b459ad/build/phpstan.neon#L10

ruudk:
@ondrejmirtes It worked 🎉  Ready for review.

ruudk:
While writing some tests, I notice that `$this->reflectionProvider->hasClass($className)` returns false, and therefore it skips my code. Isn't it possible to reflect unknown classes?

ondrejmirtes:
Looks like I'm gonna solve this in a general way for all the rules :) https://github.com/phpstan/phpstan-doctrine/pull/253

PR #239, opened 2021-12-16
Test on PHP 8.1

ruudk:


ondrejmirtes:
Thank you! I don't expect any problems, the extensions are already tested on 8.1 here: https://github.com/phpstan/phpstan/actions/workflows/extension-tests.yml

But it's useful anyway :)

PR #240, opened 2021-12-16
Composer > Remove `--no-suggest`

ruudk:
You are using the deprecated option "--no-suggest". It has no effect and will break in Composer 3.


ondrejmirtes:
Thank you!

PR #241, opened 2021-12-16
Test all jobs on PHP 8.1

ruudk:


ruudk:
@ondrejmirtes This is ready. Had to conflict with older version of `doctrine/lexer` to solve PHP 8.1 compatibility. Also bumped PHPUnit to get rid of PHP 8.1 deprecations.

ruudk:
While testing this in #238 I think we can better handle this differently.

Will update this PR.

ruudk:
@ondrejmirtes Ok, I think this is better.

Only 1 problem left, on PHP 7.1 an older version of `doctrine/orm` is installed and therefore 1 `LevelsTestCase` test fails:
```diff
- "message": "Parameter #1 $className of method Doctrine\\Persistence\\ObjectManager::getRepository() expects class-string<nonexistentClass>, string given.",\n
+ "message": "Parameter #1 $className of method Doctrine\\ORM\\EntityManagerInterface::getRepository() expects class-string<nonexistentClass>, string given.",
```

Does the `LevelsTestCase` have a way to switch based on PHP version?

ruudk:
It's finally green. What a hassle. Hope you agree with the changes :)

ruudk:
@ondrejmirtes can this be merged?

ondrejmirtes:
Thank you!

PR #260, opened 2022-01-24
We can not assume the type of Query::getResult() if ResultType may be void

### LINKED ISSUE #259: Wrong type inference for QueryBuilder::update and QueryBuilder::delete.
The following code
```
$result = $this->getEntityManager()
            ->getRepository(ProductCrossSell::class)
            ->createQueryBuilder('pcs')
            ->where('pcs.id IN (:ids)')
            ->setParameter('ids', $ids)
            ->delete()
            ->getQuery()
            ->getResult();

\PHPStan\dumpType($result);
```
Is considering `$result` as an array. But it's an int: the number of deleted elements. Same with `update()`

It was introduce by https://github.com/phpstan/phpstan-doctrine/pull/232 cc @arnaud-lb

arnaud-lb:
Fixes #259 

Normally we assume the type of `Query<ResultType>::getResult()` to be `array<ResultType>`. However, we also assume that `Query<void>::getResult()` returns a `int` (`void` denotes an INSERT or DELETE query).

So we can not assume the type of `Query<ResultType>::getResult()` to be `array<ResultType>` if there is a chance that `ResultType` is `void`.

In this PR I change the `DynamicReturnTypeExtension` of Query::getResult() so that no return type is assumed when ResultType may be void.

Note:

This has an impact on this feature when used in conjunction with template types, as this is not valid anymore:

```
/**
 * @template T
 * @param Query<T> $q
 * @return array<T>
 */
function test($q) {
    return $q->getResult(); // mixed
}
```

However, this is valid:

```
/**
 * @template T of array|object A non-void type parameter
 * @param Query<T> $q
 * @return array<T>
 */
function test($q) {
    return $q->getResult(); // array<T>
}
```

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
@@ -93,10 +93,18 @@ final class QueryResultDynamicReturnTypeExtension implements DynamicMethodReturn
 		Type $queryResultType
 	): Type
 	{
-		if ($queryResultType instanceof VoidType) {
+		$isVoidType = (new VoidType())->isSuperTypeOf($queryResultType);
+
+		if ($isVoidType->yes()) {
 			// A void query result type indicates an UPDATE or DELETE query.
 			// In this case all methods return the number of affected rows.
 			return new IntegerType();
+		}
+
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Thank you!

PR #253, opened 2022-01-21
Resolve entity metadata without objectManagerLoader

ondrejmirtes:

PR #264, opened 2022-01-24
Register DBAL types automatically so that `EntityColumnRule` works without defining a `objectManagerLoader`

ruudk:
The goal is to use the extension without having to provide a `objectManagerLoader`.

The base functionality was added in https://github.com/phpstan/phpstan-doctrine/pull/253.

Because `EntityColumnRule` uses `DescriptorRegistry` and that uses DBAL's `Type::getTypesMap` we
need to make sure that all types are defined there as well.

To make that easy, I quickly hacked the `ReflectionDescriptor` to automatically register the type.

As a bonus, I also did the `UuidTypeDescriptor`.

Now I can run it without having to instantiate the Symfony container.

@ondrejmirtes This is a POC, what's the best way to address this?

### DIFF
--- a/src/Rules/Doctrine/ORM/EntityColumnRule.php
+++ b/src/Rules/Doctrine/ORM/EntityColumnRule.php
@@ -72,7 +72,7 @@ class EntityColumnRule implements Rule
 
 	public function processNode(Node $node, Scope $scope): array
 	{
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Can you please show what kind of code pattern PHPStan fails - what is this trying to solve? I have some idea but perhaps when we explain things clearly, a solution emerges more easily :)

PR #267, opened 2022-01-24
Aggregate functions affect nullability only when the query is not grouped

### LINKED ISSUE #265: groupBy adds null to QueryResultType
I have the following QueryBuilder defined:
```php
/**
     * @return array<Category>
     */
    public function getAllForNavigation(): array
    {
        return $this->getEntityManager()->createQueryBuilder()
            ->select('cat')
            ->from(Category::class, 'cat')
            ->join(Navigation::class, 'nav', Join::WITH, 'nav.joinColumn = cat.joinColumn')
            ->groupBy('nav.joinColumn')
            ->getQuery()
            ->getResult();
    }
```
This reports an error:
> should return array<Category> but returns array<Category|null>
When the `groupBy`-call is removed, everything works fine.

arnaud-lb:
Fixes #265

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultTypeWalker.php
+++ b/src/Type/Doctrine/Query/QueryResultTypeWalker.php
@@ -97,7 +97,10 @@ class QueryResultTypeWalker extends SqlWalker
 	private $descriptorRegistry;
 
 	/** @var bool */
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

PR #270, opened 2022-01-27
queryResultTest: more tests

janedbal:
I want to point out multiple imperfections of query result type inference on this MR.


janedbal:
Closing, I dont even recall the issue here.

PR #273, opened 2022-01-28
Add IDENTITY() type inference support

arnaud-lb:
Supports `IDENTITY(a.assoc)` or `IDENTITY(a.assoc, 'id_field')` where `id_field` is the name of one of the id fields of the target class.

Does not support `IDENTITY(a.assoc)` when the target id field is also an association (this is inferred to mixed).


ondrejmirtes:
Thank you!

janedbal:
Thank you!

PR #288, opened 2022-02-04
Determine queryBuilderType if getQuery is called immediately

### LINKED ISSUE #266: Keep QueryBuilderType across method calls
I don't  know if this is currently even possible, but it would be nice, if Phpstan could determine the type of the QueryBuilder, even if it is defined in another method. Currently something like this results in a mixed return type:
```php
    /**
     * @return array<Entity>
     */
    public function reuseQueryBuilder(): array
    {
        return $this->getQueryBuilder()
            ->getQuery()
            ->getResult();
    }

    private function getQueryBuilder(): QueryBuilder
    {
        return $this->createQueryBuilder('test');
    }
```

Khartir:
Closes #266 

Uses `searchOtherMethodsForQueryBuilderBeginning` also in `QueryBuilderGetQueryDynamicReturnTypeExtension`.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -110,9 +110,7 @@ services:
 	-
 		class: PHPStan\Type\Doctrine\QueryBuilder\QueryBuilderMethodDynamicReturnTypeExtension
 		arguments:
-			parser: @defaultAnalysisParser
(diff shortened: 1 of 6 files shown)


ondrejmirtes:
Thank you! I wanted to do something similar :)

Issue #294, opened 2022-02-23
Strings not allowed in QueryBuilder::where parameters


petaak:
Shouldn't strings be allowed in the QueryBuilder::where parameters?
 
```
$qb->select(['u'])
   ->from(User::class, 'u')
   ->where(
       $qb->expr()->isNotNull('u.foo')
   )
```
I think the example above is a valid usecase and the isNotNull method returns string. But in the QueryBuilder stub string is not allowed (https://github.com/phpstan/phpstan-doctrine/commit/5a5454cf8d7916b39c6423beed592fb141743937#diff-197b8a38b2ae878d60d27968735dd0b7208e015f401be1b2c5761add8fe8b85dR25)

ondrejmirtes:
What error are you seeing?

petaak:
```
Parameter #1 $predicates of method Doctrine\ORM\QueryBuilder::where() expects array|object|literal-string, string given.
```

craigfrancis:
The [isNotNull()](https://github.com/doctrine/orm/blob/856c3143f8c45028c80d338183e2ebe0822cf6e9/lib/Doctrine/ORM/Query/Expr.php#L509) method is very simple in what it does:

```php
    /**
     * Creates an IS NOT NULL expression with the given arguments.
     *
     * @param string $x Field in string format to be restricted by IS NOT NULL.
     *
     * @return string
     */
    public function isNotNull($x)
    {
        return $x . ' IS NOT NULL';
    }
```

So the input `$x` would need to be a `literal-string` for it to return a `literal-string`, which I think is reasonable.

Having a quick look at the other methods in `Expr` that `@return string`, I should do:

1. countDistinct($x)
2. isNull($x)
3. isNotNull($x)
4. between($val, $x, $y)

In the future I will be looking at the other `Expr\*` objects (e.g. [Expr\Comparison](https://github.com/doctrine/orm/blob/856c3143f8c45028c80d338183e2ebe0822cf6e9/lib/Doctrine/ORM/Query/Expr/Comparison.php)), but those shouldn't be affected by the `where()` requirement of accepting `literal-string|object|array<mixed>`.

ondrejmirtes:
> So the input $x would need to be a literal-string for it to return a literal-string, which I think is reasonable.

Yeah, needs a dynamic return type extension then.

craigfrancis:
[...]

PR #298, opened 2022-03-06
Dynamic return types for some Expr methods

craigfrancis:
This is for [Issue 294](https://github.com/phpstan/phpstan-doctrine/issues/294).

### DIFF
--- a/src/Type/Doctrine/QueryBuilder/Expr/ExpressionBuilderDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/QueryBuilder/Expr/ExpressionBuilderDynamicReturnTypeExtension.php
@@ -59,10 +59,10 @@ class ExpressionBuilderDynamicReturnTypeExtension implements DynamicMethodReturn
 			if (count($args) > 0) {
 				$literalStringInput = true;
 				foreach ($args as $arg) {
-					if ($scope->getType($arg->value)->isLiteralString()->yes()) {
-						continue;
+					if (!$scope->getType($arg->value)->isLiteralString()->yes()) {
+						$literalStringInput = false;
+						break;
 					}
-					$literalStringInput = false;
 				}
 				if ($literalStringInput) {
 					$defaultReturnType = new IntersectionType([
--- a/src/Type/Doctrine/QueryBuilder/Expr/ExpressionBuilderDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/QueryBuilder/Expr/ExpressionBuilderDynamicReturnTypeExtension.php
@@ -8,17 +8,11 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Rules\Doctrine\ORM\DynamicQueryBuilderArgumentException;
-use PHPStan\Type\Accessory\AccessoryLiteralStringType;
-use PHPStan\Type\Accessory\AccessoryNonEmptyStringType;
 use PHPStan\Type\Doctrine\ArgumentsProcessor;
 use PHPStan\Type\Doctrine\ObjectMetadataResolver;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
-use PHPStan\Type\IntersectionType;
-use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
-use function count;
 use function get_class;
-use function in_array;
 use function is_object;
 use function method_exists;
 
@@ -53,26 +47,6 @@ class ExpressionBuilderDynamicReturnTypeExtension implements DynamicMethodReturn
 	public function getTypeFromMethodCall(MethodReflection $methodReflection, MethodCall $methodCall, Scope $scope): Type
 	{
 		$defaultReturnType = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
-
-		if ($defaultReturnType instanceof StringType && in_array($methodReflection->getName(), ['countDistinct', 'isNull', 'isNotNull', 'between', 'abs'], true)) {
-			$args = $methodCall->getArgs();
-			if (count($args) > 0) {
-				$literalStringInput = true;
-				foreach ($args as $arg) {
-					if (!$scope->getType($arg->value)->isLiteralString()->yes()) {
-						$literalStringInput = false;
-						break;
-					}
-				}
-				if ($literalStringInput) {
-					$defaultReturnType = new IntersectionType([
-						$defaultReturnType,
-						new AccessoryNonEmptyStringType(),
-						new AccessoryLiteralStringType(),
-					]);
-				}
-			}
-		}
 
 		$objectManager = $this->objectMetadataResolver->getObjectManager();
 		if ($objectManager === null) {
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Something gets lost in the type inference, I don't want those tests to fail.

craigfrancis:
Same, I'm just not sure how best to solve it...

`QueryBuilderDqlRuleSlowTest` and `QueryBuilderDqlRuleTest` use [qbExprIsNull](https://github.com/phpstan/phpstan-doctrine/blob/master/tests/Rules/Doctrine/ORM/data/query-builder-dql.php#L203) and [qbExprIsNullSyntaxError](https://github.com/phpstan/phpstan-doctrine/blob/master/tests/Rules/Doctrine/ORM/data/query-builder-dql.php#L215), both use `isNull()`, and now report "Could not analyse QueryBuilder with dynamic arguments" (what this patch does).

Should I:

1. Bypass that check (maybe update [QueryBuilderDqlRule.php](https://github.com/phpstan/phpstan-doctrine/blob/master/src/Rules/Doctrine/ORM/QueryBuilderDqlRule.php#L78));
2. Use a different function - e.g. for the syntax error test, use `->like('e.nickname', 'x')`;
3. Do something else?

I've also got some notes on [Issue 294](https://github.com/phpstan/phpstan-doctrine/issues/294#issuecomment-1059954277), as I think the dynamic arguments approach will have a problem with methods like `$qb->expr()->mod()` that [return a stringable object](https://github.com/doctrine/orm/blob/856c3143f8c45028c80d338183e2ebe0822cf6e9/lib/Doctrine/ORM/Query/Expr.php#L346), and even if their inputs are `literal-string`, this won't be known by `isNull()`, `between()`, and others.

ondrejmirtes:
I'm not sure why these tests fail now, can you please debug that and explain it to me? Thanks.

craigfrancis:
While I've not addressed the issue where methods like `$qb->expr()->abs('my_field')` return a stringable object like `return new Expr\Func('ABS', [$x])`, and these are not being seen as a `literal-string`; the rest seems to be working.

---

As to the failing tests, they were due to my code simply returning a `IntersectionType`, early in the process, if all of the arguments were `literal-string`:

```php
if ($literalStringInput) {
  return new IntersectionType([
      new StringType(),
      new AccessoryNonEmptyStringType(),
      new AccessoryLiteralStringType(),
    ]);
}
```

This was a problem because the failing tests required a `ConstantStringType` to be returned, with the `value` set to the returned string from the Expr method, which happens at the end of `getTypeFromMethodCall()`:

```php
$exprValue = $expr->{$methodReflection->getName()}(...$args);
if (is_object($exprValue)) {
  return new ExprType(get_class($exprValue), $exprValue);
}

return $scope->getTypeFromValue($exprValue);
```

For example, the `$queryBuilder->expr()->isNull('e.nickname')` was returning a generic `literal-string&non-empty-string`, when it should have been returning a `ConstantStringType('e.nickname IS NULL')`, so the final SQL could be checked.

---

With `Expr\Func`, it might involve `NewExprDynamicReturnTypeExtension.php`, as it's used for this object, but it only has `getTypeFromStaticMethodCall()` which isn't called for `__construct` or `__toString()`.

src/Type/Doctrine/QueryBuilder/Expr/ExpressionBuilderDynamicReturnTypeExtension.php
```diff
@@ -48,6 +54,26 @@ public function getTypeFromMethodCall(MethodReflection $methodReflection, Method
...
 
+		if ($defaultReturnType instanceof StringType && in_array($methodReflection->getName(), ['countDistinct', 'isNull', 'isNotNull', 'between', 'abs'], true)) {
+			$args = $methodCall->getArgs();
+			if (count($args) > 0) {
+				$literalStringInput = true;
+				foreach ($args as $arg) {
+					if ($scope->getType($arg->value)->isLiteralString()->yes()) {
+						continue;
+					}
+					$literalStringInput = false;
```
ondrejmirtes: There could be a `break`.

tests/Type/Doctrine/data/QueryResult/expressionBuilderGetQuery.php
```diff
@@ -0,0 +1,79 @@
...
+	public function isNullNonLiteralStringExprFunc(EntityManagerInterface $em): void
+	{
+		$result = $em->createQueryBuilder()->expr()->isNull($qb->expr()->mod('field', '0'));
+		assertType('string', $result);
+	}
+
+	public function betweenNonLiteralStringExprFunc(EntityManagerInterface $em): void
+	{
+		$result = $em->createQueryBuilder()->expr()->between($qb->expr()->abs('field'), '10', '30');
+		assertType('string', $result);
```
ondrejmirtes, line 76: I feel a bit crazy here because I don't see a single test that would actually execute the new added lines:

```php

				if ($literalStringInput) {
					$defaultReturnType = new IntersectionType([
						$defaultReturnType,
						new AccessoryNonEmptyStringType(),
						new AccessoryLiteralStringType(),
					]);
				}
```

I'd expect there to be an assertType for `literal-string&non-empty-string` but there isn't one. So what's happening here? From this changeset it looks like the tests would pass even without any change in `src`?

ondrejmirtes (changes requested):

src/Type/Doctrine/QueryBuilder/Expr/ExpressionBuilderDynamicReturnTypeExtension.php
craigfrancis: Good point, has been updated.

tests/Type/Doctrine/data/QueryResult/expressionBuilderGetQuery.php
craigfrancis: Note to self... a `literal-string&non-empty-string` will be returned when `$this->objectMetadataResolver->getObjectManager()` returns NULL (i.e. [`$defaultReturnType` is returned early](https://github.com/phpstan/phpstan-doctrine/blob/1.3.x/src/Type/Doctrine/QueryBuilder/Expr/ExpressionBuilderDynamicReturnTypeExtension.php#L53)).

Currently this method returns an `EntityManagerInterface`, so the rest of the method is executed, the [expr function is actually executed](https://github.com/phpstan/phpstan-doctrine/blob/1.3.x/src/Type/Doctrine/QueryBuilder/Expr/ExpressionBuilderDynamicReturnTypeExtension.php#L82), and it returns a `ConstantStringType` instead. I think I need an EntityManager that does not connect to the database.

ondrejmirtes:
I feel like a bad guy, but I reverted your `src/` changes and the tests still pass :) Please send another PR (create a new test case with different .neon config to reproduce the problem).

BTW we could try the new hot thing here, require `phpstan/phpstan:^1.6` and implement the logic in stubs with the help of conditional return types :) So that the `$defaultReturnType` would already have the correct type.

1) We need to use `selectFromArgs` instead of `selectSingle`.
2) The conditional return type for the methods would look like this: `@return ($x is literal-string ? literal-string : string)`

ondrejmirtes:
Thank you.

craigfrancis:
No worries, it's your project, and it would be good to use the "new hot thing"... unfortunately I've got meetings this afternoon, and I'm off walking on the weekend, so will continue next week.

PR #313, opened 2022-04-04
Document that `EntityManager::flush()` can throw DBAL's `UniqueConstraintViolationException`

### LINKED ISSUE #295: Allow UniqueConstraintViolationException on EntityManager::flush()
The following code will produce an error:

```php
try {
   $entityManager->flush();
} catch( UniqueConstraintViolationException $ex ) {
   // do something with $ex here
}
```

The error message is `Doctrine\DBAL\Exception\UniqueConstraintViolationException is never thrown in the try block`. This is kind-of-true, since that exception is not declared on `EntityManager::flush`, but bubbles up from the DBAL implementation. It's been raised as an issue on the repo ( https://github.com/doctrine/orm/issues/7780 ), but so far it looks like this is "by design" and won't be fixed. 

Could you create an "exception" (no pun intended) for this exception in PHPStan to ignore `catch` segments that look for `UniqueConstraintViolationException`?

VasekPurchart:
Fixes https://github.com/phpstan/phpstan-doctrine/issues/295 and as mentioned there - there is no intention in adding it in Doctrine ORM and I also think it should not be there. So either PHPStan has to understand that the exception can go from DBAL to `flush()` or we have to help him and that is why I think it should be in this repo.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -18,6 +18,7 @@ parameters:
 			- Doctrine\ORM\Query
 	stubFiles:
 		- stubs/Criteria.stub
+		- stubs/DBAL/Exception/UniqueConstraintViolationException.stub
 		- stubs/DocumentManager.stub
 		- stubs/DocumentRepository.stub
 		- stubs/EntityManager.stub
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Thank you!

PR #315, opened 2022-04-05
Test that `UniqueConstraintViolationException` from `EntityManager::flush()` is not marked as dead catch

VasekPurchart:
Test for https://github.com/phpstan/phpstan-doctrine/pull/313 .

### DIFF
--- a/tests/Rules/Exceptions/CatchWithUnthrownExceptionRuleTest.php
+++ b/tests/Rules/Exceptions/CatchWithUnthrownExceptionRuleTest.php
@@ -1,0 +1,24 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Exceptions;
+
+use PHPStan\Rules\Rule;
+use PHPStan\Testing\RuleTestCase;
+
+/**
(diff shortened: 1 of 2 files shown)


tests/Rules/Exceptions/CatchWithUnthrownExceptionRuleTest.php
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * @extends RuleTestCase<CatchWithUnthrownExceptionRule>
+ */
+class CatchWithUnthrownExceptionRuleTest extends RuleTestCase
+{
+
+	protected function getRule(): Rule
+	{
+		return new CatchWithUnthrownExceptionRule();
```
ondrejmirtes: You should be able to get rid of the BC message with: `self::getContainer()->getByType(CatchWithUnthrownExceptionRule::class);`

ondrejmirtes (changes requested):

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #324, opened 2022-04-22
Dynamic return types for some Expr methods - Part 2

craigfrancis:
Continuing on from [PR 298](https://github.com/phpstan/phpstan-doctrine/pull/298), for [Issue 294](https://github.com/phpstan/phpstan-doctrine/issues/294).

I'm starting with:

- Failing tests, confirming these functions don't return a `literal-string` when there is no Object Manager.
- Changed the requirements to `phpstan/phpstan:^1.6`.
- Changed `selectSingle` to `selectFromArgs`.

I'll continue this next week, but thought I should upload what I've got so far just incase I'm going in the wrong direction.

---

[Previous Diff](https://github.com/phpstan/phpstan-doctrine/pull/298/commits/2f6f12fa8fbfd8a702e933152878f59f08f29bc1).

"Implement the logic in **stubs** with the help of conditional return types"... "The conditional return type for the methods would look like this: `@return ($x is literal-string ? literal-string : string)`"

### DIFF
--- a/stubs/ORM/Query/Expr.stub
+++ b/stubs/ORM/Query/Expr.stub
@@ -1,0 +1,34 @@
+<?php
+
+namespace Doctrine\ORM\Query;
+
+class Expr
+{
+
+	/**
+	 * @param string $x
+	 * @return ($x is literal-string ? literal-string&non-empty-string : string)
+	 */
+	public function isNull($x)
+	{
+	}
+
+	/**
+	 * @param string $x
+	 * @return ($x is literal-string ? literal-string&non-empty-string : string)
+	 */
+	public function isNotNull($x)
+	{
+	}
+
+	/**
+	 * @param string $val
+	 * @param string $x
+	 * @param string $y
+	 * @return ($val is literal-string ? ($x is literal-string ? ($y is literal-string ? literal-string&non-empty-string : string) : string) : string)
+	 */
+	public function between($val, $x, $y)
+	{
+	}
+
+}
--- a/tests/Type/Doctrine/data/QueryResult/expressionBuilderGetQueryNoObjectManager.php
+++ b/tests/Type/Doctrine/data/QueryResult/expressionBuilderGetQueryNoObjectManager.php
@@ -8,7 +8,7 @@ use Doctrine\ORM\QueryBuilder;
 use QueryResult\Entities\Many;
 use function PHPStan\Testing\assertType;
 
-class ExpressionBuilderGetQuery
+class ExpressionBuilderGetQueryNoObjectManager
 {
 	private function nonLiteralString(string $value): string {
 		return $value; // Using the 'string' return type to provide a non `literal-string`, e.g. $_POST['field'];
@@ -17,7 +17,7 @@ class ExpressionBuilderGetQuery
 	public function isNullLiteralString(EntityManagerInterface $em): void
 	{
 		$result = $em->createQueryBuilder()->expr()->isNull('field');
-		assertType("'field IS NULL'", $result); // A ConstantStringType isLiteralString
+		assertType('literal-string&non-empty-string', $result);
 	}
 
 	public function isNullNonLiteralString(EntityManagerInterface $em): void
@@ -30,7 +30,7 @@ class ExpressionBuilderGetQuery
 	public function isNotNullLiteralString(EntityManagerInterface $em): void
 	{
 		$result = $em->createQueryBuilder()->expr()->isNotNull('field');
-		assertType("'field IS NOT NULL'", $result); // A ConstantStringType isLiteralString
+		assertType('literal-string&non-empty-string', $result);
 	}
 
 	public function isNotNullNonLiteralString(EntityManagerInterface $em): void
@@ -43,7 +43,7 @@ class ExpressionBuilderGetQuery
 	public function betweenLiteralString(EntityManagerInterface $em): void
 	{
 		$result = $em->createQueryBuilder()->expr()->between('field', "'value_1'", "'value_2'");
-		assertType("'field BETWEEN \'value_1\' AND \'value_2\''", $result); // A ConstantStringType isLiteralString
+		assertType('literal-string&non-empty-string', $result);
 	}
 
 	public function betweenNonLiteralString1(EntityManagerInterface $em): void
--- a/extension.neon
+++ b/extension.neon
@@ -39,6 +39,7 @@ parameters:
 		- stubs/ORM/Mapping/ClassMetadata.stub
 		- stubs/ORM/Mapping/ClassMetadataInfo.stub
 		- stubs/ORM/ORMException.stub
+		- stubs/ORM/Query/Expr.stub
 		- stubs/ORM/Query.stub
 		- stubs/ORM/Query/Expr/Comparison.stub
 		- stubs/ORM/Query/Expr/Composite.stub
--- a/src/Type/Doctrine/QueryBuilder/Expr/ExpressionBuilderDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/QueryBuilder/Expr/ExpressionBuilderDynamicReturnTypeExtension.php
@@ -46,7 +46,7 @@ class ExpressionBuilderDynamicReturnTypeExtension implements DynamicMethodReturn
 
 	public function getTypeFromMethodCall(MethodReflection $methodReflection, MethodCall $methodCall, Scope $scope): Type
 	{
-		$defaultReturnType = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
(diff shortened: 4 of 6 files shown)


ondrejmirtes:
Yeah, all that remains are the stubs, hopefully they work :)

craigfrancis:
Yep, I'm fairly sure it's working for `isNull($x)` and `isNotNull($x)`.

I just need to read the documentation, as I got stuck with:

- [`between($val, $x, $y)`](https://github.com/doctrine/orm/blob/d5503644314ed825fe4b276a588458a033fa3e7c/lib/Doctrine/ORM/Query/Expr.php#L645) - Not sure how to check all values, as `&`/`&&`/`AND` was returing `mixed`
- [`countDistinct($x)`](https://github.com/doctrine/orm/blob/d5503644314ed825fe4b276a588458a033fa3e7c/lib/Doctrine/ORM/Query/Expr.php#L263) - Can accept 1 or more arguments, all must be `literal-string`.

stubs/ORM/Query/Expr.stub
```diff
@@ -0,0 +1,37 @@
...
+    // /**
+    //  * @return ($x is array<literal-string> ? literal-string&non-empty-string : string)
+    //  */
+    // public function countDistinct($x)
+    // public function countDistinct(...$x)
+    // {
+    // }
+
+    // /**
+    //  * @return ($val is literal-string AND $x is literal-string AND $y is literal-string ? literal-string&non-empty-string : string)
```
VincentLanglet: You can follow this example
```
/**
 * @param int|float $a
 * @param int|float $b
 * @return ($a is int ? ($b is int ? int : float) : float)
 */
function add($a, $b) {
    return $a + $b;
}
```

I don't know if AND is supported

stubs/ORM/Query/Expr.stub
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Doctrine\ORM\Query;
+
+class Expr
+{
+
+    /**
+     * @return ($x is literal-string ? literal-string&non-empty-string : string)
```
VincentLanglet:
```suggestion
     * @param string $x
     * @return ($x is literal-string ? literal-string&non-empty-string : string)
```

stubs/ORM/Query/Expr.stub
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public function isNull($x)
+    {
+    }
+
+    /**
+     * @return ($x is literal-string ? literal-string&non-empty-string : string)
```
VincentLanglet:
```suggestion
     * @param string $x
     * @return ($x is literal-string ? literal-string&non-empty-string : string)
```

craigfrancis:
Thank you @VincentLanglet, I've added and committed your suggestions, and will spend a bit of time looking at `countDistinct()` and how it can work with any number of parameters.

tests/Type/Doctrine/data/QueryResult/expressionBuilderGetQueryNoObjectManager.php
```diff
@@ -0,0 +1,93 @@
...
+{
+	public function isNullLiteralString(EntityManagerInterface $em): void
+	{
+		$result = $em->createQueryBuilder()->expr()->isNull('field');
+		assertType('literal-string&non-empty-string', $result);
+	}
+
+	public function isNullNonLiteralString(EntityManagerInterface $em): void
+	{
+		$field = strtolower('field'); // Non literal-string, e.g. $_POST['field'];
```
VincentLanglet: I don't think it's a good idea to use `strtolower` in the test.

You should instead do
```
public function isNullNonLiteralString(EntityManagerInterface $em, string $field): void
```

Because strtolower of a literal string is a literal string. One day we'll implement the conditional type or dynamic return type extension in phpstan and it will force to update all the tests here.

VincentLanglet (changes requested):

tests/Type/Doctrine/data/QueryResult/expressionBuilderGetQueryNoObjectManager.php
craigfrancis: I'm happy to change this, but I don't think it needs it... the `literal-string` type is following the [is_literal() RFC](https://wiki.php.net/rfc/is_literal) (which I really need to do a second version of), and the only modification that can be done to a `literal-string` is when it's concatenated with another `literal-string` (to help adoption). The only native functions that can return a `literal-string` are `str_repeat()`, `str_pad()`, `implode()`, and `join()`; as these functions simply concatenate their inputs. I did mention in the RFC about supporting other functions, and noted that `strtoupper()` might be reasonable, but it was decided to not allow any other string modifications, to keep the implementation simpler, not affect performance, and make it easier to prove the security.

stubs/ORM/Query/Expr.stub
craigfrancis: Thank you @VincentLanglet, that's exactly what I needed.

If I'm reading the [TypeParser](https://github.com/phpstan/phpdoc-parser/blob/129a63b3bc7caeb593c224c41f420675e63cfefc/src/Parser/TypeParser.php#L251) correctly (when it's doing a [subParse()](https://github.com/phpstan/phpdoc-parser/blob/129a63b3bc7caeb593c224c41f420675e63cfefc/src/Parser/TypeParser.php#L43), due to there being a [TOKEN_OPEN_PARENTHESES](https://github.com/phpstan/phpdoc-parser/blob/129a63b3bc7caeb593c224c41f420675e63cfefc/src/Parser/TypeParser.php#L75)), and checking the [ConditionalTypeForParameterNode](https://github.com/phpstan/phpdoc-parser/blob/1.4.x/src/Ast/Type/ConditionalTypeForParameterNode.php) object, then conditional return types use the format `(%s %s %s ? %s : %s)` where the second `%s` is either "is" or "is not" - which implies there is no support for operators like AND.

craigfrancis:
@ondrejmirtes, there's no rush for this (I've got to get back to work)... but can you check if I'm doing this correctly...

```php
<?php

/**
 * @param literal-string ...$x
 * @return ($x is array<literal-string> ? literal-string : string)
 */
function example(...$x) {
    return implode(', ', $x);
}

$a = example('A', 'B', 'C');
$b = example('A', strtolower('B'), 'C');
```

That errors with:

- Line 7: `Condition "literal-string is array<literal-string>" in conditional return type is always false.`
- Line 12: `Parameter #2 ...$x of function example expects literal-string, non-empty-string given.`

The line 12 error is expected (for `$b`), and shows the [variadic function parameter check](https://phpstan.org/writing-php-code/phpdocs-basics#variadic-functions) is working.

As we should not be requiring `literal-string`'s for the input, we can now change that to the more relaxed `@param string ...$x`, which removes the line 12 error, and the line 7 error changes to:

- Line 7: `Condition "string is array<string>" in conditional return type is always false.`

I've probably been looking at this too long, but I'm surprised both [$subjectType](https://github.com/phpstan/phpstan-src/blob/f8be122188aa6721cec9872bd61676ee150a5311/src/Rules/PhpDoc/ConditionalReturnTypeRuleHelper.php#L44) and [$targetType](https://github.com/phpstan/phpstan-src/blob/f8be122188aa6721cec9872bd61676ee150a5311/src/Rules/PhpDoc/ConditionalReturnTypeRuleHelper.php#L58) change from `literal-string` to `string`... and `$x` (the first part, `$subjectType`) is not considered an array?

VincentLanglet:
What about splitting this PR in half ?
All the working dynamic return types can already be merged :)

craigfrancis:
Yep, good point @VincentLanglet, the `@template` approach is taking me a while to figure out.

@ondrejmirtes, is this ok to be merged, and I'll do `countDistinct()` separately?

VincentLanglet:
@craigfrancis I think you have conflict to solve

craigfrancis:
Thanks @VincentLanglet

ondrejmirtes:
Please rebase this PR (there are conflicts), thanks :)

craigfrancis:
Rebase done... hopefully (did mess up order at one point);

(ps; I'll look at the other issues later this week, just got back from holiday, and my inbox is a mess again).

ondrejmirtes:
Hi, I prefer the Git history clean - which means there should be commit(s) rebased on top of the latest branch (1.3.x). But your branch contains some merge commits like `Merge branch 'phpstan:1.3.x' into issue-294`.

Feel free to fix the history, or just do a single commit (or several commits) with your changes.

craigfrancis:
I'm not proud of myself, but one reset branch, with a new single commit from a diff, and a force-push... hopefully that's a bit cleaner.

ondrejmirtes:
Perfect, thank you :)

PR #325, opened 2022-04-26
@Version field is always written

MichaelKubovic:
Hi, I believe there is missing condition somewhere that would say that optimistic lock field, managed by ORM is always written.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -15,9 +15,6 @@ parameters:
 		- tests/*/data-php-*/*
 		- tests/Rules/Doctrine/ORM/entity-manager.php
 
-	propertyAlwaysWrittenTags:
-		- Version
-
 	reportUnmatchedIgnoredErrors: false
 
 	ignoreErrors:
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -15,6 +15,9 @@ parameters:
 		- tests/*/data-php-*/*
 		- tests/Rules/Doctrine/ORM/entity-manager.php
 
+	propertyAlwaysWrittenTags:
+        - Version
+
 	reportUnmatchedIgnoredErrors: false
 
 	ignoreErrors:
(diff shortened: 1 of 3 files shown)


phpstan.neon
```diff
@@ -15,6 +15,9 @@ parameters:
 		- tests/*/data-php-*/*
 		- tests/Rules/Doctrine/ORM/entity-manager.php
 
+	propertyAlwaysWrittenTags:
+		- Version
```
ondrejmirtes: This needs to be implemented differently so that it also works for attributes https://www.doctrine-project.org/projects/doctrine-orm/en/2.11/reference/attributes-reference.html#attrref_version

ondrejmirtes (changes requested):

phpstan.neon
MichaelKubovic: I could add a condition to `PropertiesExtension::isAlwaysWritten()` that would check if the property name matches `ClassMetadataInfo::$versionField`. Would that work?

ondrejmirtes: Yes 😊

tests/Rules/Doctrine/ORM/data/MyEntity.php
```diff
@@ -19,6 +19,12 @@ class MyEntity
 	 */
 	private $id;
 
+	/**
+	 * @ORM\Column(type="integer")
+     * @ORM\Version
```
MichaelKubovic, line 24: We currently don't have a test-suite for entities with attributes. How would you test this?

MichaelKubovic: I've found out that removing the `propertyAlwaysWrittenTags` does not make the tests fail. So blindly adding this property here is also not a good way to test it. I'm not sure how to approach this because there is no `Rule` I could refer to in test based on the `RuleTestCase`.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #326, opened 2022-04-26
Add literal-string check for andWhere

VincentLanglet:
This was introduced for where, but andWhere is working the same way.

### DIFF
--- a/stubs/bleedingEdge/ORM/QueryBuilder.stub
+++ b/stubs/bleedingEdge/ORM/QueryBuilder.stub
@@ -143,5 +143,14 @@ class QueryBuilder
 	{
 
 	}
+	
+	/**
+	 * @param literal-string|object|array<mixed> $predicates
+	 * @return $this
+	 */
+	public function andWhere($predicates)
+	{
+
+	}
 
 }


VincentLanglet:
I also have an issue with the code
```
$rootAlias = current($query->getQueryBuilder()->getRootAliases());

$queryBuilder->where($rootAlias . '.statusId IN (:statusIds)')
```

Because `getRootAliases` return string[], but should be `literal-string[]`.

I'll do another PR for this.

VincentLanglet:
Done https://github.com/phpstan/phpstan-doctrine/pull/327

craigfrancis:
Hi @VincentLanglet, I do intend to support `->andWhere()`, but:

1. I'd like to address the complications that have been brought up with `->where()` first; like [expr methods](https://github.com/phpstan/phpstan-doctrine/pull/324), and your [getRootAliases()](https://github.com/phpstan/phpstan-doctrine/pull/327) example.
2. The `->andWhere()` method uses `func_get_args()` to support any number of parameters, so while your patch does protect the first value provided (better than nothing), it won't check any others.
3. As QueryBuilder has a few similar methods, like `->orWhere()`, I want to make sure my approach is correct before updating them all.

I'd like to thank you for your patch, but is it ok if we wait a bit before applying it (or something similar)?

---

Small tangent, the [->andWhere()](https://github.com/doctrine/orm/blob/1dd2b449825df2ff7b2f4dddf06797c3a97cc1ab/lib/Doctrine/ORM/QueryBuilder.php#L1129) method starts:

```php
public function andWhere() {
    $args  = func_get_args();
```

I believe [Larry Garfield](https://github.com/Crell) would like to see the deprecation/removal of `func_get_args()`, so hopefully we can get this and other Doctrine functions (like [countDistinct()](https://github.com/doctrine/orm/blob/1dd2b449825df2ff7b2f4dddf06797c3a97cc1ab/lib/Doctrine/ORM/Query/Expr.php#L263)) to use explicit [variadics](https://www.php.net/manual/en/functions.arguments.php#functions.variable-arg-list) (available since PHP 5.6), then we can have a method signature/stub that matches.

VincentLanglet:
> I'd like to thank you for your patch, but is it ok if we wait a bit before applying it (or something similar)?

Sure, I agree it's better to solve the complications that have been brought up with ->where() first

VincentLanglet:
If https://github.com/phpstan/phpstan-doctrine/pull/324 is merged, I think we can merge this too @craigfrancis (since https://github.com/phpstan/phpstan-doctrine/pull/327 is merged).

For the func_get_args, same exists for `where` https://github.com/doctrine/orm/blob/1dd2b449825df2ff7b2f4dddf06797c3a97cc1ab/lib/Doctrine/ORM/QueryBuilder.php#L1104-L1106 and still the phpdoc was added.

I think it's ok to add this phpdoc, if someone is getting a phpstan error because he rely on the `func_get_args` call, he can update his code to follow the recommended signature of the method.

craigfrancis:
Yep, that’s a good point… I’ll just note that I’m just about to disappear for a week, as I try to have a holiday with no internet :-)

VincentLanglet:
> I’ll just note that I’m just about to disappear for a week, as I try to have a holiday with no internet :-)

Have a good holiday !

@ondrejmirtes Do you have time to merge 
https://github.com/phpstan/phpstan-doctrine/pull/324 and this PR ? :)

ondrejmirtes:
Thank you.

PR #327, opened 2022-04-26
Precise getRootAliases return type

VincentLanglet:
This allow
```
$rootAlias = current($query->getQueryBuilder()->getRootAliases());

$queryBuilder->where($rootAlias . '.statusId IN (:statusIds)')
```

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -24,7 +24,6 @@ parameters:
 		- stubs/EntityManager.stub
 		- stubs/EntityManagerDecorator.stub
 		- stubs/EntityManagerInterface.stub
-		- stubs/EntityRepository.stub
 		- stubs/ServiceEntityRepository.stub
 		- stubs/MongoClassMetadataInfo.stub
 
--- a/src/Stubs/Doctrine/StubFilesExtensionLoader.php
+++ b/src/Stubs/Doctrine/StubFilesExtensionLoader.php
@@ -28,6 +28,7 @@ class StubFilesExtensionLoader implements StubFilesExtension
 
 		return [
 			$path . '/ORM/QueryBuilder.stub',
+			$path . '/EntityRepository.stub',
 		];
 	}
 
--- a/stubs/ORM/QueryBuilder.stub
+++ b/stubs/ORM/QueryBuilder.stub
@@ -5,20 +5,20 @@ namespace Doctrine\ORM;
 class QueryBuilder
 {
 
-    /**
-     * @param \Doctrine\Common\Collections\ArrayCollection<array-key, mixed>|array<mixed> $parameters
-     * @return static
-     */
-    public function setParameters($parameters)
-    {
+	/**
+	 * @param \Doctrine\Common\Collections\ArrayCollection<array-key, mixed>|array<mixed> $parameters
+	 * @return static
+	 */
+	public function setParameters($parameters)
+	{
 
-    }
+	}
 
-    /**
-     * @return Query<mixed>
-     */
-    public function getQuery()
-    {
-    }
+	/**
+	 * @return Query<mixed>
+	 */
+	public function getQuery()
+	{
+	}
 
 }
--- a/stubs/ORM/QueryBuilder.stub
+++ b/stubs/ORM/QueryBuilder.stub
@@ -5,19 +5,27 @@ namespace Doctrine\ORM;
 class QueryBuilder
 {
 
-	/**
-	 * @param \Doctrine\Common\Collections\ArrayCollection<array-key, mixed>|array<mixed> $parameters
-	 * @return static
-	 */
-	public function setParameters($parameters)
-	{
+    /**
+     * @param \Doctrine\Common\Collections\ArrayCollection<array-key, mixed>|array<mixed> $parameters
+     * @return static
+     */
+    public function setParameters($parameters)
+    {
 
-	}
+    }
 
     /**
      * @return Query<mixed>
      */
     public function getQuery()
     {
+    }
+
+    /**
+    * @return list<literal-string>
+    */
+    public function getRootAliases()
+    {
+
     }
 }
--- a/stubs/ORM/QueryBuilder.stub
+++ b/stubs/ORM/QueryBuilder.stub
@@ -21,11 +21,4 @@ class QueryBuilder
     {
     }
 
-    /**
-    * @return list<literal-string>
-    */
-    public function getRootAliases()
-    {
-
-    }
 }
--- a/stubs/bleedingEdge/EntityRepository.stub
+++ b/stubs/bleedingEdge/EntityRepository.stub
@@ -1,0 +1,74 @@
+<?php
+
+namespace Doctrine\ORM;
+
+use Doctrine\Common\Collections\Criteria;
+use Doctrine\Persistence\ObjectRepository;
+
+/**
+ * @template TEntityClass of object
+ * @implements ObjectRepository<TEntityClass>
+ */
+class EntityRepository implements ObjectRepository
+{
+
+	/** @var class-string<TEntityClass> */
+	protected $_entityName;
+
+	/**
+	 * @phpstan-param mixed $id
+	 * @phpstan-param int|null $lockMode
+	 * @phpstan-param int|null $lockVersion
+	 * @phpstan-return TEntityClass|null
+	 */
+	public function find($id, $lockMode = null, $lockVersion = null);
+
+	/**
+	 * @phpstan-return TEntityClass[]
+	 */
+	public function findAll();
+
+	/**
+	 * @phpstan-param array<string, mixed> $criteria
+	 * @phpstan-param array<string, string>|null $orderBy
+	 * @phpstan-param int|null $limit
+	 * @phpstan-param int|null $offset
+	 * @phpstan-return TEntityClass[]
+	 */
+	public function findBy(array $criteria, ?array $orderBy = null, $limit = null, $offset = null);
+
+	/**
+	 * @phpstan-param array<string, mixed> $criteria The criteria.
+	 * @phpstan-param array<string, string>|null $orderBy
+	 * @phpstan-return TEntityClass|null
+	 */
+	public function findOneBy(array $criteria, array $orderBy = null);
+
+	/**
+	 * @phpstan-return class-string<TEntityClass>
+	 */
+	public function getClassName();
+
+	/**
+	 * @phpstan-return class-string<TEntityClass>
+	 */
+	protected function getEntityName();
+
+	/**
+	 * @param \Doctrine\Common\Collections\Criteria $criteria
+	 *
+	 * @return \Doctrine\Common\Collections\Collection
+	 *
+	 * @psalm-return \Doctrine\Common\Collections\Collection<int, TEntityClass>
+	 */
+	public function matching(Criteria $criteria);
+
+	/**
+     * @param literal-string      $alias
+     * @param literal-string|null $indexBy
+     *
+     * @return QueryBuilder
+     */
+    public function createQueryBuilder($alias, $indexBy = null);
+
+}
--- a/stubs/bleedingEdge/EntityRepository.stub
+++ b/stubs/bleedingEdge/EntityRepository.stub
@@ -64,11 +64,11 @@ class EntityRepository implements ObjectRepository
 	public function matching(Criteria $criteria);
 
 	/**
-     * @param literal-string      $alias
-     * @param literal-string|null $indexBy
-     *
-     * @return QueryBuilder
-     */
-    public function createQueryBuilder($alias, $indexBy = null);
+	 * @param literal-string      $alias
+	 * @param literal-string|null $indexBy
+	 *
+	 * @return QueryBuilder
+	 */
+	public function createQueryBuilder($alias, $indexBy = null);
 
 }
--- a/stubs/bleedingEdge/ORM/QueryBuilder.stub
+++ b/stubs/bleedingEdge/ORM/QueryBuilder.stub
@@ -2,6 +2,8 @@
 
 namespace Doctrine\ORM;
 
+use Doctrine\ORM\Query\Expr;
+
 class QueryBuilder
 {
 
--- a/stubs/bleedingEdge/ORM/QueryBuilder.stub
+++ b/stubs/bleedingEdge/ORM/QueryBuilder.stub
@@ -22,9 +22,25 @@ class QueryBuilder
 	}
 
 	/**
+	 * @return literal-string
+	 */
+	public function getRootAlias()
+	{
+
+	}
+
+	/**
 	 * @return list<literal-string>
 	 */
 	public function getRootAliases()
+	{
+
+	}
+
+	/**
+	 * @return list<literal-string>
+	 */
+	public function getAllAlias()
 	{
 
 	}
--- a/stubs/bleedingEdge/ORM/QueryBuilder.stub
+++ b/stubs/bleedingEdge/ORM/QueryBuilder.stub
@@ -21,6 +21,94 @@ class QueryBuilder
 
 	}
 
+    /**
+     * @param string                                                                                  $dqlPartName
+     * @psalm-param literal-string|object|list<literal-string>|array{join: array<int|string, object>} $dqlPart
+     * @param bool                                                                                    $append
+     *
+     * @return $this
+     */
+    public function add($dqlPartName, $dqlPart, $append = false)
+    {
+
+    }
+
+    /**
+     * @param literal-string|null $delete
+     * @param literal-string|null $alias
+     *
+     * @return $this
+     */
+    public function delete($delete = null, $alias = null)
+    {
+
+    }
+
+    /**
+     * @param literal-string|null $update
+     * @param literal-string|null $alias
+     *
+     * @return $this
+     */
+    public function update($update = null, $alias = null)
+    {
+
+    }
+
+    /**
+     * @param literal-string      $from
+     * @param literal-string      $alias
+     * @param literal-string|null $indexBy
+     *
+     * @return $this
+     */
+    public function from($from, $alias, $indexBy = null)
+    {
+
+    }
+
+    /**
+     * @param literal-string                                     $join
+     * @param literal-string                                     $alias
+     * @param Expr\Join::ON|Expr\Join::WITH|null                 $conditionType
+     * @param literal-string|Expr\Comparison|Expr\Composite|null $condition
+     * @param literal-string|null                                $indexBy
+     *
+     * @return $this
+     */
+    public function innerJoin($join, $alias, $conditionType = null, $condition = null, $indexBy = null)
+    {
+
+    }
+
+    /**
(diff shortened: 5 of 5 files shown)


craigfrancis:
Initial response for now... I don't think [getRootAliases()](https://github.com/doctrine/orm/blob/1dd2b449825df2ff7b2f4dddf06797c3a97cc1ab/lib/Doctrine/ORM/QueryBuilder.php#L452) or [getRootAlias()](https://github.com/doctrine/orm/blob/1dd2b449825df2ff7b2f4dddf06797c3a97cc1ab/lib/Doctrine/ORM/QueryBuilder.php#L426) is guaranteed to return `literal-string`, so I'm fairly sure it needs a dynamic return type.

It's very similar to the `Expr\Func` issue, where I need to work out how to dynamically return the type based on earlier input.

For example, [$qb->expr()->abs($x)](https://github.com/doctrine/orm/blob/2.12.x/lib/Doctrine/ORM/Query/Expr.php#L335) simply calls `return new Expr\Func('ABS', [$x])`, which is an object that can be used with methods like `->expr()->between()` and `->where()`, and because we aren't going to require `non-literal` strings for input ([reasons](https://github.com/phpstan/phpstan-doctrine/issues/294#issuecomment-1058189758)), I need to work out how the `__toString()` method in [Func](https://github.com/doctrine/orm/blob/1dd2b449825df2ff7b2f4dddf06797c3a97cc1ab/lib/Doctrine/ORM/Query/Expr/Func.php) can conditionally @return a `literal-string`.

VincentLanglet:
I just discovered what literal-string is today with the errors reported by PHPStan.

> Initial response for now... I don't think [getRootAliases()](https://github.com/doctrine/orm/blob/1dd2b449825df2ff7b2f4dddf06797c3a97cc1ab/lib/Doctrine/ORM/QueryBuilder.php#L452) or [getRootAlias()](https://github.com/doctrine/orm/blob/1dd2b449825df2ff7b2f4dddf06797c3a97cc1ab/lib/Doctrine/ORM/QueryBuilder.php#L426) is guaranteed to return literal-string, so I'm fairly sure it needs a dynamic return type.

Indeed, the getRootAliases might not always return a literal string, since
```
$qb = $em->createQueryBuilder()->select($userAlias)->from('User', $userAlias);
$qb->getRootAliases();
```
could be written.

But I don't see how you'll be able to write a dynamicReturnTypeExtension for this. You'll have to make the QueryBuilder class a generic with an array of all the alias, isn't it ? This seems kinda complicated ; and keep in mind that adding a generic param is creating an issue for people using both phpstan and psalm if psalm is not defining the generic too.

Couldn't the problem taken the opposite way ?
If you set a non-literal-string to the from alias or the join ones, you've already messed up the security of your query, so phpstan doesn't need to annoy you with the use of `getRootAliases` in your where conditions. I mean, why checking that you don't have any SQL injection in your WHERE clause, when you might already have one in your JOIN.

IMHO, we should consider that getRootAliases returns a literal-string.

Then, if we want to add an extra check it should be on the `from` method, asking for a `literal-string` for the alias and same for the joins. https://github.com/doctrine/orm/blob/1dd2b449825df2ff7b2f4dddf06797c3a97cc1ab/lib/Doctrine/ORM/QueryBuilder.php#L897 https://github.com/doctrine/orm/blob/1dd2b449825df2ff7b2f4dddf06797c3a97cc1ab/lib/Doctrine/ORM/QueryBuilder.php#L970

> because we aren't going to require non-literal strings for input (https://github.com/phpstan/phpstan-doctrine/issues/294#issuecomment-1058189758)

For the `from`, `join`, `leftJoin`, `innerJoin` aliases, it's different. You don't use them to create part of query and I don't see reason where it's fine to have string instead of literal-string.

craigfrancis:
Yep, that all seems reasonable, as those methods do feed into the SQL (well, DQL)... the idea is to check all inputs (WHERE/FROM/JOIN/etc), as you're right, they should all be written by the developer (or handled/escaped specially by the library).

Did you want to update your PR with your suggestion, so that `getRootAliases()`/`getRootAlias()` can return a `literal-string`?

Also, can you keep this behind the "bleedingEdge" flag (the separate stub file), so we don't accidentally break things for normal users :-)

I'm trying to work out how to complete [PR 324](https://github.com/phpstan/phpstan-doctrine/pull/324) at the moment, and probably won't have enough time this week (clients expecting me to do work for them, which just isn't fair). And thanks for your thoughts on the dynamic return types. As to the psalm implementation, I want to make sure I can get all of the checks working here first, then I can look at psalm, and then talk to the Doctrine developers.

VincentLanglet:
> Yep, that all seems reasonable, as those methods do feed into the SQL (well, DQL)... the idea is to check all inputs (WHERE/FROM/JOIN/etc), as you're right, they should all be written by the developer (or handled/escaped specially by the library).
> 
> Did you want to update your PR with your suggestion, so that `getRootAliases()`/`getRootAlias()` can return a `literal-string`?
> 
> Also, can you keep this behind the "bleedingEdge" flag (the separate stub file), so we don't accidentally break things for normal users :-)

Sure done.

> I'm trying to work out how to complete [PR 324](https://github.com/phpstan/phpstan-doctrine/pull/324) at the moment, and probably won't have enough time this week (clients expecting me to do work for them, which just isn't fair). And thanks for your thoughts on the dynamic return types. As to the psalm implementation, I want to make sure I can get all of the checks working here first, then I can look at psalm, and then talk to the Doctrine developers.

I didn't add the `literal-string` requirement for the FROM/JOIN/etc, since you prefer to wait for https://github.com/phpstan/phpstan-doctrine/pull/326. But I can do the PR too if you want.

VincentLanglet:
Can you confirm this is ok for you now @craigfrancis ?

craigfrancis:
Yes, and while it can be merged, I would like to have the inputs checked (on my to do list).

VincentLanglet:
> I would like to have the inputs checked (on my to do list).

What do you mean ? I can help if u want.

craigfrancis:
Notes for now...

Similar to the methods you've already identified (where I simply don't want these 3 methods saying they return a `literal-string`, that can be trusted, when they may not)...

---

The methods being updated by this PR (`getRootAlias()`, `getRootAliases()`, and `getAllAliases()`); they get their inputs from the private property `->_dqlParts`; in particular `->_dqlParts['from']` (in turn using the method `getAlias()`).

When looking at how `->_dqlParts` is updated, we can ignore `resetDQLPart()` and `__clone()`, but we do need to specify the `literal-string` input parameter types for:

- [QueryBuilder->add()](https://github.com/doctrine/orm/blob/a9309d748b9bab8185b593dcdfed18f1db8104b3/lib/Doctrine/ORM/QueryBuilder.php#L702)

And, because PHPStan does not check usage within the library, we need to specify the `literal-string` input parameter types for everything that uses `->add()` (specifically 'from', or with a variable that could be 'from'), i.e.

- [QueryBuilder->delete()](https://github.com/doctrine/orm/blob/a9309d748b9bab8185b593dcdfed18f1db8104b3/lib/Doctrine/ORM/QueryBuilder.php#L843)
- [QueryBuilder->update()](https://github.com/doctrine/orm/blob/a9309d748b9bab8185b593dcdfed18f1db8104b3/lib/Doctrine/ORM/QueryBuilder.php#L870)
- [QueryBuilder->from()](https://github.com/doctrine/orm/blob/a9309d748b9bab8185b593dcdfed18f1db8104b3/lib/Doctrine/ORM/QueryBuilder.php#L897)

I cannot see `QueryBuilder->add()` being used elsewhere in the ORM library, and checking these 3 other functions, I can only find:

- [EntityRepository->createQueryBuilder()](https://github.com/doctrine/orm/blob/a9309d748b9bab8185b593dcdfed18f1db8104b3/lib/Doctrine/ORM/EntityRepository.php#L86)

Just to note `->conn->delete()` is for the DBAL library ([source](https://github.com/doctrine/dbal/blob/ea58a77060da2c43f57996141873d7b6098c7446/src/Connection.php#L631), to look at later); and `->persister->delete()` doesn't seem to use the QueryBuilder.

And, small tangent, regarding `Expr\From`, which we won't consider at the moment (because we are not looking at developers initiating these themselves), the inputs might (should?) be a `literal-string`, and like the [Expr\Func](https://github.com/phpstan/phpstan-doctrine/pull/327#issuecomment-1110038672) problem, I cannot currently think of a way to dynamically return the type for methods like [From->getAlias()](https://github.com/doctrine/orm/blob/a9309d748b9bab8185b593dcdfed18f1db8104b3/lib/Doctrine/ORM/Query/Expr/From.php#L46) when they are called later.

---

Also `->getAllAliases()` reads the private property keys in `->joinRootAliases`, which is set via the private method `->findRootAlias()`, which is called by:

- [QueryBuilder->innerJoin()](https://github.com/doctrine/orm/blob/a9309d748b9bab8185b593dcdfed18f1db8104b3/lib/Doctrine/ORM/QueryBuilder.php#L997)
- [QueryBuilder->leftJoin()](https://github.com/doctrine/orm/blob/a9309d748b9bab8185b593dcdfed18f1db8104b3/lib/Doctrine/ORM/QueryBuilder.php#L1038)
- [QueryBuilder->join()](https://github.com/doctrine/orm/blob/a9309d748b9bab8185b593dcdfed18f1db8104b3/lib/Doctrine/ORM/QueryBuilder.php#L970)

VincentLanglet:
> Similar to the methods you've already identified (where I simply don't want these 3 methods saying they return a literal-string, that can be trusted, when they may not)...

I'm confused...
Should I update in this PR the phpdoc of all the method you listed ? @craigfrancis 

Or do you want me to do it in another PR ?

VincentLanglet:
> Should I update in this PR the phpdoc of all the method you listed ?

I updated the stubs with all the methods listed in order to check a literal-string is passed.

craigfrancis:
Thank you @VincentLanglet, I'll try to do some checking later this week (sorry, busy with other work/life things).

craigfrancis:
Hi @VincentLanglet, thank you for doing these, I really appreciate it.

Just having a quick look at the changes, I've got a couple of small notes/questions:

I think PHPStan generally uses tabs for indentation, which would also avoid the whitespace change in `stubs/ORM/QueryBuilder.stub`.

### EntityRepository.stub

In `/extension.neon`, under `stubFiles`, [line 27](https://github.com/phpstan/phpstan-doctrine/blob/dd53db81bc1f877248de3819321d84126a9e9f65/extension.neon#L27) includes `stubs/EntityRepository.stub`. I think that should be removed, as that file or `stubs/bleedingEdge/EntityRepository.stub` is included via `StubFilesExtensionLoader` (depending on `bleedingEdge`).

Don't do this... I'm just wondering if these two files should be moved into their respective `/ORM/` sub-folders, to match the Doctrine ORM folder structure. I don't think it should be moved as I suspect other things depend on that current location, and other files like `EntityManager.stub` are there as well.

Will also note... the `_entityName` property, and methods `find()` to `matching()`, you've copied them here to preserve the existing functionality from `/stubs/EntityRepository.stub` (i.e. the file used when not `bleedingEdge`). Ignoring this duplicate code, we're only adding [createQueryBuilder()](https://github.com/doctrine/orm/blob/a9309d748b9bab8185b593dcdfed18f1db8104b3/lib/Doctrine/ORM/EntityRepository.php#L78).

### QueryBuilder.stub

The `add()` method notes `@psalm-param`, assuming this is supported by PHPStan, is it worth changing it to `@param`?

VincentLanglet:
> I think PHPStan generally uses tabs for indentation, which would also avoid the whitespace change in `stubs/ORM/QueryBuilder.stub`.

Changed

> ### EntityRepository.stub
> In `/extension.neon`, under `stubFiles`, [line 27](https://github.com/phpstan/phpstan-doctrine/blob/dd53db81bc1f877248de3819321d84126a9e9f65/extension.neon#L27) includes `stubs/EntityRepository.stub`. I think that should be removed, as that file or `stubs/bleedingEdge/EntityRepository.stub` is included via `StubFilesExtensionLoader` (depending on `bleedingEdge`).

Removed.

> Don't do this... I'm just wondering if these two files should be moved into their respective `/ORM/` sub-folders, to match the Doctrine ORM folder structure. I don't think it should be moved as I suspect other things depend on that current location, and other files like `EntityManager.stub` are there as well.

I prefer to delay this to another pr.
 
> Will also note... the `_entityName` property, and methods `find()` to `matching()`, you've copied them here to preserve the existing functionality from `/stubs/EntityRepository.stub` (i.e. the file used when not `bleedingEdge`). Ignoring this duplicate code, we're only adding [createQueryBuilder()](https://github.com/doctrine/orm/blob/a9309d748b9bab8185b593dcdfed18f1db8104b3/lib/Doctrine/ORM/EntityRepository.php#L78).

Since only one of the two stubs is loaded, I have to duplicate the code.

> ### QueryBuilder.stub
> The `add()` method notes `@psalm-param`, assuming this is supported by PHPStan, is it worth changing it to `@param`?

Fixed

VincentLanglet:
@ondrejmirtes Removing `stubs/EntityRepository.stub` from the extension.neon file is creating a lot of errors.
But I added this files to the `StubFilesExtensionLoader`. Isn't that enough ?

ondrejmirtes:
Not sure why it's unhappy, you need to debug that.

ondrejmirtes:
I realized the reason of failure - https://github.com/phpstan/phpstan-src/blob/1.7.x/src/PhpDoc/StubSourceLocatorFactory.php reads just the `stubFiles` array, not the files from extensions. Needs a systematic solution though.

ondrejmirtes:
Done: https://github.com/phpstan/phpstan-src/commit/0ecddcca760fc220b1a907d921b2d2c02809551f

If you require PHPStan `^1.7`, it's gonna work without the last commit too.

VincentLanglet:
> Done: [phpstan/phpstan-src@0ecddcc](https://github.com/phpstan/phpstan-src/commit/0ecddcca760fc220b1a907d921b2d2c02809551f)
> 
> If you require PHPStan `^1.7`, it's gonna work without the last commit too.

Nice, I was trying to add back the stub in the config and see if they were correctly overridden by the bleedingEdge one.
But your fix is way better.

I added the `^1.7` requirement.

VincentLanglet:
I fixed the conflict @ondrejmirtes so I think this is ready to be merged. :)

ondrejmirtes:
For reference the differences between normal/bleedingEdge stubs. Do you agree with them @craigfrancis?

```diff
--- stubs/EntityRepository.stub	2021-11-02 14:29:16.000000000 +0100
+++ stubs/bleedingEdge/EntityRepository.stub	2022-05-23 09:15:27.000000000 +0200
@@ -63,4 +63,12 @@
 	 */
 	public function matching(Criteria $criteria);

+	/**
+	 * @param literal-string      $alias
+	 * @param literal-string|null $indexBy
+	 *
+	 * @return QueryBuilder
+	 */
+	public function createQueryBuilder($alias, $indexBy = null);
+
 }
```

```diff
--- stubs/ORM/QueryBuilder.stub	2022-05-23 09:15:27.000000000 +0200
+++ stubs/bleedingEdge/ORM/QueryBuilder.stub	2022-05-23 09:15:27.000000000 +0200
@@ -2,10 +2,19 @@

 namespace Doctrine\ORM;

+use Doctrine\ORM\Query\Expr;
+
 class QueryBuilder
 {

 	/**
+	 * @return Query<mixed>
+	 */
+	public function getQuery()
+	{
+	}
+
+	/**
 	 * @param \Doctrine\Common\Collections\ArrayCollection<array-key, mixed>|array<mixed> $parameters
 	 * @return static
 	 */
@@ -15,10 +24,124 @@
 	}

 	/**
-	 * @return Query<mixed>
+	 * @param string                                                                            $dqlPartName
+	 * @param literal-string|object|list<literal-string>|array{join: array<int|string, object>} $dqlPart
+	 * @param bool                                                                              $append
+	 *
+	 * @return $this
 	 */
-	public function getQuery()
+	public function add($dqlPartName, $dqlPart, $append = false)
+	{
+
+	}
+
+	/**
+	 * @param literal-string|null $delete
+	 * @param literal-string|null $alias
+	 *
+	 * @return $this
+	 */
+	public function delete($delete = null, $alias = null)
+	{
+
+	}
+
+	/**
+	 * @param literal-string|null $update
+	 * @param literal-string|null $alias
+	 *
+	 * @return $this
+	 */
+	public function update($update = null, $alias = null)
+	{
+
+	}
+
+	/**
+	 * @param literal-string      $from
+	 * @param literal-string      $alias
+	 * @param literal-string|null $indexBy
+	 *
+	 * @return $this
+	 */
+	public function from($from, $alias, $indexBy = null)
 	{
+
+	}
+
+	/**
+	 * @param literal-string                                     $join
+	 * @param literal-string                                     $alias
+	 * @param Expr\Join::ON|Expr\Join::WITH|null                 $conditionType
+	 * @param literal-string|Expr\Comparison|Expr\Composite|null $condition
+	 * @param literal-string|null                                $indexBy
+	 *
+	 * @return $this
+	 */
+	public function innerJoin($join, $alias, $conditionType = null, $condition = null, $indexBy = null)
+	{
+
+	}
+
+	/**
+	 * @param literal-string                                     $join
+	 * @param literal-string                                     $alias
+	 * @param Expr\Join::ON|Expr\Join::WITH|null                 $conditionType
+	 * @param literal-string|Expr\Comparison|Expr\Composite|null $condition
+	 * @param literal-string|null                                $indexBy
+	 *
+	 * @return $this
+	 */
+	public function leftJoin($join, $alias, $conditionType = null, $condition = null, $indexBy = null)
+	{
+
+	}
+
+	/**
+	 * @param literal-string                                     $join
+	 * @param literal-string                                     $alias
+	 * @param Expr\Join::ON|Expr\Join::WITH|null                 $conditionType
+	 * @param literal-string|Expr\Comparison|Expr\Composite|null $condition
+	 * @param literal-string|null                                $indexBy
+	 *
+	 * @return $this
+	 */
+	public function join($join, $alias, $conditionType = null, $condition = null, $indexBy = null)
+	{
+
+	}
+
+	/**
+	 * @return literal-string
+	 */
+	public function getRootAlias()
+	{
+
+	}
+
+	/**
+	 * @return list<literal-string>
+	 */
+	public function getRootAliases()
+	{
+
+	}
+
+	/**
+	 * @return list<literal-string>
+	 */
+	public function getAllAlias()
+	{
+
+	}
+
+	/**
+	 * @param literal-string|object|array<mixed> $predicates
+	 * @return $this
+	 */
+	public function where($predicates)
+	{
+
 	}

 }
```

craigfrancis:
Yep, I've checked them, and they seem to be correct.

---

This might cause a few more issues to crop up, as we are doing a bit more than the `where()` method, but this is behind the `bleedingEdge` flag, and we will need to do them at some point.

I'd also note that your diff for `EntityRepository.stub` confirms my checks, where the patch implies a much bigger change, but most of the methods are simply being copied over from the non-`bleedingEdge` version, where it is only the `createQueryBuilder()` method that is getting its two `@param`'s updated.

ondrejmirtes:
Alright, thanks :)

PR #329, opened 2022-05-05
add stub for EntityManager::getClassMetadata

Khartir:
Copied from https://github.com/doctrine/orm/blob/8291a7f09b12d14783ed6537b4586583d155869e/lib/Doctrine/ORM/EntityManagerInterface.php#L337-L344

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultTypeWalker.php
+++ b/src/Type/Doctrine/Query/QueryResultTypeWalker.php
@@ -244,8 +244,10 @@ class QueryResultTypeWalker extends SqlWalker
 
 			case AST\PathExpression::TYPE_SINGLE_VALUED_ASSOCIATION:
 				if (isset($class->associationMappings[$fieldName]['inherited'])) {
-					assert(is_string($class->associationMappings[$fieldName]['inherited']));
-					$class = $this->em->getClassMetadata($class->associationMappings[$fieldName]['inherited']);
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
What do these changes achieve please? Can you please add a test that fails without these changes? Thanks.

Khartir:
The problem is code like this:
```php
/**
 * @template T of object
 * @template-extends EntityRepository<T>
 */
class IntermediateRepository extends EntityRepository
{
	/**
	 * @param ClassMetadata<T> $class
	 */
	public function __construct(EntityManagerInterface $em, ClassMetadata $class)
	{
		parent::__construct($em, $class);
	}
}

/**
 * @extends IntermediateRepository<MyEntity>
 */
class MyRepository extends IntermediateRepository
{
	public function __construct(
		EntityManagerInterface $manager
	) {
		parent::__construct($manager, $manager->getClassMetadata(MyEntity::class));
	}
}
```
This fails with `Parameter #2 $class of method IntermediateRepository<MyEntity>::__construct() expects Doctrine\ORM\Mapping\ClassMetadata<MyEntity>, Doctrine\ORM\Mapping\ClassMetadata<object> given.`

With the extension of the stub this now works, but I'm having trouble writing a test for this.

Khartir:
OK writing a test for this wasn't that hard after all, I was just missing an import. :laughing:

ondrejmirtes:
Thank you!

PR #336, opened 2022-05-27
Fixing handling of boolean literals

### LINKED ISSUE #311: `CASE WHEN EXISTS(...) THEN TRUE ELSE FALSE END AS my_column` is not inferred as boolean
- Create an ORM QueryBuilder
- Call `->select('CASE WHEN EXISTS(SELECT 1) THEN TRUE ELSE FALSE END AS my_column')`
- Call `->getQuery()->getResult()`

See the type of `my_column` inferred as `0|'0'`

arnaud-lb:
Fixes #311


ondrejmirtes:
Thank you!

PR #339, opened 2022-05-30
Add failing tests for `MissingReadOnlyPropertyAssignRule`

herndlm:
Closes https://github.com/phpstan/phpstan/issues/7337

Looks like the problem is actually here in `PropertiesExtension` (which is now used at least). https://github.com/phpstan/phpstan-doctrine/blob/1.3.6/src/Rules/Doctrine/ORM/PropertiesExtension.php#L85 returns always `false`, because `$metadata->isReadOnly` (which marks if this is a Doctrine readonly entity?) is already false.

Does this need another early exit dealing with native readonly or smth like that? I gave it a shot, but don't know much about this here. Feel free to finish it :)

### DIFF
--- a/src/Rules/Doctrine/ORM/PropertiesExtension.php
+++ b/src/Rules/Doctrine/ORM/PropertiesExtension.php
@@ -44,34 +44,15 @@ class PropertiesExtension implements ReadWritePropertiesExtension
 			return false;
 		}
 
-		if ($this->isGeneratedIdentifier($metadata, $propertyName)) {
-			return true;
-		}
-
 		if ($metadata->isReadOnly && !$declaringClass->hasConstructor()) {
 			return true;
-		}
-
-		if ($metadata->isIdentifierNatural()) {
-			return false;
 		}
 
 		if ($metadata->versionField === $propertyName) {
 			return true;
 		}
 
-		try {
-			$identifiers = $metadata->getIdentifierFieldNames();
-		} catch (Throwable $e) {
-			$mappingException = 'Doctrine\ORM\Mapping\MappingException';
-			if (!$e instanceof $mappingException) {
-				throw $e;
-			}
-
-			return false;
-		}
-
-		return in_array($propertyName, $identifiers, true);
+		return $this->isGeneratedIdentifier($metadata, $propertyName);
 	}
 
 	public function isInitialized(PropertyReflection $property, string $propertyName): bool
@@ -96,11 +77,20 @@ class PropertiesExtension implements ReadWritePropertiesExtension
 
 	private function isGeneratedIdentifier(ClassMetadataInfo $metadata, string $propertyName): bool
 	{
-		if (!in_array($propertyName, $metadata->getIdentifierFieldNames(), true)) {
+		if ($metadata->isIdentifierNatural()) {
 			return false;
 		}
 
-		return $metadata->generatorType !== ClassMetadataInfo::GENERATOR_TYPE_NONE;
+		try {
+			return in_array($propertyName, $metadata->getIdentifierFieldNames(), true);
+		} catch (Throwable $e) {
+			$mappingException = 'Doctrine\ORM\Mapping\MappingException';
+			if (!$e instanceof $mappingException) {
+				throw $e;
+			}
+
+			return false;
+		}
 	}
 
 }
--- a/src/Rules/Doctrine/ORM/PropertiesExtension.php
+++ b/src/Rules/Doctrine/ORM/PropertiesExtension.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Rules\Doctrine\ORM;
 
+use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use PHPStan\Reflection\PropertyReflection;
 use PHPStan\Rules\Properties\ReadWritePropertiesExtension;
 use PHPStan\Type\Doctrine\ObjectMetadataResolver;
@@ -43,6 +44,10 @@ class PropertiesExtension implements ReadWritePropertiesExtension
 			return false;
 		}
 
+		if ($this->isGeneratedIdentifier($metadata, $propertyName)) {
+			return true;
+		}
+
 		if ($metadata->isReadOnly && !$declaringClass->hasConstructor()) {
 			return true;
 		}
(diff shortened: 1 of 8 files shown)


src/Rules/Doctrine/ORM/PropertiesExtension.php
```diff
@@ -82,7 +83,15 @@ public function isInitialized(PropertyReflection $property, string $propertyName
 			return false;
 		}
 
-		return $metadata->isReadOnly && !$declaringClass->hasConstructor();
+		if ($declaringClass->hasConstructor()) {
+			return false;
+		}
+
+		if ($property instanceof PhpPropertyReflection && ($property->isReadOnly() || $property->isReadOnlyByPhpDoc())) {
```
herndlm: what if this is an instance of another class? hmm

herndlm: also: the same or a similar thing might be needed in `isAlwaysWritten` too

ondrejmirtes: This logic seems wrong. It's not that `readonly` makes the property initialized, it's that `#[ORM\GeneratedValue]` / `@ORM\GeneratedValue` makes it initialized/always written.

You should look at `$metadata->generatorType` and `$metadata->getIdentifierFieldNames()` IMHO.

ondrejmirtes (changes requested):
Also the PropertieExtension is completely untested - when I initially wrote it, I was happy it removed all false-positives from a project I test each release with. (And I thought I'll never have to touch it again, haha.)

Can you write a few tests for the current behaviour too? Especially - if the whole entity is marked as readonly by Doctrine metadata + doesn't have a constructor, it means that the entity is inserted into database by other means and the ORM is used only for reading it -> properties should be always initialized. But if it's not readonly / has a constructor, it means the properties should not be initialized.

src/Rules/Doctrine/ORM/PropertiesExtension.php
herndlm: thx for providing the details! I was actually half-blindly modifying this without knowing much of the surroundings which was a bad idea..

herndlm:
ok, that looks better I think. The `isAlwaysWritten` and `isAlwaysRead` are still not tested though. Any good test case ideas? UPDATE: ah, that's another rule.. let me quickly add some tests for that too

Or anything else you need? I can add it here or in follow-ups, whatever is preferred.

herndlm:
Most of `PropertiesExtension` should be covered now, sorry for the many force-pushes 😅

src/Rules/Doctrine/ORM/PropertiesExtension.php
```diff
@@ -43,6 +44,10 @@ public function isAlwaysWritten(PropertyReflection $property, string $propertyNa
 			return false;
 		}
 
+		if ($this->isGeneratedIdentifier($metadata, $propertyName)) {
```
ondrejmirtes: This exact thing is already checked in this method. I believe if you remove these lines, all tests should still pass.

src/Rules/Doctrine/ORM/PropertiesExtension.php
```diff
@@ -82,7 +87,20 @@ public function isInitialized(PropertyReflection $property, string $propertyName
...
 		return $metadata->isReadOnly && !$declaringClass->hasConstructor();
 	}
 
+	private function isGeneratedIdentifier(ClassMetadataInfo $metadata, string $propertyName): bool
+	{
+		if (!in_array($propertyName, $metadata->getIdentifierFieldNames(), true)) {
+			return false;
+		}
+
+		return $metadata->generatorType !== ClassMetadataInfo::GENERATOR_TYPE_NONE;
```
ondrejmirtes: You can call `$metadata->isIdentifierNatural()` :)

ondrejmirtes (changes requested):

tests/Rules/Properties/data/missing-readonly-property-assign-phpdoc.php
```diff
@@ -0,0 +1,73 @@
...
+
+	/**
+	 * @ORM\Column
+	 * @readonly
+	 */
+	private int $doubleAssigned;
+
+	public function __construct(int $doubleAssigned)
+	{
+		$this->doubleAssigned = $doubleAssigned;
```
ondrejmirtes, line 35: I just got an idea for a rule improvement in phpstan/src. If an extension says `isInitialized=true` about a property, even the first assignment should say "readonly property is already assigned" :)

src/Rules/Doctrine/ORM/PropertiesExtension.php
herndlm: oh, good find! yeah, they still pass

tests/Rules/Properties/data/missing-readonly-property-assign-phpdoc.php
herndlm: interesting, I guess I can take a look at this in phpstan-src

herndlm:
based on your comments I found another simplification that I did. Not sure why it is there exactly, but I also took over that `getIdentifierFieldNames` exception handling. do you know still why this was added?

src/Rules/Doctrine/ORM/PropertiesExtension.php
```diff
@@ -47,26 +48,11 @@ public function isAlwaysWritten(PropertyReflection $property, string $propertyNa
 			return true;
 		}
 
-		if ($metadata->isIdentifierNatural()) {
-			return false;
-		}
-
 		if ($metadata->versionField === $propertyName) {
 			return true;
 		}
```
herndlm, line 53: hmm this looks risky even though it is covered below..

herndlm: ah ok, that just marks the versioning column in doctrine. I _guess_ it should be fine then since this is not related to the ID

ondrejmirtes (approved):

ondrejmirtes:
@herndlm Because ClassMetadata can throw various exceptions and I don't want PHPStan crashing on that. There's a new bleedingEdge rule (https://github.com/phpstan/phpstan-doctrine/blob/1.3.x/src/Rules/Doctrine/ORM/EntityMappingExceptionRule.php) that's gonna tell you if Doctrine doesn't like your mappings.

ondrejmirtes:
Thank you very much :)

herndlm:
> @herndlm Because ClassMetadata can throw various exceptions and I don't want PHPStan crashing on that. There's a new bleedingEdge rule (https://github.com/phpstan/phpstan-doctrine/blob/1.3.x/src/Rules/Doctrine/ORM/EntityMappingExceptionRule.php) that's gonna tell you if Doctrine doesn't like your mappings.

thx, makes sense. I assumed something like that, but didn't see the exception in the class/interface and wasn't sure. but I didn't look too long, maybe I just missed it ;)

PR #340, opened 2022-05-31
Fix ObjectRepository template covariance

fluffycondor:
According to https://github.com/doctrine/persistence/pull/292


ondrejmirtes:
Thank you.

PR #348, opened 2022-07-03
Enum support in query type inference

### LINKED ISSUE #344: Enum type not correctly detected in DQL query result
Consider this setup:

```php
#[ORM\Entity]
class ProductCountry extends Entity
{

	#[ORM\Id]
	#[ORM\GeneratedValue]
	#[ORM\Column]
	private ?int $id = null;

	#[ORM\Column]
	#[Set(visibility: 'private')]
	private Country $country;
}

enum Country: string
{
    case CZE = 'cze';
    //....
}
```

when doing DQL to country from this entity 
```php
public function getUsedCountriesInTravelProducts(): array
{
		->select('pc.country')
		->from(ProductCountry::class, 'pc')
		// .. some conditions
		->addGroupBy('pc.country')
		->getQuery();
	return array_map(static fn (array $result): Country => $result['country'], $query->getResult());
}
```

phpstan reports error ` Anonymous function should return Country but returns string` but in reality, `$result['country']` is properly hydrated to enum

arnaud-lb:
Fixes #344

## Doctrine behavior

For reference:

Selecting an enum column directly returns an enum case (an instance of the enum class).

Using an enum column in an expression returns the raw value returned by the database driver. This is consistent with other column types (e.g. using a datetime column in an expression returns a string).

## phpstan-doctrine behavior

In this PR I add support for enums as described above.

Since we know all the possible values of an enum, we can refine the database column type to a union of these values.

### Examples

With the following classes:

``` php
#[Entity]
class X {
    #[Column]
    public Foo $f;
}

enum Foo: string {
    case A = 'a';
    case B = 'b';
}
```

``` sql
--  Returns results of type array{f: Foo}
SELECT x.f FROM X

-- Returns results of type array{1: 'a'|'b'|''}
SELECT COALESCE(x.f, '') FROM X
```

## Doctrine 2.11 issues

Enum support was introduced in Doctrine 2.11, but there was issues such as https://github.com/doctrine/orm/issues/9622.

Here I add support for enums as implemented in Doctrine >= 2.12.

We have multiple options to handle Doctrine 2.11:
- Do nothing: Ignore Doctrine 2.11 issues. For users with Doctrine 2.11, type inference will be lying.
- Add special cases for 2.11
- Conflict with 2.11

I chose the third option here. I think that option two would be acceptable as well, but this adds complexity just for a minor version. WDYT ?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,7 @@
 		"doctrine/persistence": "^1.3.8 || ^2.2.1",
 		"nesbot/carbon": "^2.49",
 		"nikic/php-parser": "^4.13.2",
+		"ocramius/package-versions": "*",
 		"php-parallel-lint/php-parallel-lint": "^1.2",
 		"phpstan/phpstan-phpunit": "^1.0",
 		"phpstan/phpstan-strict-rules": "^1.0",
@@ -41,7 +42,8 @@
 		},
 		"sort-packages": true,
 		"allow-plugins": {
-			"composer/package-versions-deprecated": true
+			"composer/package-versions-deprecated": true,
+			"ocramius/package-versions": true
 		}
 	},
 	"extra": {
--- a/composer.json
+++ b/composer.json
@@ -39,11 +39,7 @@
 			"ext-mongo": "1.12",
 			"ext-mongodb": "1.6.16"
 		},
-		"sort-packages": true,
-		"allow-plugins": {
-			"composer/package-versions-deprecated": true,
-			"ocramius/package-versions": true
-		}
+		"sort-packages": true
 	},
 	"extra": {
 		"phpstan": {
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,6 @@
 		"doctrine/persistence": "^1.3.8 || ^2.2.1",
 		"nesbot/carbon": "^2.49",
 		"nikic/php-parser": "^4.13.2",
-		"ocramius/package-versions": "*",
 		"php-parallel-lint/php-parallel-lint": "^1.2",
 		"phpstan/phpstan-phpunit": "^1.0",
 		"phpstan/phpstan-strict-rules": "^1.0",
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -17,6 +17,10 @@ parameters:
 
 	reportUnmatchedIgnoredErrors: false
 
+	bootstrapFiles:
+		- stubs/runtime/Enum/UnitEnum.php
+		- stubs/runtime/Enum/BackedEnum.php
+
 	ignoreErrors:
 		-
 			message: '~^Variable method call on Doctrine\\ORM\\QueryBuilder~'
--- a/stubs/runtime/Enum/BackedEnum.php
+++ b/stubs/runtime/Enum/BackedEnum.php
@@ -1,0 +1,22 @@
+<?php
+
+if (\PHP_VERSION_ID < 80100) {
+	if (interface_exists('BackedEnum', false)) {
+		return;
+	}
+
+	interface BackedEnum extends UnitEnum
+	{
+		/**
+		 * @param int|string $value
+		 * @return static
+		 */
+		public static function from($value);
+
+		/**
+		 * @param int|string $value
+		 * @return ?static
+		 */
+		public static function tryFrom($value);
+	}
+}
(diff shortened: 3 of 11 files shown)


ondrejmirtes:
About these options:

We have multiple options to handle Doctrine 2.11:

1) Do nothing: Ignore Doctrine 2.11 issues. For users with Doctrine 2.11, type inference will be lying.
2) Add special cases for 2.11
3) Conflict with 2.11

I think option 1) is the best. Conflict is too aggressive, if the user isn't using enums at all, the conflict wouldn't make sense for them 😊 So I think we can ignore the issue, it's not our bug to fix 😊

stubs/runtime/Enum/BackedEnum.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+if (\PHP_VERSION_ID < 80100) {
+	if (interface_exists('BackedEnum', false)) {
+		return;
+	}
+
+	interface BackedEnum extends UnitEnum
```
arnaud-lb, line 8: File copied from https://github.com/phpstan/phpstan-src/blob/07e341daedc018cf60bd7e8827bb59ae24566bee/stubs/runtime/Enum/BackedEnum.php

phpstan.neon
```diff
@@ -17,6 +17,10 @@ parameters:
 
 	reportUnmatchedIgnoredErrors: false
 
+	bootstrapFiles:
+		- stubs/runtime/Enum/UnitEnum.php
+		- stubs/runtime/Enum/BackedEnum.php
```
arnaud-lb, line 22: This makes supporting PHP versions bellow 8.1 much easier

Specifically, this allows code such as these to pass analysis:
```
is_subclass_of($className, BackedEnum::class)
```
```
/** @return class-string<BackedEnum> */
```

tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
```diff
@@ -208,1126 +229,1247 @@ public function test(Type $expectedType, string $dql, ?string $expectedException
 	}
 
 	/**
-	 * @return array<array-key,array{Type,string,2?:string}>
+	 * @return iterable<string,array{Type,string,2?:string}>
 	 */
-	public function getTestData(): array
+	public function getTestData(): iterable
```
arnaud-lb, line 234: Changed this data provider to `yield` test cases so that I could return some cases conditionally.

Best reviewed with `?w=1` :D

tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
```diff
@@ -208,1126 +229,1247 @@ public function test(Type $expectedType, string $dql, ?string $expectedException
...
+			',
+		];
+
+		if (property_exists(Column::class, 'enumType') && PHP_VERSION_ID >= 80100) {
+			assert(class_exists(StringEnum::class));
+			assert(class_exists(IntEnum::class));
+
+			// https://github.com/doctrine/orm/issues/9622
+			if (!$this->isDoctrine211()) {
+				yield 'enum' => [
```
arnaud-lb, line 516: The PHP 8.1 build with lowest dependencies happens to run Doctrine 2.11

composer.json
```diff
@@ -27,6 +27,7 @@
 		"doctrine/persistence": "^1.3.8 || ^2.2.1",
 		"nesbot/carbon": "^2.49",
 		"nikic/php-parser": "^4.13.2",
+		"ocramius/package-versions": "*",
```
arnaud-lb: We need this to skip some test when running Doctrine 2.11

phpstan.neon
ondrejmirtes: Would be nice to ignore these runtime stubs in `.gitattributes` :)

composer.json
ondrejmirtes: There's now a native Composer feature for this so we don't need a package I think? https://getcomposer.org/doc/07-runtime.md

ondrejmirtes (changes requested):
Otherwise 👍 :)

composer.json
arnaud-lb: Great, didn't know that. I've updated the PR to use this instead.

phpstan.neon
arnaud-lb: Ok, done :+1:

composer.json
ondrejmirtes: We can remove this from composer.json now? :)

arnaud-lb: Ah, gone too fast and forgot this. I've removed it now

ondrejmirtes:
Thank you!

PR #351, opened 2022-07-06
phpcs: Disallow trailing commas in function calls, declarations, use list

arnaud-lb:
I'm always getting trailing comma errors when submitting PRs on phpstan-doctrine because of the PHP 7.2 support.

It would help a lot if `make cs` would detect these problems.

This PR adds 3 rules to phpcs to do exactly this.

Ideally `make lint` would detect these problems instead, but making it run with PHP 7.2 on everyone's machine is much more involved.


ondrejmirtes:
Thank you.

PR #352, opened 2022-07-18
Use a @template for countDistinct

craigfrancis:
Not complete.

The tests that use a non-`literal-string` for parameters 2 and 3, incorrectly return a `literal-string`.

---

These tests should work ([results](https://phpstan.org/r/5b70ac59-4e91-4dd5-ac7d-b63eec30cef8)).

I suspect the original function definition, which implies only 1 parameter is used ([source](https://github.com/doctrine/orm/blob/ab4844b82af9b18cdff20a4d14dc1aeb4ddd08da/lib/Doctrine/ORM/Query/Expr.php#L263)):

    public function countDistinct($x)
    {
        return 'COUNT(DISTINCT ' . implode(', ', func_get_args()) . ')';
    }

Maybe the Stub file should be overriding this? Because I suspect the `@template` only considers the first parameter in this case, even though the Stub tries to correct it ([source](https://github.com/craigfrancis/phpstan-doctrine/blob/issue-294/stubs/ORM/Query/Expr.stub#L39)):

	/**
	 * @template T of string
	 * @param T ...$x
	 * @return (T is literal-string ? literal-string : string)
	 */
	public function countDistinct(...$x) {
	}

### DIFF
--- a/stubs/ORM/Query/Expr.stub
+++ b/stubs/ORM/Query/Expr.stub
@@ -31,4 +31,12 @@ class Expr
 	{
 	}
 
+	/**
+	 * @template T of string
+	 * @param T ...$x
+	 * @return (T is literal-string ? literal-string : string)
+	 */
+	public function countDistinct(...$x) {
+	}
+
 }
--- a/tests/Type/Doctrine/data/QueryResult/expressionBuilderGetQuery.php
+++ b/tests/Type/Doctrine/data/QueryResult/expressionBuilderGetQuery.php
@@ -67,6 +67,44 @@ class ExpressionBuilderGetQuery
 		assertType('string', $result);
 	}
 
+	public function countDistinctLiteralString(EntityManagerInterface $em): void
+	{
+		$result = $em->createQueryBuilder()->expr()->countDistinct('A', 'B', 'C');
+		assertType("'COUNT(DISTINCT A, B, C)'", $result); // A ConstantStringType isLiteralString
+	}
+
+	public function countDistinctNonLiteralString1(EntityManagerInterface $em): void
+	{
+		$value = $this->nonLiteralString('A');
+		$result = $em->createQueryBuilder()->expr()->countDistinct($value);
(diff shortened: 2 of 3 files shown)


craigfrancis:
I have opened [PR #9911](https://github.com/doctrine/orm/pull/9911) on [doctrine/orm](https://github.com/doctrine/orm/), as it should be fixed there as well.

ondrejmirtes:
Stub files are only for overriding PHPDocs, not variadic-ness of parameters :)

craigfrancis:
Yep, thanks Ondřej, hopefully Doctrine will accept my PR :-)

craigfrancis:
My PR to fix the `Expr::countDistinct()` arguments has been accepted & merged for Doctrine ORM 3.0.

But 3.0 doesen't seem to have an expected release date. There are "what to expect" type articles from 2019, and it looks like they are working on 2.12 (current) and 2.13 (dev)... so I doubt this will be resolved any time soon.

Should I:

1. Leave this PR here until `phpstan-doctrine` starts using `doctrine-orm` 3.0 (could be a long wait).
2. Temporarily comment out the 2 tests that check the 2nd and 3rd arguments, `countDistinctNonLiteralString3` and `countDistinctNonLiteralString4` (this PR kinda works, but will incorrectly say something is a `literal-string` even if any of the 2nd or later arguments are provided with an unsafe `string`; so it won't identify mistakes until 3.0 is in use, but will allow `Expr::countDistinct()` to return a `literal-string`).
3. Close this PR, and try again when 3.0 is in use (could be a while)?

ondrejmirtes:
Feel free to do option number 2) so that we have benefits faster :)

craigfrancis:
Thanks, have commented out those tests.

The failing tests are for PHP 8.2, which seems to be affecting main as well... and I did re-create the patch, as I made a mess when trying to rebase (one day I'll get that right).

ondrejmirtes:
Thank you :)

PR #355, opened 2022-07-27
Add AssociationMapping type

VincentLanglet:
Hi @ondrejmirtes, I was trying the doctrine/orm 2.13.x branch, after the merge of https://github.com/doctrine/orm/pull/9794
And I'm getting a log of 
```
Access to offset 'isOwningSide' on an unknown class Doctrine\ORM\Mapping\AssociationMapping.
```

Seems like it's because the type is not defined in the stub. When I modified my vendor, it fixed the issue.


VincentLanglet:
I updated the type following the PR https://github.com/doctrine/orm/pull/9965

This seems pretty annoying having to copy paste all changes from orm here.

I think the best would be to "fix" phpstan in order to be able to understand the type defined in the original class, even if it was re-declared in the stub. (Especially here because the type are not used in the stub)
Any idea if it's possible ?

I reproduce the error here: https://github.com/sonata-project/EntityAuditBundle/pull/503

VincentLanglet:
> I think the best would be to "fix" phpstan in order to be able to understand the type defined in the original class, even if it was re-declared in the stub. (Especially here because the type are not used in the stub) Any idea if it's possible ?

I found the issue.

As soon as the stub has a some class phpdoc (I reproduced with `@phpstan-type` but I think it can occur with some other tags), it override all the types declared in the original class. A reproducer can be found here:
https://github.com/phpstan/phpstan-src/pull/1603

ondrejmirtes:
Thank you.

VincentLanglet:
You're welcome. If you have time for a release, it would help me a lot until a solution is found for https://github.com/phpstan/phpstan/issues/7755 :)

ondrejmirtes:
The build is failing and I need to investigate it - if it's a bug or if it's a consequence of your change and I can remove some asserts.

VincentLanglet:
I see. I'll tried to take a look.

For the record my PR just copied-pasted the types from doctrine/orm. So if there is an error in the type, I'll need to do the pr on doctrine/orm too ^^

VincentLanglet:
I just check and it seems you can indeed remove a lot of assert/phpdoc/cast.
I opened https://github.com/phpstan/phpstan-doctrine/pull/359

PR #359, opened 2022-08-08
Fix phpstan build

VincentLanglet:
I checked, and all the errors are related to the new type declared in doctrine/orm.

It allow to remove a lot of useless assert/cast/phpdoc.


ondrejmirtes:
Perfect, thank you.

PR #361, opened 2022-08-14
Add ReadWritePropertiesExtension for Gedmo annotations/attributes

### LINKED ISSUE #231: [RFC] Support by default that Gedmo properties are always written
From PHPStan 1.0 there is a better [Detecting Unused Private Properties, Methods, and Constants](https://phpstan.org/blog/detecting-unused-private-properties-methods-constants) enabled.

Since then, many entity properties get reported as

> Property $foo is never written, only read.

In my case, all of them were properties using `Timestampable` from Doctrine Extensions, without setter/getter.

I suggest to include by default with phpstan-doctrine a PHPStan extension that marks properties using `Timestampable` as always written, following [the official PHPStan instructions](https://phpstan.org/blog/detecting-unused-private-properties-methods-constants#what-if-my-code-is-%E2%80%9Cspecial%E2%80%9D%3F)

What do you think?

---

Here a basic example of that extension that work in my case:

```php
<?php

namespace App\PHPStan;

use PHPStan\Reflection\PropertyReflection;
use PHPStan\Rules\Properties\ReadWritePropertiesExtension;

final class PropertiesExtension implements ReadWritePropertiesExtension
{
    public function isAlwaysRead(PropertyReflection $property, string $propertyName): bool
    {
        return false;
    }

    public function isAlwaysWritten(PropertyReflection $property, string $propertyName): bool
    {
        return str_contains($property->getDocComment(), '@Gedmo\Timestampable');
    }

    public function isInitialized(PropertyReflection $property, string $propertyName): bool
    {
        return false;
    }
}
```

acrobat:
Gedmo annotations (attributes) are used by gedmo listeners to automatically write data to properties and some of the properties are only used by gedmo internally. This ReadWritePropertiesExtension solves the unused property errors for these cases.

```
 ------ ----------------------------------------------------------------------------------
  Line   Example.php
 ------ ----------------------------------------------------------------------------------
  22     Property App\Entity\Example::$locale is unused.
         💡 See: https://phpstan.org/developing-extensions/always-read-written-properties
 ------ ----------------------------------------------------------------------------------
```

Fixes phpstan/phpstan-doctrine#231
Related to phpstan/phpstan#7758

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
 		"doctrine/mongodb-odm": "^1.3 || ^2.1",
 		"doctrine/orm": "^2.11.0",
 		"doctrine/persistence": "^1.3.8 || ^2.2.1",
+		"gedmo/doctrine-extensions": "^3.8",
 		"nesbot/carbon": "^2.49",
 		"nikic/php-parser": "^4.13.2",
 		"php-parallel-lint/php-parallel-lint": "^1.2",
--- a/src/Rules/Gedmo/PropertiesExtension.php
+++ b/src/Rules/Gedmo/PropertiesExtension.php
@@ -1,0 +1,103 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Gedmo;
+
+use Doctrine\Common\Annotations\AnnotationReader;
+use Gedmo\Mapping\Annotation as Gedmo;
+use PHPStan\Reflection\PropertyReflection;
+use PHPStan\Rules\Properties\ReadWritePropertiesExtension;
+use PHPStan\Type\Doctrine\ObjectMetadataResolver;
+use function class_exists;
+use function get_class;
+use function in_array;
+
+class PropertiesExtension implements ReadWritePropertiesExtension
+{
+
+	private const GEDMO_WRITE_CLASSLIST = [
+		Gedmo\Blameable::class,
+		Gedmo\IpTraceable::class,
+		Gedmo\Locale::class,
+		Gedmo\Language::class,
+		Gedmo\Slug::class,
+		Gedmo\SortablePosition::class,
+		Gedmo\Timestampable::class,
+		Gedmo\TreeLeft::class,
+		Gedmo\TreeLevel::class,
+		Gedmo\TreeParent::class,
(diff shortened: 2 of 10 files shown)


acrobat:
The failing php 8.2 checks don't seem to be related to my change

composer.json
```diff
@@ -7,7 +7,8 @@
 	],
 	"require": {
 		"php": "^7.2 || ^8.0",
-		"phpstan/phpstan": "^1.8.0"
+		"phpstan/phpstan": "^1.8.0",
+		"doctrine/annotations": "^1.11.0"
```
ondrejmirtes: Why is this in `require`?

src/Rules/Gedmo/PropertiesExtension.php
```diff
@@ -0,0 +1,88 @@
...
+		Gedmo\Locale::class,
+		Gedmo\Language::class,
+	];
+
+	/** @var AnnotationReader */
+	private $annotationReader;
+
+	public function __construct()
+	{
+		$this->annotationReader = new AnnotationReader();
```
ondrejmirtes: Maybe we could check if this class exists, and if it doesn't, all of the extension methods would return false.

ondrejmirtes (changes requested):

ondrejmirtes:
One more idea - if the class isn't an entity, the extension should return false even if the annotations are present. Please implement this with a test.

Please use `ObjectMetadataResolver::isTransient()` for this.

composer.json
acrobat: I've move this to require because it was used in the constructor but with the other comment you had this can indeed be reverted and the extension will have it as an optional dependency!

src/Rules/Gedmo/PropertiesExtension.php
acrobat: Fixed!

acrobat:
@ondrejmirtes I've implemented the requested changes, thanks for the suggestions!

acrobat:
@ondrejmirtes are there any other changes needed here?

ondrejmirtes (approved):

ondrejmirtes:
I like this, thank you :)

PR #368, opened 2022-09-21
Fix Doctrine PropertiesExtensions

alexndlm:
Property may be writable if it mapped by doctrine


ondrejmirtes:
Please write a test so that we can see what problem does this solve. You can get inspired by tests in https://github.com/phpstan/phpstan-doctrine/tree/1.3.x/tests/Rules/Properties

(You might have to write a new test for UnusedPrivatePropertyRule, I'm not sure.)

ondrejmirtes:
Feel free to open a new PR if you ever decide to finish it, thanks.

PR #379, opened 2022-10-17
Add support for Doctrine generated fields

### LINKED ISSUE #378: Generated fields not correctly treated as written by Doctrine
Having a generated field in an entity:
```
    #[Column(type: Types::DATETIME_MUTABLE, insertable: false, updatable: false,
        columnDefinition: 'datetime GENERATED ALWAYS AS (DATE_ADD(Start, INTERVAL Hours * 60 MINUTE))',
        generated: 'ALWAYS')]
    private \DateTimeInterface $end;
```
Obviously only with a plain getter:
```
    public function getEnd(): \DateTimeInterface
    {
        return $this->end;
    }
```
Makes phpstan not a happy camper:
```
  75     Property WorkEntry::$end is never written, only read.
         💡 See: https://phpstan.org/developing-extensions/always-read-written-properties
```
Solution is to properly parse the `generated` property of the attribute, and consider the property always-written when `ALWAYS` or `INSERT`, ref https://www.doctrine-project.org/projects/doctrine-orm/en/2.13/reference/annotations-reference.html#column

Feature has actually been in Doctrine ORM since 2.11: https://www.doctrine-project.org/2022/01/11/orm-2.11.html

curry684:
Closes #378

### DIFF
--- a/src/Rules/Doctrine/ORM/PropertiesExtension.php
+++ b/src/Rules/Doctrine/ORM/PropertiesExtension.php
@@ -44,6 +44,11 @@ class PropertiesExtension implements ReadWritePropertiesExtension
 			return false;
 		}
 
+		$mapping = $metadata->getFieldMapping($propertyName);
+		if (array_key_exists('generated', $mapping) && $mapping['generated'] !== ClassMetadataInfo::GENERATED_NEVER) {
+			return true;
+		}
+
 		if ($metadata->isReadOnly && !$declaringClass->hasConstructor()) {
 			return true;
 		}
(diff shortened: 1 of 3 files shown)


curry684:
The phpstan error is really interesting as it's comparing this constant:
```
    public const GENERATED_NEVER = 0;
```
To this field:
```
  'generated' =>
  int(2)
```
So the strict comparison is valid. Would appear to be a typing issue in Doctrine itself.

ondrejmirtes:
I fixed the local stub + sent PR to Doctrine: https://github.com/doctrine/orm/pull/10144

ondrejmirtes:
Thank you!

curry684:
Mmm this broke a lot of stuff, you may want to pull that auto release. Apparently the field mapping in Doctrine can be empty in more cases than triggered by the tests.

ondrejmirtes:
> Mmm this broke a lot of stuff

This isn't useful feedback. Please open an issue and describe what problems do you have.

curry684:
See #380, weird that tests didn't pick this up, I got 25 internal errors on a local project right after updating.

PR #380, opened 2022-10-17
Fix missing field mappings in Doctrine

curry684:
Field mappings are apparently not guaranteed to exist.


ondrejmirtes:
A failing test (how does an entity that makes this fail looks like?) too would be nice.

ondrejmirtes:
I cherry-picked your fix, thanks.

curry684:
> A failing test (how does an entity that makes this fail looks like?) too would be nice.

Fixing the borked release had a higher priority  😉    it would seem I overlooked unmapped fields such as join columns in the initial feature.

Issue #377, opened 2022-10-12
QueryBuilder->getQuery() return type conflicts with Psalm


annuh:
Since a recent release of the Psalm-doctrine plugin (https://github.com/psalm/psalm-plugin-doctrine/pull/115) the following code conflicts with Phpstan.

Can the new TKey parameter also be supported by the phpstan-docrine plugin? It looks like Psalm is more correct here.

Code:
```php
/** @var Query<array-key, Country> $query */
$query = $this->createQueryBuilder('country')->getQuery();
```

Without TKey (`/** @var Query<Country> $query */)` produces the following error in Psalm:
```
psalm: InvalidTemplateParam: Extended template param TKey of Doctrine\ORM\Query<App\Entity\Country\Country> expects type array-key, type App\Entity\Country\Country given
psalm: MissingTemplateParam: Doctrine\ORM\Query has missing template params, expecting 2
```

With TKey (`/** @var Query<array-key, Country> $query */`) procesed the following error in PHPStan:
```
phpstan: Generic type Doctrine\ORM\Query<(int|string), App\Entity\Country\Country> in PHPDoc tag @var for variable $query specifies 2 template types, but class Doctrine\ORM\Query supports only 1: TResult
```

ondrejmirtes:
Yeah, feel free to update the stubs :)

ondrejmirtes:
`Query<key, value>` is now supported: https://github.com/phpstan/phpstan-doctrine/commit/4490e56f6fe3f8985a03be073d4545c3fb59cdbb

PR #384, opened 2022-10-25
Result infering: support indexBy and lists

### LINKED ISSUE #381: Infering getResult as list
Hi, I'm currently testing upcoming version of phpstan (1.9.x-dev) where you cannot pass `array<Foo>` argument to `list<Foo>` parameter. Currently, simple queries like this
```php
$list = $qb->select('e')->from(Entity::class, 'e')->getResult();
```
are infered as `array<mixed, Entity>` ([source](https://github.com/phpstan/phpstan-doctrine/blob/1.3.x/src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php#L129)) even though it is actually a list<Entity>. This is getting problematic in phpstan 1.9, would it be possible to return it as list when no `indexBy` is used? Or is there any blocker? 

I assume, this is mainly question for @arnaud-lb
Thank you

janedbal:
Closes https://github.com/phpstan/phpstan-doctrine/issues/381

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
@@ -11,6 +11,7 @@ use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Accessory\AccessoryArrayListType;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\Constant\ConstantIntegerType;
+use PHPStan\Type\Doctrine\Type\ListIndexMarkerType;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
 use PHPStan\Type\Generic\GenericObjectType;
 use PHPStan\Type\IntegerType;
@@ -142,7 +143,7 @@ final class QueryResultDynamicReturnTypeExtension implements DynamicMethodReturn
 					$queryResultType
 				);
 			default:
-				return $queryIndexType instanceof MixedType
+				return $queryIndexType instanceof ListIndexMarkerType
 					? AccessoryArrayListType::intersectWith(
 						TypeCombinator::intersect(
 							new ArrayType(new IntegerType(), $queryResultType),
@@ -150,7 +151,7 @@ final class QueryResultDynamicReturnTypeExtension implements DynamicMethodReturn
 						)
 					)
 					: new ArrayType(
-						$queryIndexType,
+						new MixedType(),
 						$queryResultType
 					);
 		}
--- a/src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
@@ -8,6 +8,7 @@ use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\ShouldNotHappenException;
+use PHPStan\Type\Accessory\AccessoryArrayListType;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
@@ -18,6 +19,7 @@ use PHPStan\Type\MixedType;
 use PHPStan\Type\NullType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
+use PHPStan\Type\TypeUtils;
 use PHPStan\Type\VoidType;
 
 final class QueryResultDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtension
@@ -70,12 +72,25 @@ final class QueryResultDynamicReturnTypeExtension implements DynamicMethodReturn
 
 		$queryType = $scope->getType($methodCall->var);
 		$queryResultType = $this->getQueryResultType($queryType);
+		$queryIndexType = $this->getQueryIndexType($queryType);
 
 		return $this->getMethodReturnTypeForHydrationMode(
 			$methodReflection,
 			$hydrationMode,
-			$queryResultType
+			$queryResultType,
+			$queryIndexType
 		);
+	}
+
+	private function getQueryIndexType(Type $queryType): Type
+	{
+		if (!$queryType instanceof GenericObjectType) {
+			return new MixedType();
+		}
+
+		$types = $queryType->getTypes();
+
+		return $types[1] ?? new MixedType();
 	}
 
 	private function getQueryResultType(Type $queryType): Type
@@ -92,7 +107,8 @@ final class QueryResultDynamicReturnTypeExtension implements DynamicMethodReturn
 	private function getMethodReturnTypeForHydrationMode(
 		MethodReflection $methodReflection,
 		Type $hydrationMode,
-		Type $queryResultType
+		Type $queryResultType,
+		?Type $queryIndexType
 	): Type
 	{
 		$isVoidType = (new VoidType())->isSuperTypeOf($queryResultType);
@@ -126,10 +142,17 @@ final class QueryResultDynamicReturnTypeExtension implements DynamicMethodReturn
 					$queryResultType
 				);
 			default:
-				return new ArrayType(
-					new MixedType(),
-					$queryResultType
-				);
+				return $queryIndexType instanceof MixedType
+					? AccessoryArrayListType::intersectWith(
+						TypeCombinator::intersect(
+							new ArrayType(new IntegerType(), $queryResultType),
+							...TypeUtils::getAccessoryTypes($queryResultType)
+						)
+					)
+					: new ArrayType(
+						$queryIndexType,
+						$queryResultType
+					);
 		}
 	}
 
--- a/src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
@@ -91,7 +91,7 @@ final class QueryResultDynamicReturnTypeExtension implements DynamicMethodReturn
 
 		$types = $queryType->getTypes();
 
-		return $types[1] ?? new MixedType();
+		return $types[0] ?? new MixedType();
 	}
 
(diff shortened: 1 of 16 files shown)


src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
```diff
@@ -126,10 +142,17 @@ private function getMethodReturnTypeForHydrationMode(
...
-				);
+				return $queryIndexType instanceof MixedType
+					? AccessoryArrayListType::intersectWith(
+						TypeCombinator::intersect(
+							new ArrayType(new IntegerType(), $queryResultType),
+							...TypeUtils::getAccessoryTypes($queryResultType)
+						)
+					)
+					: new ArrayType(
+						$queryIndexType,
```
janedbal: This can really be only string or int. We should handle nullable values, objects, booleans etc. 

**Or** we could simplify this MR not to try handle those cases and just distinguish list / array whether indexBy was used or not.

tests/Type/Doctrine/data/QueryResult/config.neon
```diff
@@ -3,3 +3,5 @@ includes:
 parameters:
 	doctrine:
 		objectManagerLoader: entity-manager.php
+	featureToggles:
+		listType: true
```
janedbal, line 7: Not sure I can keep this here if we would merge this.

src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
```diff
@@ -70,14 +72,27 @@ public function getTypeFromMethodCall(
...
 			$hydrationMode,
-			$queryResultType
+			$queryResultType,
+			$queryIndexType
 		);
 	}
 
+	private function getQueryIndexType(Type $queryType): Type
+	{
+		if (!$queryType instanceof GenericObjectType) {
```
janedbal, line 88: I think this should rather be `PHPStan\Type\Doctrine\Query\QueryType`, but this was used below.

src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
janedbal: This is the main question. Do we want to infer indexBy types? Personally, I dont need that.

janedbal: I'd start with just distinguishing list / array while having the new generic type of Query as seen in this MR to be able to improve it later.

src/Type/Doctrine/Query/QueryType.php
```diff
@@ -14,10 +14,15 @@ class QueryType extends GenericObjectType
...
-		parent::__construct('Doctrine\ORM\Query', [$resultType]);
+	public function __construct(
+		string $dql,
+		?Type $resultType,
+		?Type $indexType
+	) {
+		parent::__construct('Doctrine\ORM\Query', [
+			$resultType ?? new MixedType(),
+			$indexType ?? new MixedType(),
+		]);
```
arnaud-lb: I think that `$indexType` can legitimately be `mixed`, depending on the type of the field. As an effect, `getResult()` would be typed as `list<...>`, which is not right.

Maybe we could introduce a subtype of `Query<TResult>`, like `IndexedQuery<TIndex,TResult>` ?

This would also avoid the BC break of adding a new type parameter to `Query`.

src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
arnaud-lb: > I'd start with just distinguishing list / array while having the new generic type of Query as seen in this MR to be able to improve it later.

:+1:

src/Type/Doctrine/Query/QueryType.php
ondrejmirtes: Right now we're incompatible with Psalm https://github.com/phpstan/phpstan-doctrine/issues/377 so I wouldn't avoid the change now.

janedbal: What about using some artificial custom type for this purpose? Like `ListIndexMarkerType extends MixedType`?

janedbal: I think I fixed this issue, hopefully. 

Default value of `QueryType` is still `MixedType`. But only when we get valid answer from `QueryResultTypeBuilder::getIndexType`, we get either `ListIndexMarkerType` or the type of indexBy type (can be mixed). But the extension now infers lists only for `ListIndexMarkerType` so we should be getting `list<T>` only in those cases where it really passed the `QueryResultTypeWalker` without any issues.

arnaud-lb: If we want to be compatible with Psalm, we need to put the TIndex param before TResult as well.

The artificially custom type is a good idea. Will the user be able to type a list Query in type annotations like in the following snippet ?

``` php
/** @return Query<ListQuery,Foo> */
function getQuery(): Query;
```

janedbal: > If we want to be compatible with Psalm, we need to put the TIndex param before TResult as well.

But that will be BC break. Do we really want to do it?

> Will the user be able to type a list Query

I dont think so, I'm not sure how phpstan converts phpdoc strings to certain types, but I added nothing to support our artificial type, so it cannot work. I'd need to dig into that. You should be able to use any other type like int or string, but that does not yet infer key types as we agreed above.

ondrejmirtes: We can change the types only for bleedingEdge. We're doing this whole PR for the list type which is only in bleedingEdge anyway.

janedbal: That seems hard. Wouldnt it be actually better to sync with psalm even in non-bleedingEdge? Meaning perform this (imo minor) BC-break. We would need to do it some time in future anyway.

We use phpstan-doctrine infering quite lot in our codebase, but we never needed to use the generics ourselves. Do you @arnaud-lb use it in userland?

janedbal: Note: I just used this branch as-is in our codebase without any issues and it allowed me to actually enable listType feature.

arnaud-lb: > Maybe we could introduce a subtype of `Query<TResult>`, like `IndexedQuery<TIndex,TResult>` ?

I'm withdrawing this suggestion, it wouldn't work.

After all we should probably add `TKey`, as we will have to do this change if we add full INDEX BY support anyway, and this makes us compatible with Psalm at the same time.

Users will have to change `Query<Something>` to `Query<int,Something>` for non-indexed queries or `Query<mixed,Something>` for indexed queries (and to something more specific than `mixed` if we add full INDEX BY support).

> We use phpstan-doctrine infering quite lot in our codebase, but we never needed to use the generics ourselves. Do you @arnaud-lb use it in userland?

I used the `Query<...>` or `AbstractQuery<...>` types on only a few occasions, but it would have been very inconvenient if the result of some `createQuery()` call couldn't be accepted by any `Query<...>` type. It would be good enough if `Query<mixed,Something>` or `Query<int,Something>` accepts a QueryType that have TKey=ListIndexMarkerType (I don't know if it does currently, we may need to declare TKey as @template-covariant, which would be correct).

PHPDoc types are resolved in [TypeNodeResolver](https://github.com/phpstan/phpstan-src/blob/d0207ed79e94e4b4daceb0f83abb1f481ee3250b/src/PhpDoc/TypeNodeResolver.php), and plugins can implement TypeNodeResolverExtensions to add custom types: https://phpstan.org/developing-extensions/custom-phpdoc-types

Alternatively, we could bind TKey to the `void` type to denote a non-indexed Query. We bind TResult to `void` to denote non-SELECT queries already.

janedbal: > Users will have to change Query<Something> to Query<int,Something> for non-indexed queries or Query<mixed,Something> for indexed queries (and to something more specific than mixed if we add full INDEX BY support).

I'm ok with this BC. I can swap the types. Will do that.

> It would be good enough if Query<mixed,Something> or Query<int,Something> accepts a QueryType that have TKey=ListIndexMarkerType

As `ListIndexMarkerType exteds MixedType`, any key type accepts it, even `Query<string, Something>`. Maybe we could make `ListIndexMarkerType extends IntegerType`, that would be more precise and should handle your cases as described. Will do that.

> Alternatively, we could bind TKey to the void type to denote a non-indexed Query. We bind TResult to void to denote non-SELECT queries already.

Personally, I see the custom type as better option than abusing existing type with some meaning.

----------

Anything else that needs to be done before merging this?

janedbal: > PHPDoc types are resolved in [TypeNodeResolver](https://github.com/phpstan/phpstan-src/blob/d0207ed79e94e4b4daceb0f83abb1f481ee3250b/src/PhpDoc/TypeNodeResolver.php), and plugins can implement TypeNodeResolverExtensions to add custom types: https://phpstan.org/developing-extensions/custom-phpdoc-types

I can add support of some `list-marker` phpdoc for the `ListIndexMarkerType`, but I think it might be confusing for people seeing it error messages. But the truth is that this "hidden" type is not ideal as well (it is outputted as any other `mixed` now).

janedbal: - [x] Swapped TKey and TResult so that we are psalm-compatible
- [x] ListIndexMarkerType extends IntegerType

janedbal: I'd love to finalize this. I'm just waiting for your answers / review. Thanks.

VincentLanglet:
I think you can rebase/retrigger the ci to fix the 8.2 builds

janedbal:
> I think you can rebase/retrigger the ci to fix the 8.2 builds

Rebased

ondrejmirtes:
I took this PR and reworked it with some changes according to my taste: https://github.com/phpstan/phpstan-doctrine/commit/4490e56f6fe3f8985a03be073d4545c3fb59cdbb

1) List-indexed query is marked with `null` TKey instead of a custom marker type
2) `Query<X>` is still allowed thanks to a custom TypeNodeResolver extension (https://phpstan.org/developing-extensions/custom-phpdoc-types)

Thank you :)

PR #388, opened 2022-11-03
Fix Repository stubs

enumag:
Using `something[]` in docblocks is a bad practice (in fact I'd very much love to have a rule that would disallow it) because then you don't know if the keys are ints or strings. And in many cases when the syntax is used, it is actually a `list<>`.

I changed the types according to those found in Doctrine repositories. I find it a bit odd for ObjectRepository to not return `list<>` but this is how it is currently typed in Doctrine. EntityRepository is typed to return `list<>` as expected.

### DIFF
--- a/stubs/DocumentRepository.stub
+++ b/stubs/DocumentRepository.stub
@@ -20,21 +20,21 @@ class DocumentRepository implements ObjectRepository
 	public function find($id, $lockMode = null, $lockVersion = null);
 
 	/**
-	 * @phpstan-return TDocumentClass[]
+	 * @phpstan-return array<int, TDocumentClass>
 	 */
 	public function findAll();
 
 	/**
-	 * @phpstan-param mixed[] $criteria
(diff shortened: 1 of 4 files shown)


enumag:
CI failure seems to be unrelated.

ondrejmirtes (changes requested):
Are your distinctions here between `array<int, TEntityClass>` and `list<TEntityClass>` precise? For example why `findAll()` doesn't return a list?

Thanks.

enumag:
I explained in the PR description... I used the types from Doctrine repositories:
- ObjectReposirory is typed to return `array<int, object>` instead of list, not sure why
- DocumentRepository extends ObjectRepository without changing the type so I used the same
- EntityRepository extends ObjectRepository but narrows the type down to `list<object>` so I did here as well

If you want I can change it list everywhere but I can't tell whether such change is correct.

ondrejmirtes:
Alright, thank you.

PR #390, opened 2022-11-10
Fix type for enumType in ClassMetaData

### LINKED ISSUE #391: Incorrect enumType in ClassMetadataInfo
Due to a missing `\` the \BackedEnum type for enumType in ClassMetadataInfo is incorrect. I have created a PR to fix this, but I don't know how to make the checks happy: https://github.com/phpstan/phpstan-doctrine/pull/390. What should I do to fix that?

tjveldhuizen:
Since the ClassMetadataInfo-class is namespaced, the BackedEnum-reference should be prefixed with a back slash. Without it, the type of `enumType` becomes `Doctrine\ORM\Mapping\BackedEnum` which is invalid.

Fixes #391

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultTypeWalker.php
+++ b/src/Type/Doctrine/Query/QueryResultTypeWalker.php
@@ -45,7 +45,6 @@ use function intval;
 use function is_numeric;
 use function is_object;
 use function is_string;
-use function is_subclass_of;
 use function serialize;
 use function sprintf;
 use function strtolower;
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
About:

```

 ------ --------------------------------------------------------------------- 
  Line   src/Type/Doctrine/Query/QueryResultTypeWalker.php                    
 ------ --------------------------------------------------------------------- 
  1303   Call to function is_subclass_of() with class-string<BackedEnum> and  
         'BackedEnum' will always evaluate to true.                           
 ------ --------------------------------------------------------------------- 
```

This means you can simplify this condition, from:

```php
if (!is_string($enumType) || !class_exists($enumType) || !is_subclass_of($enumType, BackedEnum::class)) {
```

to:

```php
if (!is_string($enumType) || !class_exists($enumType)) {
```

tjveldhuizen:
OK, I was assuming that this code was copied from the Doctrine repository to test if the PHPStan-code was correct. I'll update the PR

ondrejmirtes:
Thank you!

PR #392, opened 2022-11-16
Reduce big int type from string to numeric-string

VincentLanglet:
When querying a bigInt field, the property is currently considered as a `string`.
I think we can narrow this to `numeric-string` since the internal type is an `int`.


VincentLanglet:
Tests are still failing @ondrejmirtes, but I don't think it's related

VincentLanglet:
8.2 are passing now.
Only the lowest builds are failing, but this doesn't seem related.

ondrejmirtes:
Thank you.

PR #395, opened 2022-11-25
#202 Add rule to prevent final constructors in doctrine entities

### LINKED ISSUE #202: Implement a rule that prevents final constructors in Doctrine Entities
Hello there,

I'm having quite a hard time getting my head around this : [final methods in Doctrine Entities](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/architecture.html#entities). There's [a PR](https://github.com/phpstan/phpstan-doctrine/pull/143) that added a check that Entities cannot be final (the class itself), but [a comment](https://github.com/phpstan/phpstan-doctrine/pull/143#issuecomment-698225313) from the author said he doesn't want to add checks for final methods.

However, from what I can see in my Doctrine proxied classes in var/cache in production, *all* methods of the class is inherited/overridden in proxies. So it just cannot work with a final method, whichever it is.

Why hasn't this check been added then ?

I'm curious about your insights. Thanks!
gnutix

LeoVie:
This fixes #202.

- Prevent Doctrine entities from containing final constructors
- At the same time, don't forbid non-constructor final methods in Doctrine entities

What do you think about that?

### DIFF
--- a/rules.neon
+++ b/rules.neon
@@ -23,7 +23,6 @@ rules:
 	- PHPStan\Rules\Doctrine\ORM\DqlRule
 	- PHPStan\Rules\Doctrine\ORM\RepositoryMethodCallRule
 	- PHPStan\Rules\Doctrine\ORM\EntityNotFinalRule
-	- PHPStan\Rules\Doctrine\ORM\EntityConstructorNotFinalRule
 
 conditionalTags:
 	PHPStan\Rules\Doctrine\ORM\EntityMappingExceptionRule:
--- a/rules.neon
+++ b/rules.neon
@@ -23,9 +23,12 @@ rules:
 	- PHPStan\Rules\Doctrine\ORM\DqlRule
 	- PHPStan\Rules\Doctrine\ORM\RepositoryMethodCallRule
 	- PHPStan\Rules\Doctrine\ORM\EntityNotFinalRule
+	- PHPStan\Rules\Doctrine\ORM\EntityConstructorNotFinalRule
 
 conditionalTags:
 	PHPStan\Rules\Doctrine\ORM\EntityMappingExceptionRule:
+		phpstan.rules.rule: %featureToggles.bleedingEdge%
+	PHPStan\Rules\Doctrine\ORM\EntityConstructorNotFinalRule:
 		phpstan.rules.rule: %featureToggles.bleedingEdge%
 
 services:
@@ -54,3 +57,5 @@ services:
 			bleedingEdge: %featureToggles.bleedingEdge%
 		tags:
 			- phpstan.rules.rule
+	-
+		class: PHPStan\Rules\Doctrine\ORM\EntityConstructorNotFinalRule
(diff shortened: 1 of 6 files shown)


LeoVie:
The failing checks don't seem to be related to the changes of this PR.

src/Rules/Doctrine/ORM/EntityConstructorNotFinalRule.php
```diff
@@ -0,0 +1,45 @@
...
+			return [];
+		}
+
+		$classReflection = $scope->getClassReflection();
+
+		if ($classReflection === null) {
+			throw new ShouldNotHappenException();
+		}
+
+		return [sprintf(
```
ondrejmirtes: Please use RuleErrorBuilder here.

src/Rules/Doctrine/ORM/EntityConstructorNotFinalRule.php
```diff
@@ -0,0 +1,45 @@
...
+	{
+		if ($node->name->name !== '__construct') {
+			return [];
+		}
+
+		if (!$node->isFinal()) {
+			return [];
+		}
+
+		$classReflection = $scope->getClassReflection();
```
ondrejmirtes, line 43: You're reporting in all classes, not just entities.

ondrejmirtes (changes requested):
Also, the rule needs to be registered in rules.neon, but only for bleedingEdge.

Thanks.

src/Rules/Doctrine/ORM/EntityConstructorNotFinalRule.php
LeoVie: If I'm correct, this should work now only for entities. I've taken this code from `EntityNotFinalRule` for the detection, if the class is an entity

```php
if ($this->objectMetadataResolver->isTransient($classReflection->getName())) {
	return [];
}

$metadata = $this->objectMetadataResolver->getClassMetadata($classReflection->getName());
if ($metadata !== null && $metadata->isEmbeddedClass === true) {
	return [];
}
```

LeoVie:
Hi, sorry for the delay :)
I've updated the PR according to your suggestions. I hope, everything is correct now.

rules.neon
```diff
@@ -23,10 +23,13 @@ rules:
 	- PHPStan\Rules\Doctrine\ORM\DqlRule
 	- PHPStan\Rules\Doctrine\ORM\RepositoryMethodCallRule
 	- PHPStan\Rules\Doctrine\ORM\EntityNotFinalRule
+	- PHPStan\Rules\Doctrine\ORM\EntityConstructorNotFinalRule
```
ondrejmirtes: You need to remove it from the `rules` section.

ondrejmirtes (changes requested):

rules.neon
LeoVie: Ah, okay. I removed it there. :)

ondrejmirtes:
Thank you!

PR #397, opened 2022-12-09
Update AbstractQuery.stub

enumag:


enumag:
The test failures are unrelated to my patch.

enumag:
@ondrejmirtes Can you merge this please? Because of the current impossibility to override a stub on my side it's currently kinda blocking me.

ondrejmirtes:
Thank you.

enumag:
Thanks!

stof:
This breaks things when using `INDEX BY` in DQL, as the result is not a list in such case. Returning a list is something that needs to be handled by improving the type extension analysing the query, not by this broken stub

enumag:
I see. Didn't know that. Alright let's revert this and handle it better.

VincentLanglet:
There is the wip https://github.com/phpstan/phpstan-doctrine/pull/384

enumag:
@VincentLanglet That PR only handles object result while my case is for array result but it does the hard work of detecting the list result. I should be able to fix this correctly after that PR is merged.

PR #401, opened 2022-12-18
Add impure annotation in some Collection methods

### LINKED ISSUE #203: Phpstan is messing with isEmpty and first/last value.
In 
https://github.com/phpstan/phpstan-doctrine/pull/172
https://github.com/phpstan/phpstan-doctrine/pull/173
`TypeSpecifyingExtension` were introduced to narrow the return type of the `Collection::first` and `Collection::last` methods.

With the following tests
```
$entityOrFalse = $collection->first();
$entityOrFalse;

if ($collection->isEmpty()) {
	$false = $collection->first();
	$false;
}

if (!$collection->isEmpty()) {
	$entity = $collection->first();
	$entity;
}
```

I discovered that the following code it introduce some issues if you modify the `$collection` AFTER the `isEmpty` check and BEFORE the `first` or `last` call. Of course, I couldn't call this a good practice, so I don't know if it's a big deal.

For instance
```
if (!$collection->isEmpty()) {
        $collection->remove($foo);
	$entity = $collection->first();
	$entity; // This now might be false if the collection had only one element, `$foo`. But phpstan consider it to be an entity.
}
```
Or 
```
if ($collection->isEmpty()) {
        $collection->add($bar);
	$false = $collection->first();
	$false; // This now should return `$bar` but phpstan consider it to be `false`.
}
```

Signature of Collection method can be found here: https://github.com/doctrine/collections/blob/1.6.x/lib/Doctrine/Common/Collections/Collection.php
- `add` return TRUE.
- `remove` return the removed element or NULL
- `removeElement` return TRUE
- `clear` return void
- `set` return void
- Since the interface extends `ArrayAccess`, I'm not sure about the behavior of `offsetSet`.

If I understand correctly the comment https://github.com/phpstan/phpstan-doctrine/pull/172#issuecomment-890553122
it require to add `@phpstan-impure` to every of those methods witch return a value (`add`, `remove`, `removeElement` and `offsetSet` maybe (?)). And to add tests too.

VincentLanglet:
Closes https://github.com/phpstan/phpstan-doctrine/issues/203

### DIFF
--- a/tests/Type/Doctrine/Collection/data/collection.php
+++ b/tests/Type/Doctrine/Collection/data/collection.php
@@ -2,6 +2,8 @@
 
 use Doctrine\Common\Collections\ArrayCollection;
 
+use function PHPStan\Testing\assertType;
+
 class MyEntity
 {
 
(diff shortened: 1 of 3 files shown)


VincentLanglet:
Lowest failure are unrelated since they are on the 1.3.x branch.

tests/Type/Doctrine/Collection/data/collection.php
```diff
@@ -26,6 +26,13 @@ class MyEntity
...
 	$false2 = $collection->last();
 	$false2;
+
+	$collection->add($new);
+	$result1 = $collection->first();
+	$result1;
+
+	$result2 = $collection->last();
+	$result2;
 }
```
ondrejmirtes, line 48: Please do not modify the test like that, the diff of the rule test case is unreadable. Also, it'd make sense to refactor this into a TypeInferenceTestCase :) Thank you.

ondrejmirtes (changes requested):

tests/Type/Doctrine/Collection/data/collection.php
VincentLanglet: The test is now refactored into a TypeInferenceTestCase ;)

ondrejmirtes:
Thank you.

PR #402, opened 2022-12-23
Update ClassMetadataInfo.stub

VincentLanglet:
Copy pasted from https://github.com/doctrine/orm/blob/2.14.x/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php

It avoid the error
```
Access to offset 'name' on an unknown class Doctrine\ORM\Mapping\DiscriminatorColumnMapping.       
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols
```
until https://github.com/phpstan/phpstan/issues/7755 is fixed.


ondrejmirtes:
Thank you!

PR #404, opened 2022-12-26

### LINKED ISSUE #403: Add support for other hydration mode in QueryResultDynamicReturnTypeExtension
I think we can "easily" improve the QueryResultDynamicReturnTypeExtension to supports more than `HYDRATE_OBJECT`.

<details>
<summary>Click here to see the different hydration modes</summary>

- HYDRATE_OBJECT: Default behavior, return the result without modifications.

- HYDRATE_ARRAY: Every object are transformed to an array, so
```
array<array{id: int, language: string}>
```
is untouched, but
```
array<array{id: int, user: User}>
```
would be transformed to
```
array<array{id: int, user: array{username: ..., email:..., ....}}>
```

- HYDRATE_SCALAR: Kinda like the HYDRATE_ARRAY, but everything is merged inside a single array, so
```
array<array{id: int, language: string}>
```
is untouched, but
```
array<array{id: int, user: User}>
```
would be transformed to
```
array<array{id: int, u_username: ..., u_email:..., ....}>
```

- HYDRATE_SINGLE_SCALAR: Kinda like the HYDRATE_SCALAR, but only one key inside the array will be supported
```
array<array{id: int}>
```
is untouched, but
```
array<array{user: User}>
```
would be transformed to
```
array<array{u_username: ...}> // Only if there is only one property inside the object
```

- HYDRATE_SCALAR_COLUMN: Kinda like the HYDRATE_SINGLE_SCALAR, but the single-array element is replaced by the element.
```
array<array{id: int}>
```
is changed to
```
array<int>
```

- HYDRATE_SIMPLEOBJECT: I don't really know this one.
</details>

I would say that good steps would be
- Infer the result when there is no `object` inside the result (for most of the hydration mode it's like HYDRATE_OBJECT)
- Infer the result when there is an `object` inside the result by replacing to `array<string, mixed>`

VincentLanglet:
Closes #403

### DIFF
--- a/tests/Type/Doctrine/data/QueryResult/queryResult.php
+++ b/tests/Type/Doctrine/data/QueryResult/queryResult.php
@@ -143,11 +143,11 @@ class QueryResultTest
 	}
 
 	/**
-	 * Test that we properly infer the return type of Query methods with explicit hydration mode that is not HYDRATE_OBJECT
+	 * Test that we properly infer the return type of Query methods with explicit hydration mode of HYDRATE_ARRAY
 	 *
-	 * We are never able to infer the return type here
+	 * We can infer the return type by changing every object by an array
 	 */
-	public function testReturnTypeOfQueryMethodsWithExplicitNonObjectHydrationMode(EntityManagerInterface $em): void
+	public function testReturnTypeOfQueryMethodsWithExplicitArrayHydrationMode(EntityManagerInterface $em): void
 	{
 		$query = $em->createQuery('
 			SELECT		m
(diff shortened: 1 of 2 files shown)


VincentLanglet:
Hi @arnaud-lb, I'm trying to improve https://github.com/phpstan/phpstan-doctrine/pull/232 to supports all hydration modes.

Do you mind taking a look ?
I dunno if we can support everything but maybe some of my ideas are going in the right direction :)

VincentLanglet:
Hi @greg0ire, maybe you can help me on this.

I'm not familiar with all the hydratation mode.
Do you agree with the expected types I wrote in the tests ?

greg0ire:
Not super familiar either, but nothing strikes me as incorrect in what you wrote.

VincentLanglet:
> Not super familiar either, but nothing strikes me as incorrect in what you wrote.

Thanks, if you have a suggestion about who I should/could ping ; I'm open to it.

@ondrejmirtes I would say, this is ready to be reviewed.

tests/Type/Doctrine/data/QueryResult/queryResult.php
```diff
@@ -143,45 +143,391 @@ public function testReturnTypeOfQueryMethodsWithExplicitObjectHydrationMode(Enti
...
+	public function testReturnTypeOfQueryMethodsWithExplicitArrayHydrationMode(EntityManagerInterface $em): void
 	{
 		$query = $em->createQuery('
 			SELECT		m
 			FROM		QueryResult\Entities\Many m
 		');
 
 		assertType(
-			'mixed',
+			'list<array>',
```
ondrejmirtes, line 158: We should be more specific here - based on the entity, we should know the array shape.

ondrejmirtes (changes requested):

tests/Type/Doctrine/data/QueryResult/queryResult.php
VincentLanglet: I agree. But I thought this could have been a good first step.
Wouldn't be ok to merge this and improve the array shape in a second PR ?

`list<array>` is better than mixed,
then `list<array{id: int, ...}>` will be better than `list<array>`.

Also I had no idea how to improve the 
```
if ($isObject->yes()) {
    return new ArrayType(new MixedType(), new MixedType());
}
```
part with the PHPStan kit.

ondrejmirtes (approved):

ondrejmirtes:
Yes, that's fine. I'm gonna merge this and test on real-world projects before releasing it. Thanks.

ondrejmirtes:
I think it's mostly fine, but it can lead to new errors. Previously where there was `mixed` now there's a more specific type. Which means that if you're on level 8, you can get a new error like:

```
should return array<array{x: string, y: string|null, z: string|null, aa: int|null}> but returns list<array>.
```

This is happening for `HYDRATE_SCALAR` when selecting a few specific columns in `SELECT`.

ondrejmirtes:
But on a different project it leads to 108 new errors.

So there are situations where it's wrong.

Some examples:

```
			->select('IDENTITY(n.related)')
			->from(DistrictNeighbour::class, 'n')
			->andWhere('n.center = ?', $districtId)
			->getQuery();

		return array_map(static fn (array $row): UuidInterface => $row['related'], $query->getArrayResult());
```

leading to: `Offset 'related' does not exist on array{1: string}.`

Or:

```
			->select('d.name, d.id')
			->from(Destination::class, 'd')
			->addOrderBy('d.name')
			->getQuery()
			->getArrayResult();

		$destinations = [];
		foreach ($result as $item) {
			$destinations[(string) $item['id']->toString()] = (string) $item['name'];
		}
```

leading to: `Cannot call method toString() on array.`

Or:

```
$eventsCount = $this->entityManager->createQueryBuilder()
			->select('COUNT(cle.id)')
			->from(CountryLocalityEvent::class, 'cle')
			->andWhere('cle.user = ?', $user)
			->andWhere('cle.expirationTime > ?', DateTimeProvider::now())
			->getQuery()
			->getSingleScalarResult();

		return $eventsCount > 0;
```

leading to: `Comparison operation ">" between list<array{1: int<0, max>|numeric-string}> and 0 results in an error.`

Because of these problems, I'm gonna revert it now. Please send a new PR with improvements. Thank you.

VincentLanglet:
> But on a different project it leads to 108 new errors.
> 
> So there are situations where it's wrong.

Thanks for the report.
It should be better with https://github.com/phpstan/phpstan-doctrine/pull/412

PR #405, opened 2022-12-26
Fix build

VincentLanglet:
@ondrejmirtes I found a way to improve https://github.com/phpstan/phpstan-doctrine/commit/04482af57151003ba49ac246bae2d6c22cd15d36 in order to have every build (highest/lowest) green :)

### DIFF
--- a/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
+++ b/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
@@ -6,6 +6,7 @@ use Composer\InstalledVersions;
 use DateTime;
 use DateTimeImmutable;
 use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\Column;
 use Doctrine\ORM\Query\AST\TypedExpression;
@@ -48,6 +49,7 @@ use function class_exists;
 use function count;
 use function property_exists;
 use function sprintf;
+use function strpos;
 use function version_compare;
 use const PHP_VERSION_ID;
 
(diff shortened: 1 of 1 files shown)


tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
```diff
@@ -1310,6 +1312,8 @@ public function getTestData(): iterable
 			',
 		];
 
+		// There is no error after the merge of https://github.com/doctrine/dbal/pull/5755
+		$moduloError = strpos((new SqlitePlatform())->getModExpression('', ''), '%') === false;
```
VincentLanglet, line 1316: Dunno the full reason but the modExpression implementation was changed from 
```
Mod(A, B)
```
to
```
A % B
```
and this is not throwing an error anymore.

ondrejmirtes:
Thank you!

ondrejmirtes:
Because you were able to fix this problem, could you please look at this one too? You can reproduce it locally if you use development php.ini that reports deprecations. https://github.com/phpstan/phpstan-doctrine/actions/runs/3861924994/jobs/6583179674

Thank you very much.

VincentLanglet:
Sure

ondrejmirtes:
Don't worry about it too much if it's hard, we can remove the test or change it to a function that SQLite knows natively. I don't think the intention was to test user-defined function via DBAL's SqlitePlatform.

VincentLanglet:
> Don't worry about it too much if it's hard, we can remove the test or change it to a function that SQLite knows natively. I don't think the intention was to test user-defined function via DBAL's SqlitePlatform.

I have three suggestions:
- Removing this case: This is the only one which is testing that we get an exception when we're using sql incorrectly. Not sure it's the purpose of the test. And we have different behavior (exception, deprecation, etc...) because of the different doctrine version.

- Removing half of the case: Skipping it when no exception is expected
```
        $moduloError = strpos((new SqlitePlatform())->getModExpression('', ''), '%') === false;
		if ($moduloError) {
			yield 'mod function error' => [
				$this->constantArray([
					[new ConstantIntegerType(1), TypeCombinator::addNull($this->uintStringified())],
				]),
				'
				SELECT		MOD(10, NULLIF(m.intColumn, m.intColumn))
			',
				'Modulo by zero',
			];
		}
```

- Adding a parameter to test the deprecation, but since `expectDeprecation` is not working here we will need something hacky with the error_handler to catch the deprecation.

ondrejmirtes:
Yeah, please delete it.

Issue #367, opened 2022-09-14
stubs/ORM/Query/Expr.stub wrongs params


Thierry-Kq:
Hi, i got this error : 

```
Parameter #2 $x of method Doctrine\ORM\Query\Expr::between() expects string, int|string given
```

But if i look in lib/Doctrine/ORM/Query/Expr.php, i got theses params : 

```
     * @param mixed      $val Valued to be inspected by range values.
     * @param int|string $x   Starting range value to be used in BETWEEN() function.
     * @param int|string $y   End point value to be used in BETWEEN() function.
```

I solved  it on my side by casting my int into string, but i guess the stubs/ORM/Query/Expr.stub is wrong, or maybe i misunderstood something.

```
	 * @param string $val
	 * @param string $x
	 * @param string $y
```

Im using phpstan-doctrine 1.3.13, phpstan/phpstan 1.8.5, phpstan/phpstan-symfony 1.2.13 and doctrine/orm 2.13.1

ondrejmirtes:
1) Research when the int type was added/removed in the Git history of the library
2) If the int type doesn't look deprecated and is going to be supported into the future, send a PR here with a stub update.

VincentLanglet:
The stub was introduced by @craigfrancis in https://github.com/phpstan/phpstan-doctrine/pull/324

And this doesn't seems to be deprecated
https://github.com/doctrine/orm/blob/0aa45dd607100182294d31dee42b89338773f87e/lib/Doctrine/ORM/Query/Expr.php#L638-L650

But there is a conditional return type:
```
@return ($val is literal-string ? ($x is literal-string ? ($y is literal-string ? literal-string&non-empty-string : string) : string) : string)
```
which will maybe need to be updated since `$x` and `$y` can be `int`.

If I look at https://phpstan.org/r/9324b003-b39b-49c9-b911-13a6d4521638
- `(string) $int` is not considered as literal
- literal-int doesn't exists.

So, I'm not sure there is a solution to infer
```
->between('literal', 1, 9)
```
as a literal-string since
```
->between('literal', '1', '9')
```
does.

craigfrancis:
Thanks for finding Vincent,

I've created [PR #407](https://github.com/phpstan/phpstan-doctrine/pull/407), which simply allows int values for these arguments.

There is no need to change the conditional return type, as it shouldn't return a `literal-string` when an integer is used.

The failing "mod function error" test should be unrelated to my changes.

[...]

PR #407, opened 2022-12-28
Add int to the between x/y values, and don't treat as a literal-string

craigfrancis:
For [Issue 367](https://github.com/phpstan/phpstan-doctrine/issues/367)


ondrejmirtes:
Thank you!

PR #408, opened 2023-01-07
Fix deprecations in tests

VincentLanglet:


ondrejmirtes:
Please keep deleting until it passes.

VincentLanglet:
> Please keep deleting until it passes.

I found a way to avoid deleting those tests.
Since there are specific part of the PHPStan extensions to supports `locate` or `sqrt`, I thought it could be better to have test for those method.

I just had to add some logic about what I expect (error/deprecation/nothing).

Please tell me if it's ok like this or if you prefer to delete those tests.

ondrejmirtes:
Do you have an understanding why this error is happening?

tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
```diff
@@ -1233,6 +1235,12 @@ public function getTestData(): iterable
 							SQRT(1)
 				FROM		QueryResult\Entities\Many m
 			',
+			InstalledVersions::satisfies(new VersionParser(), 'doctrine/dbal', '<3.4') && PHP_VERSION_ID >= 80100
+				? 'sqrt(): Passing null to parameter #1 ($num) of type float is deprecated'
+				: null,
+			InstalledVersions::satisfies(new VersionParser(), 'doctrine/dbal', '>=3.4') && PHP_VERSION_ID >= 80100
+				? 'sqrt(): Passing null to parameter #1 ($num) of type float is deprecated'
+				: null,
```
ondrejmirtes, line 1243: What does this code mean? Does this mean that with one DBAL version, it throws an exception, with a different version it throws a deprecation?

ondrejmirtes (changes requested):

tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
VincentLanglet: That's the result I got indeed.

With older versions of DBAL:
```
Doctrine\DBAL\Exception: An exception occurred while executing 'SELECT SQRT(m0_.intColumn) AS sclr_0, SQRT(NULLIF(m0_.intColumn, 1)) AS sclr_1, SQRT(1) AS sclr_2 FROM Many m0_':

sqrt(): Passing null to parameter #1 ($num) of type float is deprecated
```

This is because 
- In old versions of DBAL a catch Throwable is done
- With recent version, it's catching a specific exception

And the logic behind the 
```
convertDeprecationsToExceptions="true"
```
option is to throw a `DeprecatedException` when a deprecation occur so it's was caught by doctrine try/catch in older versions.

But I was wrong about the version constraint. This changed between dbal 2 and dbal 3.

ondrejmirtes: Yeah but why is it an error at all? Whose bug is it? Are we calling it correctly? Is there something to fix in Doctrine?

ondrejmirtes: Isn't it about sample data inserted into the database? Can we fix the inserted data?

VincentLanglet: It's because of the `SQRT(NULLIF(m.intColumn, 1))` part which is calling `sqrt` on `null` in the Dbal implementation.
This was added to check that PHPStan resolve correctly the type to add `|null`.

The deprecation need to be fixed in dbal.
But since they are changing the implementation in 4.0 I'm not sure it's useful.

ondrejmirtes:
Thank you very much!

PR #412, opened 2023-01-08

VincentLanglet:
Follow up of https://github.com/phpstan/phpstan-doctrine/pull/404

[Fix SingleScalarResult](https://github.com/phpstan/phpstan-doctrine/pull/412/commits/89797b8e3e5880775428d092d94901735e56a166): to fix 
```
$eventsCount = $this->entityManager->createQueryBuilder()
			->select('COUNT(cle.id)')
			->from(CountryLocalityEvent::class, 'cle')
			->andWhere('cle.user = ?', $user)
			->andWhere('cle.expirationTime > ?', DateTimeProvider::now())
			->getQuery()
			->getSingleScalarResult();

		return $eventsCount > 0;
```

[Fix ArrayResult](https://github.com/phpstan/phpstan-doctrine/pull/412/commits/4337cd4aacd5a3d6a1992d491ccad31d9a1bf2ba): to fix
```
			->select('d.name, d.id')
			->from(Destination::class, 'd')
			->addOrderBy('d.name')
			->getQuery()
			->getArrayResult();

		$destinations = [];
		foreach ($result as $item) {
			$destinations[(string) $item['id']->toString()] = (string) $item['name'];
		}
```

For the last issue @ondrejmirtes, 
```
			->select('IDENTITY(n.related)')
			->from(DistrictNeighbour::class, 'n')
			->andWhere('n.center = ?', $districtId)
			->getQuery();

		return array_map(static fn (array $row): UuidInterface => $row['related'], $query->getArrayResult());
```
I wonder if it couldn't be a bug in your codebase.
- There is a Test for the Identity function checking it's indexed by 1, 2, 3, ... https://github.com/phpstan/phpstan-doctrine/blob/1.3.x/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php#L1394-L1419
- Expectations are double-checked with the result from a database https://github.com/phpstan/phpstan-doctrine/blob/1.3.x/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php#L215-L233


VincentLanglet:
> For the last issue @ondrejmirtes,
> 
> ```
> $query = $this->entityManager->createQueryBuilder()
> 			->select('IDENTITY(n.related)')
> 			->from(DistrictNeighbour::class, 'n')
> 			->andWhere('n.center = ?', $districtId)
> 			->getQuery();
> 
> 		return array_map(static fn (array $row): UuidInterface => $row['related'], $query->getArrayResult());
> ```
> 
> I wonder if it couldn't be a bug in your codebase.
> 
> * There is a Test for the Identity function checking it's indexed by 1, 2, 3, ... [`1.3.x`/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php#L1394-L1419](https://github.com/phpstan/phpstan-doctrine/blob/1.3.x/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php?rgh-link-date=2023-01-08T20%3A01%3A35Z#L1394-L1419)
> * Expectations are double-checked with the result from a database [`1.3.x`/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php#L215-L233](https://github.com/phpstan/phpstan-doctrine/blob/1.3.x/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php?rgh-link-date=2023-01-08T20%3A01%3A35Z#L215-L233)

I tried your query in my codebase and I get results indexed by 1 rather than `related`.
So to me, your method has a bug @ondrejmirtes which is not found by phpstan.

If you have time to take a new look at this fixed-PR. :)

VincentLanglet:
I understand it's not an easy PR to review/try, but if you find time to take a new look it would be great @ondrejmirtes.

Specifically because you reported in https://github.com/phpstan/phpstan-doctrine/pull/404#issuecomment-1374832594
that
```
			->select('IDENTITY(n.related)')
			->from(DistrictNeighbour::class, 'n')
			->andWhere('n.center = ?', $districtId)
			->getQuery();

		return array_map(static fn (array $row): UuidInterface => $row['related'], $query->getArrayResult());
```
was reported as an error ; and I believe there might be an error in your code base rather than in my PR
since my PR is consistent with the existing test
https://github.com/phpstan/phpstan-doctrine/blob/1.3.x/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php#L1394-L1419
https://github.com/phpstan/phpstan-doctrine/blob/1.3.x/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php#L215-L233

So I would need a confirmation from you
(and also need to know if the 108 new errors are fixed or if there is others I need to fix).

Thanks.

ondrejmirtes:
Alright, I'm gonna merge this, try it on a few real-world projects, and either release it, or report back the errors. Thank you!

VincentLanglet:
> Alright, I'm gonna merge this, try it on a few real-world projects, and either release it, or report back the errors. Thank you!

Thanks ! :)
I tried this PR on my work project, so far I got no false positive.
I'll be available to fix as quickly as possible any false error reported after this PR.

ondrejmirtes:
One more failure:

```php
    /**
     * @return array{id: int, x: mixed, y: string, z: int, zz: string}|null
     */
    private function b(): ?array
    {
        return $this->getEntityManager()->createQueryBuilder()
            ->from(xyz::class, 'abc')
            ->select('abc')
            ->getQuery()
            ->getOneOrNullResult(Query::HYDRATE_ARRAY);
    }
```

Here PHPStan complains:

```php
  157    Method
         a::b()
         should return array{id: int, x: mixed, y: string, z: int,
         zz: string}|null but returns array|null.
```

PHPStan should be able to look at entity fields and figure out that array shape is returned.

VincentLanglet:
> One more failure:
> 
> ```
>     /**
>      * @return array{id: int, x: mixed, y: string, z: int, zz: string}|null
>      */
>     private function b(): ?array
>     {
>         return $this->getEntityManager()->createQueryBuilder()
>             ->from(xyz::class, 'abc')
>             ->select('abc')
>             ->getQuery()
>             ->getOneOrNullResult(Query::HYDRATE_ARRAY);
>     }
> ```
> 
> Here PHPStan complains:
> 
> ```
>   157    Method
>          a::b()
>          should return array{id: int, x: mixed, y: string, z: int,
>          zz: string}|null but returns array|null.
> ```
> 
> PHPStan should be able to look at entity fields and figure out that array shape is returned.

This is kinda "tested" with https://github.com/phpstan/phpstan-doctrine/commit/03741b41dad3095a080e817c6f75245e627bfdff#diff-28394df5dba3e85fe86f9c8d9b05d6087fe897d0efa1d0c0ad3533a1f0252b99R158
and is a consequence of
https://github.com/phpstan/phpstan-doctrine/commit/03741b41dad3095a080e817c6f75245e627bfdff#diff-d465b53a864937dbadc4795d83fc507747a0d603b3f517132f05c7098ffea8b7R204

I don't master enough this lib to look at entity fields in order to transform the object to an array shape (thanks to doctrine and object manager). Does something like this already exists in this plugin ?

I understand `Array<Mixed, Mixed>` can create issues...
One option, as a first step, could be to revert `Array<Mixed, Mixed>` to `Mixed` in case of object.
I don't think there is something like a `BenevolentArray` which would:
- Return no error for `@return array`, `@return array{foo: string}`, ...
- Return an error for `@return Foo`

Is there ? This could be useful here...

ondrejmirtes:
> I don't master enough this lib to look at entity fields in order to transform the object to an array shape (thanks to doctrine and object manager). Does something like this already exists in this plugin ?

I don't know what the logic is in Doctrine to transform entity into an array with array hydration.

But you can get all there is to know about an entity with `ObjectMetadataResolver`, especially getClassMetadata. So you should look at the logic in Doctrine itself and replicate it here.

VincentLanglet:
> But you can get all there is to know about an entity with `ObjectMetadataResolver`, especially getClassMetadata. So you should look at the logic in Doctrine itself and replicate it here.

Thanks, I find a solution with https://github.com/phpstan/phpstan-doctrine/pull/443
But I currently have an issue with multi-selects.

Until I find a solution, if you prefer, I provided a general fix https://github.com/phpstan/phpstan-doctrine/pull/444 which just revert this PR in the special case of getArrayResult with entities in the select.

I'll prioritize fixing all the false-positive before improving the QueryResult analysis.
I'll work on https://github.com/phpstan/phpstan-src/pull/2209#issuecomment-1497070699 now, but it's not easy since I still didn't reproduce the error...

VincentLanglet:
> On one real world project, it shows 101 new errors.
> 
> One example:
> 
> ```
> 	/**
> 	 * @return array<int, array<string, int>>
> 	 */
> 	public function getIdsByStatus(BankTransactionStatus $status): array
> 	{
> 		return $this->entityManager->createQueryBuilder()
> 			->select('bt.id')
> 			->from(BankTransaction::class, 'bt')
> 			->andWhere('bt.status = :status')->setParameter('status', $status)
> 			->addOrderBy('bt.id', 'ASC')
> 			->getQuery()
> 			->getArrayResult();
> 	}
> ```
> 
> PHPStan reports:
> 
> ```
>  ------ -------------------------------------------------------------------------------------------
>   Line   app/services/Bank/BankTransactionRepository.php
>  ------ -------------------------------------------------------------------------------------------
>   473    Method Slevomat\Bank\BankTransactionRepository::getIdsByStatus() should return array<int,
>          array<string, int>> but returns array<int, array<string, int|null>>.
>  ------ -------------------------------------------------------------------------------------------
> ```
> 
> The primary key really can't be null. The definition:
> 
> ```
> 	#[ORM\Id]
> 	#[ORM\GeneratedValue]
> 	#[ORM\Column(type: 'big_integer', options: ['unsigned' => true])]
> 	#[Get]
> 	private ?int $id = null;
> ```

I tried multiple times on my project similar queries and:
I correctly get 
```
Dumped type: array<int, array{id: numeric-string}>
```
With `bigint` witch is the typed used by ORM
```
#[ORM\Id]
#[ORM\GeneratedValue]
#[ORM\Column(type: 'bigint', options: ['unsigned' => true])]
private ?int $id = null;
```
(Notice a bigint is returned as a string by Doctrine and not an int).

With `big_integer`:
```
#[ORM\Id]
#[ORM\GeneratedValue]
#[ORM\Column(type: 'big_integer', options: ['unsigned' => true])]
private ?int $id = null;
```
I get
```
Dumped type: array<int, array{id: mixed}>
```

Is it a custom type @ondrejmirtes ?

ondrejmirtes:
Good news - the problem with nullable primary key is project-specific problem that I was able to fix.

I'm gonna try to reproduce another problem I'm seeing.

ondrejmirtes:
See https://github.com/phpstan/phpstan-doctrine/pull/446

Issue #302, opened 2022-03-16
Why the object of a particular class can not be passed as argument when `mixed` expected?


Anamelash:
Why the object of a particular class can not be passed as argument when `mixed` expected?

```php
class Foo
{
    /**
     * @phpstan-var array<string, mixed>
     */
    protected $parameters;

    /**
     * @return ArrayCollection|Parameter[]
     * @phpstan-return ArrayCollection<array-key, Parameter>
     */
    public function getParameters(): ArrayCollection
    {
        $parameters = [];
        foreach ($this->parameters as $name => $value) {
            $parameters[] = new Parameter($name, $value);
        }

        return new ArrayCollection($parameters);
    }
}
```

```php
$query = $this->entityManager->createNativeQuery('some query');
$query->setParameters($foo->getParameters());
```

Here is error text:
```
Parameter #1 $parameters of method Doctrine\ORM\AbstractQuery<mixed>::setParameters() expects 
array|Doctrine\Common\Collections\ArrayCollection<(int|string), mixed>, 
Doctrine\Common\Collections\ArrayCollection<(int|string), Doctrine\ORM\Query\Parameter> given.
```

ossinkine:
Hi @ondrejmirtes,

The issue is [`setParameters`](https://github.com/doctrine/orm/blob/d2206152bbe9cdc4df0c6da1d328833b6cbb4091/lib/Doctrine/ORM/AbstractQuery.php#L348) already receives `ArrayCollection<(int|string), Parameter>` but this extension brokes this. If you disable this extension all work fine.

ossinkine:
Looks like [`AbstractQuery` stub](https://github.com/phpstan/phpstan-doctrine/blob/9a1711a659829590f2ceb928674565258e176cbc/stubs/ORM/AbstractQuery.stub) is not relevant anymore.

PR #419, opened 2023-01-19
Remove AbstractQuery::setParameters stub

ossinkine:
Original code already has [correct type hints]( https://github.com/doctrine/orm/blob/ba7387fd8c310087b5bc3e176d67671b9739e28c/lib/Doctrine/ORM/AbstractQuery.php#L347) so the stub is redundant.

Ref https://github.com/phpstan/phpstan-doctrine/issues/302#issuecomment-1071116180

### DIFF
--- a/stubs/ORM/AbstractQuery.stub
+++ b/stubs/ORM/AbstractQuery.stub
@@ -10,13 +10,4 @@ use Doctrine\Common\Collections\ArrayCollection;
  */
 abstract class AbstractQuery
 {
-
-	/**
-	 * @param ArrayCollection<array-key, mixed>|array<mixed> $parameters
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
The correct typehint is there only since https://github.com/doctrine/orm/commit/f4524a8bb0a95f544e4469d7940f7d558b8598d8 which means only since version 2.7.4. The stub file is still valuable for older versions.

ossinkine:
@ondrejmirtes The package has dev requirement `doctrine/orm ^2.11` so looks like versions below 2.7.4 is not supported. Also we can change version in `conflict` section to avoid install old versions.

PR #427, opened 2023-02-14
Fix the support for Gedmo attributes

stof:
Attributes should be supported even without an annotation reader, as they are usable without doctrine/annotations

### DIFF
--- a/src/Rules/Gedmo/PropertiesExtension.php
+++ b/src/Rules/Gedmo/PropertiesExtension.php
@@ -72,10 +72,6 @@ class PropertiesExtension implements ReadWritePropertiesExtension
 	 */
 	private function isGedmoAnnotationOrAttribute(PropertyReflection $property, string $propertyName, array $classList): bool
 	{
-		if ($this->annotationReader === null) {
-			return false;
-		}
-
 		$classReflection = $property->getDeclaringClass();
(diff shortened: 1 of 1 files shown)


src/Rules/Gedmo/PropertiesExtension.php
```diff
@@ -72,12 +72,19 @@ public function isInitialized(PropertyReflection $property, string $propertyName
...
 	{
-		if ($this->annotationReader === null) {
+		$classReflection = $property->getDeclaringClass();
+		if ($this->objectMetadataResolver->isTransient($classReflection->getName())) {
 			return false;
 		}
 
-		$classReflection = $property->getDeclaringClass();
-		if ($this->objectMetadataResolver->isTransient($classReflection->getName())) {
+		$attributes = $propertyReflection->getAttributes();
```
mbabker, line 82:
```suggestion
		$propertyReflection = $classReflection->getNativeReflection()->getProperty($propertyName);

		$attributes = $propertyReflection->getAttributes();
```

Need to move the property reflector up (then remove from [here](https://github.com/phpstan/phpstan-doctrine/pull/427/files#diff-1d1d864e71324b0342c2e8faec3c88eefd2673ae50572bec85b0dfb22b61e910R91))

stof:
@mbabker fixed

ondrejmirtes:
Thank you.

PR #430, opened 2023-03-06
Drop inline `@var`

staabm:
phpstan is able to infer this types without additional magic required


ondrejmirtes:
It's not, phpstan-nette isn't here. Please verify it with `\PHPStan\dumpType()`.

staabm:
grafik

it works [because of the generics involved](https://github.com/phpstan/phpstan-src/blob/cfb784a8a5eaefb62b483f0bb348488b389e872b/src/DependencyInjection/Container.php#L16-L22). no extension required.

ondrejmirtes:
Oh right, thank you!

PR #431, opened 2023-03-06
Use ReflectionProvider instead of Broker

staabm:


ondrejmirtes:
Thank you.

ondrejmirtes:
I just cursed at the screen a little bit: https://github.com/phpstan/phpstan-doctrine/commit/3dbb999bd56cc21d6733bab60898ebdf388245b5 (found out about this after releasing)

staabm:
uff.. sorry about that. interessting that this did not fail any unit test and phpstan also did not catch any error..

PR #437, opened 2023-03-17
Add support for union of class string in getRepositoryDynamicReturnType

VincentLanglet:


src/Type/Doctrine/GetRepositoryDynamicReturnTypeExtension.php
```diff
@@ -96,22 +97,26 @@ public function getTypeFromMethodCall(
...
+				return $this->getDefaultReturnType($scope, $methodCall->getArgs(), $methodReflection, $defaultRepositoryClass);
+			}
+
+			$repositoryTypes[] = new GenericObjectType($repositoryClass, [$classType]);
 		}
 
-		return new GenericObjectType($repositoryClass, [
-			$classType,
-		]);
+		return TypeCombinator::union(...$repositoryTypes);
```
staabm, line 119: not sure its relevant here, but its a mistake I do myself again and again.

`TypeCombinator::union(...$array)` will return `NeverType` when `$array` is empty.
it might make sense to check non-empty-ness before calling `TypeCombinator::union` and return the default-return-type in this case

VincentLanglet: There is already
```
if (count($objectNames) === 0) {
			return new GenericObjectType(
				$defaultRepositoryClass,
				[$classType]
			);
		}
```
before the foreach, so the `$repositoryTypes` is non empty ; no ?

ondrejmirtes:
Thank you!

Issue #436, opened 2023-03-17
This extension represents an incorrect type mismatch for the simple_array type


stof:
The `simple_array` type converts `NULL` database values to an empty array: https://github.com/doctrine/dbal/blob/3.6.1/src/Types/SimpleArrayType.php#L53-L57
This means that a nullable column does not mean that the PHP property needs to be nullable.

This would probably require a way for type descriptors to tell whether they have a special handling of `null` or no.

ondrejmirtes:
What's your code and what error phpstan-doctrine reports?

stof:
the code looks like this:

```

    /**
     * @var list<string>
     *
     * @ORM\Column(type="simple_array", nullable=true)
     */
    private $subnetFilters = [];
```

phpstan-doctrine report this:

```
Property Incenteev\WebBundle\Entity\Organization::$subnetFilters type mapping mismatch: database can contain array|null but property expects list<string>.
```

The `array` vs `list<string>` part is easy to solve (the type description is not precise enough right now, but SimpleArrayType indeed returns a `list<string>`).
But the nullability part is handled at https://github.com/phpstan/phpstan-doctrine/blob/62bd362b432fe29e175168689510ddd927b698f8/src/Type/Doctrine/Query/QueryResultTypeWalker.php#L1342C8-L1365 (and maybe other places). The TypeDescriptor interface currently describes only the type for non-null database values and the code elsewhere assumes that a `NULL` value from the DB is converted to `null` in PHP all the type. But this assumption is wrong (most types work this way but not all).
I guess we should have `TypeDescriptor::getDatabaseNullType(): Type` returning the type used to represent the NULL value.

ondrejmirtes:
I get it, your suggestion makes sense, but I see Doctrine's behaviour as very surprising here. You can also get rid of this mess if you make the column non-nullable and represent empty arrays not in two ways (null and []), but only a single way ([]).

stof:
@ondrejmirtes in the DB, an empty array is always represented as NULL, not as an empty string: https://github.com/doctrine/dbal/blob/3.6.1/src/Types/SimpleArrayType.php#L37-L44

So making the DB field non-nullable actually forbids storing an empty array.

ondrejmirtes:
This leads to more constraints that should be reported here: Non-nullable column should only be combined with non-empty-array

stof:
[...]

PR #438, opened 2023-03-17
Improve the type descriptor for SimpleArrayType

stof:
When converting database values to PHP, this type always produces a `list<string>`

### DIFF
--- a/src/Rules/Doctrine/ORM/EntityColumnRule.php
+++ b/src/Rules/Doctrine/ORM/EntityColumnRule.php
@@ -160,13 +160,14 @@ class EntityColumnRule implements Rule
 			return [];
 		}
 
-		$propertyTransformedType = TypeTraverser::map($propertyType, static function (Type $type, callable $traverse): Type {
+		// If the type descriptor does not precise the types inside the array, don't report errors if the field has a more precise type
+		$propertyTransformedType = $writableToPropertyType->equals(new ArrayType(new MixedType(), new MixedType())) ? TypeTraverser::map($propertyType, static function (Type $type, callable $traverse): Type {
 			if ($type instanceof ArrayType) {
 				return new ArrayType(new MixedType(), new MixedType());
 			}
 
 			return $traverse($type);
-		});
+		}) : $propertyType;
 
 		if (!$propertyTransformedType->isSuperTypeOf($writableToPropertyType)->yes()) {
 			$errors[] = sprintf(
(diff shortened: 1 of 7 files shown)


stof:
@ondrejmirtes any chance to merge this ? Even though it does not solve all issues related to this SimpleArrayType (due to #436), it improves the situation.

ondrejmirtes:
This needs a test to understand what kind of error is this preventing. A new method in EntityColumnRuleTest should be fine.

stof:
@ondrejmirtes I added more tests but they fail. It looks like the `@var list<string>` is transformed into `array<int, mixed>` somewhere when checking the EntityColumnRule.

stof:
And I found the fix for that issue.

src/Rules/Doctrine/ORM/EntityColumnRule.php
```diff
@@ -160,13 +160,14 @@ public function processNode(Node $node, Scope $scope): array
 			return [];
 		}
 
-		$propertyTransformedType = TypeTraverser::map($propertyType, static function (Type $type, callable $traverse): Type {
+		// If the type descriptor does not precise the types inside the array, don't report errors if the field has a more precise type
```
ondrejmirtes, line 163: I disagree with this. If Doctrine can write any array to a property, it's wrong to expect a narrower type.

ondrejmirtes (changes requested):

src/Rules/Doctrine/ORM/EntityColumnRule.php
stof: This comment is about explaining the *existing* logic that you added in https://github.com/phpstan/phpstan-doctrine/commit/e745610aa35228935edf6e7fbf79e02d505c37a6

The change I'm doing here is *disabling* this logic for some cases (when the type descriptor is precise), not enabling it, to avoid reporting errors that are wrong.

I tried to remove this transformation entirely, but this broke some existing tests (the ones added in that commit)

stof: Note however that the only type descriptors that return `array` as type (and so would trigger this logic) are `json_array` (deprecated in DBAL 2.10 and removed in DBAL 3.0) and `array` (deprecated in DBAL 3.4+).

So maybe we could drop this logic and return `NeverType` as done in the JsonType descriptor

ondrejmirtes: Oh I get it now, it's nice 👍 

I'm not rushing to remove functionality, people use phpstan-doctrine with years-old versions of Doctrine.

ondrejmirtes:
Please fix the build, otherwise good to go.

stof:
@ondrejmirtes the build is now green

ondrejmirtes:
Thank you!

stof:
@ondrejmirtes do you have any idea when this will be released ?

ondrejmirtes:
I had to revert some blocking work that wasn't in ideal state, and I can now release this.

ondrejmirtes:
Hey, I've got some false positives because of this PR, can you please tell me what should be done?

```
 ------ ----------------------------------------------------------------------------------------------
  Line   app/services/PointOfInterest/PointOfInterest.php
 ------ ----------------------------------------------------------------------------------------------
  109    Property Slevomat\PointOfInterest\PointOfInterest::$seasonMonths type mapping mismatch:
         database can contain list<string> but property expects array<int, float|int|numeric-string>.
  109    Property Slevomat\PointOfInterest\PointOfInterest::$seasonMonths type mapping mismatch:
         property can contain array<int, float|int|string> but database expects array<string>.
 ------ ----------------------------------------------------------------------------------------------

 ------ ------------------------------------------------------------------------------------------------
  Line   app/services/Product/Cancel/ProductCancel.php
 ------ ------------------------------------------------------------------------------------------------
  99     Property Slevomat\Product\Cancel\ProductCancel::$goodsStatuses type mapping mismatch: database
         can contain array<int, string>|null but property expects array<int, int>|null.
  99     Property Slevomat\Product\Cancel\ProductCancel::$goodsStatuses type mapping mismatch: property
         can contain array<int, int>|null but database expects array<string>|null.
 ------ ------------------------------------------------------------------------------------------------
```

For these properties:

```php
	/** @var array<int, numeric> */
	#[ORM\Column(type: 'simple_array')]
	private array $seasonMonths = [];

	/** @var array<int, int>|null */
	#[ORM\Column(type: 'simple_array')]
	private ?array $goodsStatuses = null;
```

stof:
This reports an actual issue. When loaded from the DB, the `simple_array` does a `explode()`, so values in the array are strings, not integers. So there is indeed a type mismatch.

See https://github.com/doctrine/dbal/blob/47589fdf56929f1510fd9b394c328c31a7ce13fb/src/Types/SimpleArrayType.php#L53-L62 for the conversion code

ondrejmirtes:
Thank you!

PR #444, opened 2023-04-05
Avoid false positive about getArrayResult when selecting objects

VincentLanglet:
This would be a ok quick-fix for the false positive https://github.com/phpstan/phpstan-doctrine/pull/412#issuecomment-1497092934

It's basically a revert of the previous PR but just in the case of object provided in the `getArrayResult`.

I'm working on a long-term fix/improvement with https://github.com/phpstan/phpstan-doctrine/pull/443 but it's not easy for multi-select queries.


ondrejmirtes:
Thank you.

PR #446, opened 2023-04-05
Query type inference failures reproductions

ondrejmirtes:


ondrejmirtes:
Alright @VincentLanglet, here's one:

```
--- Expected
+++ Actual
@@ @@
-'list<array{id: int}>'
+'list<array{id: int|numeric-string|null}>'
```

VincentLanglet:
> Alright @VincentLanglet, here's one:
> 
> ```
> --- Expected
> +++ Actual
> @@ @@
> -'list<array{id: int}>'
> +'list<array{id: int|numeric-string|null}>'
> ```

If I dump the Query, I get
```
Doctrine\ORM\Query<null, array{id: int|numeric-string|null}>
```
so `QueryResultDynamicReturnTypeExtension` works fine but the issue seems to be with `QueryResultTypeWalker`

This is a consequence of https://github.com/phpstan/phpstan-doctrine/pull/273 (cc @arnaud-lb)
and there is the following test case.
```
'identity function' => [
				$this->constantArray([
					[new ConstantIntegerType(1), TypeCombinator::addNull($this->numericStringOrInt())],
					[new ConstantIntegerType(2), $this->numericStringOrInt()],
					[new ConstantIntegerType(3), TypeCombinator::addNull($this->numericStringOrInt())],
					[new ConstantIntegerType(4), TypeCombinator::addNull(new StringType())],
					[new ConstantIntegerType(5), TypeCombinator::addNull(new StringType())],
					[new ConstantIntegerType(6), TypeCombinator::addNull($this->numericStringOrInt())],
					[new ConstantIntegerType(7), TypeCombinator::addNull(new MixedType())],
					[new ConstantIntegerType(8), TypeCombinator::addNull($this->numericStringOrInt())],
				]),
				'
					SELECT		IDENTITY(m.oneNull),
								IDENTITY(m.one),
								IDENTITY(m.oneDefaultNullability),
								IDENTITY(m.compoundPk),
								IDENTITY(m.compoundPk, \'id\'),
								IDENTITY(m.compoundPk, \'version\'),
								IDENTITY(m.compoundPkAssoc),
								IDENTITY(m.compoundPkAssoc, \'version\')
				',
			],
```

I'll need to investigate why `numericStringOrInt` is used

VincentLanglet:
I identified that the `numeric-string` is because of the following code
https://github.com/phpstan/phpstan-doctrine/blob/a1ba4549d2f4167adae5a91871b0fc88fe66240d/src/Type/Doctrine/Query/QueryResultTypeWalker.php#L837-L855

VincentLanglet:
I try finding some "solutions" for this issue.

But I discovered a new one.
```
			->select('p.id AS id')
			->join('oi.product', 'p')
			->getQuery()
			->getArrayResult();
		assertType('list<array{id: int}>', $result);
```
Works

```
			->select('p.id AS id')
			->leftJoin('oi.product', 'p')
			->getQuery()
			->getArrayResult();
		assertType('list<array{id: int|null}>', $result);
```
Also works But
```
			->select('p.id AS id')
			->leftJoin('oi.product', 'p')
			->where('p.id IS NOT NULL')
			->getQuery()
			->getArrayResult();
		assertType('list<array{id: int}>', $result);
```
doesn't.

The `where` condition are not supported, and this may be too complicated to support them.

So far, I think I'll have to change every Union to BenevolentUnion then...

VincentLanglet:
I will need your opinion on this solution @ondrejmirtes https://github.com/phpstan/phpstan-doctrine/pull/447

ondrejmirtes:
Continued in https://github.com/phpstan/phpstan-doctrine/pull/593

PR #447, opened 2023-04-05
Use benevolent union for scalar in queries

VincentLanglet:
This PR could be an idea of solution for https://github.com/phpstan/phpstan-doctrine/pull/446

I assume that before https://github.com/phpstan/phpstan-doctrine/pull/412, the resultType of Query were less used a lot so we didn't have a lot of feedback about the type resolution. But I tried to reproduce your "bug", and even the merge of my PR
```
			->select('DISTINCT IDENTITY(oi.product) AS id')
			->join('oi.product', 'p')
			->getQuery()
			->getResult();
```
were reporting `list<array{id: int|numeric-string|null}>`. I just give the same result with `getArrayResult` now.

For Identity, I found some specific solution to improve the return type,
but still, I anticipate a lot of similar issue with results like
```
			->select('p.id AS id')
			->leftJoin('oi.product', 'p')
			->where('p.id IS NOT NULL')
			->getQuery()
			->getArrayResult();
```
which is reported as `list<array{id: int|null}>` when we would expect `list<array{id: int}>`. Notice that in the same way, before https://github.com/phpstan/phpstan-doctrine/pull/412, the query was already returning `list<array{id: int|null}>` for `getResult` calls.

Since we currently cannot be sure that all those `|null` are true because, for example, we don't check the `WHERE` condition (and this would be complex). I think the best solution so far would be to use BenevolentUnion for this instead of default unions.

Before fixing the tests I'd like your opinion on such change.

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
@@ -10,6 +10,7 @@ use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Accessory\AccessoryArrayListType;
 use PHPStan\Type\ArrayType;
+use PHPStan\Type\BenevolentUnionType;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\Doctrine\ObjectMetadataResolver;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
@@ -21,6 +22,7 @@ use PHPStan\Type\ObjectWithoutClassType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeTraverser;
+use PHPStan\Type\TypeUtils;
 use PHPStan\Type\TypeWithClassName;
 use PHPStan\Type\VoidType;
 use function count;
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Hi, I need tests to understand what will this fix :) I'd say we need benevolent union even for the case without `WHERE` because the `int|numeric-string` for `int` keys is really annoying.

VincentLanglet:
> I'd say we need benevolent union even for the case without `WHERE` because the `int|numeric-string` for `int` keys is really annoying.

I added benevolent union in two situations so far:
- When there are `WHERE` condition in the queries
- When expression (like IDENTITY) are casted to string because we don't know how the database will behave (your case of `int|numeric-string`).

I updated the tests (about int|numeric-string) and added one test about where conditions @ondrejmirtes.

VincentLanglet:
> I added benevolent union in two situations so far:
> 
> * When there are `WHERE` condition in the queries
> * When expression (like IDENTITY) are casted to string because we don't know how the database will behave (your case of `int|numeric-string`).
> 
> I updated the tests (about int|numeric-string) and added one test about where conditions @ondrejmirtes.

Does this solution seems ok to you @ondrejmirtes ?

This should avoid false positive from https://github.com/phpstan/phpstan-doctrine/pull/412 but still be an improvement of query results. And possibly allow you to make new releases, unless you find others issues. :)

ondrejmirtes:
FYI I had to revert some of your commits so I can release phpstan-doctrine with other changes. Unfortunately I currently cannot give this the work priority and the needed timeframe of multiple days in order to finish this. https://github.com/phpstan/phpstan-doctrine/compare/84bf895dd82bf732f65481fc25fcfd9fc9b7b8cf...f1499e54358052fef17d349881ee87a7c4cc4c67

VincentLanglet:
> FYI I had to revert some of your commits so I can release phpstan-doctrine with other changes. Unfortunately I currently cannot give this the work priority and the needed timeframe of multiple days in order to finish this. [84bf895...f1499e5](https://github.com/phpstan/phpstan-doctrine/compare/84bf895dd82bf732f65481fc25fcfd9fc9b7b8cf...f1499e54358052fef17d349881ee87a7c4cc4c67)

I understand.

I recreated the PR with the commits reverted https://github.com/phpstan/phpstan-doctrine/pull/453 then

PR #448, opened 2023-04-07
Prevent case mismatch between MySQLPlatform and MySqlPlatform

FlashBlack:
In conjunction with `symfony/debug` when the debug mod is enabled, the platform existence check from dbal v2 `MySqlPlatform` generates a lot of errors in output.

Who generated errors:

- method [DebugClassLoader::checkClass#L221](https://github.com/symfony/debug/blob/1a692492190773c5310bc7877cb590c04c2f05be/DebugClassLoader.php#L221) throw exceptions on class case mismatch. 

I propose that the dbal version should be checked instead class_exists.

### DIFF
--- a/src/Doctrine/Mapping/ClassMetadataFactory.php
+++ b/src/Doctrine/Mapping/ClassMetadataFactory.php
@@ -2,16 +2,18 @@
 
 namespace PHPStan\Doctrine\Mapping;
 
+use Composer\InstalledVersions;
 use Doctrine\Common\Annotations\AnnotationReader;
 use Doctrine\Common\Annotations\DocParser;
 use Doctrine\Common\EventManager;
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Hi, I don't understand what error are you experiencing with PHPStan and what does this fix. Please elaborate.

FlashBlack:
@ondrejmirtes This only applies indirectly to phpstan. I want to prevent this problem by replacing the platform selection condition with a dbal version check.

Trying to clarify it: there is a problem with name compatibility with different case.
Because there is a MySQLPlatform file and the import wants to find MySqlPlatform, symfony-debug throws exception of name case mismatch

ondrejmirtes:
Do you have a test in this repo that fails under certain conditions for you and thus you need this change? If not then there's no need for this PR.

PR #461, opened 2023-06-26
Add missing throws in stubs

VincentLanglet:
Hi @ondrejmirtes,

EntityManagerInterface::getReference throws an ORMException
https://github.com/doctrine/orm/blob/4aadba65cefd4efc89e5946c8b0304656c269a4a/lib/Doctrine/ORM/EntityManagerInterface.php#L182

But this is missing in the stub, so PHPStan reports
```
Foo has Doctrine\ORM\Exception\ORMException in PHPDoc @throws tag but it's not thrown.
```

### DIFF
--- a/stubs/EntityManagerInterface.stub
+++ b/stubs/EntityManagerInterface.stub
@@ -4,6 +4,7 @@ namespace Doctrine\ORM;
 
 use Doctrine\Persistence\ObjectManager;
 use Doctrine\Persistence\ObjectRepository;
+use Doctrine\ORM\Exception\ORMException;
 use Doctrine\ORM\Mapping\ClassMetadata;
 
 interface EntityManagerInterface extends ObjectManager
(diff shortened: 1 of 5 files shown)


stubs/EntityManagerInterface.stub
```diff
@@ -36,6 +37,8 @@ interface EntityManagerInterface extends ObjectManager
 	 * @phpstan-param class-string<T> $entityName
 	 * @phpstan-param mixed $id
 	 * @phpstan-return T|null
+     *
```
ondrejmirtes: Wrong indent

ondrejmirtes (changes requested):

stubs/EntityManagerInterface.stub
VincentLanglet: Fixed

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #466, opened 2023-08-07
Add AbstractQuery::getSingleScalarResult to stub

janlanger:
`AbstractQuery::getSingleScalarResult()` had its phpDoc narrowed from mixed in latest 2.16 version.

However, this phpDoc is missing return null, as the method can return it https://github.com/doctrine/orm/pull/10870

PR #467, opened 2023-08-09
Add missing throws tag

VincentLanglet:
This is often forgotten, dunno if something general could be done...

Fix https://github.com/phpstan/phpstan-doctrine/pull/466


ondrejmirtes:
Thank you.

PR #468, opened 2023-08-28
Add throws annotation about NoResultException on getSingleScalarResult

VincentLanglet:
Hi @ondrejmirtes,

Similar to the doctrine/orm fix: https://github.com/doctrine/orm/pull/10907


ondrejmirtes:
Thank you.

PR #472, opened 2023-09-15
add stub for paginator

84m:
This makes the paginator aware of the type, when created from a query

### DIFF
--- a/stubs/ORM/Tools/Pagination/Paginator.stub
+++ b/stubs/ORM/Tools/Pagination/Paginator.stub
@@ -1,0 +1,33 @@
+<?php
+
+namespace Doctrine\ORM\Tools\Pagination;
+
+use ArrayIterator;
+use Countable;
+use Doctrine\ORM\Query;
+use Doctrine\ORM\QueryBuilder;
+use IteratorAggregate;
+
+/**
+ * @template T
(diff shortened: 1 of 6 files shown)


84m:
This is a follow-up to [this discussion](https://github.com/phpstan/phpstan/discussions/9825).

stubs/ORM/Tools/Pagination/Paginator.stub
```diff
@@ -0,0 +1,33 @@
...
+namespace Doctrine\ORM\Tools\Pagination;
+
+use ArrayIterator;
+use Countable;
+use Doctrine\ORM\Query;
+use Doctrine\ORM\QueryBuilder;
+use IteratorAggregate;
+
+/**
+ * @template T
```
84m, line 12: unlike the [original](https://github.com/doctrine/orm/blob/2.16.x/lib/Doctrine/ORM/Tools/Pagination/Paginator.php#L30) this was changed to invariant because otherwise the return from `getIterator()` in l27 would throw a error

ondrejmirtes:
Merged as https://github.com/phpstan/phpstan-doctrine/commit/cba50e9ad665687547e476ebd877cb4a3c8d798c, thank you.

PR #473, opened 2023-10-02
Support calls over Andx/Orx objects

janedbal:


janedbal:
@ondrejmirtes just revisited this and fixed the issue. Method calls over `\Doctrine\ORM\Query\Expr\Base` were not supported, leaving that exprObject in `ExprType` without any valid data and causing DQL syntax errors.

Rebased to 1.3.x

ondrejmirtes:
Thank you!

PR #477, opened 2023-10-13
Allow doctrine/lexer v2

talentrydev:
Locking to v1 makes it impossible to upgrade certain libraries (e.g. jms/serializer)


ondrejmirtes:
This is only in require-dev. It does not conflict with your other dependencies.

talentrydev:
You're right. Composer was just being stubborn. Works fine now.

PR #483, opened 2023-11-08
Add throws annotation on `EntityManagerInterface::flush`

VincentLanglet:
This is the same than https://github.com/phpstan/phpstan-doctrine/pull/313 but on the EntityManagerInterface in order to avoid the deadcatch with
```
try {
    $this->entityManagerInterface->flush();
} catch (\Doctrine\DBAL\Exception\UniqueConstraintViolationException $e) {
    // pass
}
```
or avoid `but it's not thrown.` phpstan error.

Injecting the interface rather than `EntityManager` is a really common thing when working with Symfony/doctrine (thanks to autowire/autoconfigure) so the `@throws` annotation will be more useful on the interface.

### DIFF
--- a/stubs/EntityManagerInterface.stub
+++ b/stubs/EntityManagerInterface.stub
@@ -60,19 +60,20 @@ interface EntityManagerInterface extends ObjectManager
 	 */
 	public function copy($entity, $deep = false);
 
-    /**
-     * @return void
-     *
-     * @throws ORMException
-     * @throws \Doctrine\DBAL\Exception\UniqueConstraintViolationException
-     */
-    public function flush();
+	/**
+	 * @return void
+	 *
+	 * @throws ORMException
+	 * @throws \Doctrine\DBAL\Exception\UniqueConstraintViolationException
+	 */
+	public function flush();
 
-  /**
-   * @template T of object
-   * @phpstan-param class-string<T> $className
-   *
-   * @phpstan-return ClassMetadata<T>
-   */
-  public function getClassMetadata($className);
+	/**
+	 * @template T of object
+	 * @phpstan-param class-string<T> $className
+	 *
+	 * @phpstan-return ClassMetadata<T>
+	 */
+	public function getClassMetadata($className);
+
 }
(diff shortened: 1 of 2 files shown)


VincentLanglet:
Should I target 1.3.x or 1.4.x @ondrejmirtes ?

ondrejmirtes:
1.3.x

VincentLanglet:
> 1.3.x

Thanks, done

stubs/EntityManagerInterface.stub
```diff
@@ -60,6 +60,14 @@ interface EntityManagerInterface extends ObjectManager
 	 */
 	public function copy($entity, $deep = false);
 
+    /**
+     * @return void
+     *
+     * @throws ORMException
+     * @throws \Doctrine\DBAL\Exception\UniqueConstraintViolationException
+     */
+    public function flush();
```
ondrejmirtes: Please fix indenting of `flush` and `getClassMetadata` in this file, thanks.

ondrejmirtes (changes requested):

stubs/EntityManagerInterface.stub
VincentLanglet: It should be better now

ondrejmirtes:
Thank you.

VincentLanglet:
Hi @ondrejmirtes, I think this PR doesn't work and I'm not sure why.

I'm still getting a dead catch error when writing
```
try {
     $entityManager->flush();
} catch (ORMException|UniqueConstraintViolationException) {}
```
I can reproduce the bug if I had `@throws void` in the ObjectManager class but I cannot reproduce the error in the PHPStan tests yet, even with `implicitThrows: false`...

I think it's because `flush` method is in the ObjectManager 
https://github.com/doctrine/persistence/blob/3.2.x/src/Persistence/ObjectManager.php#L94
and not in the EntityManagerInterface.

Is there a way to say to PHPStan that ORMException|UniqueConstraintViolationException can be thrown by `EntityManagerInterface::flush` without adding those exception to `ObjectManager::flush` ?

ondrejmirtes:
@VincentLanglet It could be done only with a dynamic throw type extension.

VincentLanglet:
> @VincentLanglet It could be done only with a dynamic throw type extension.

Thanks for the advice, I made https://github.com/phpstan/phpstan-doctrine/pull/486
I check the branch on my personal project, it solves indeed the error.

PR #485, opened 2023-11-09
Allow class string in from and join

VincentLanglet:
Doctrine write the following code
```
    /**
     * @psalm-var class-string<T>
     */
    protected $_entityName;

   /**
     * @param string      $alias
     * @param string|null $indexBy The index for the from.
     *
     * @return QueryBuilder
     */
    public function createQueryBuilder($alias, $indexBy = null)
    {
        return $this->_em->createQueryBuilder()
            ->select($alias)
            ->from($this->_entityName, $alias, $indexBy);
    }
```
So they are passing a `class-string<T of object>` to the first from argument.

When doing the same, we get an error
```
Parameter #1 $from of method Doctrine\ORM\QueryBuilder::from() expects literal-string, class-string<T of object> given.
```
I think allowing `class-string|literal-string` doesn't introduce a security issue since no "variable" classname can inject sql.

In the same way `join` methods, can be used in two ways
```
->join('project.organization', ....)
->join(Organization::class, ...)
```
here again, a "variable" classname shouldn't introduce security issues so `class-string|literal-string` should be allowed in order to write
```
->join($someClassName, ...)
```

I didn't work on literal for a long time since https://github.com/phpstan/phpstan-doctrine/pull/327 so I'd be interested in your opinion on this change @craigfrancis.

### DIFF
--- a/stubs/bleedingEdge/ORM/QueryBuilder.stub
+++ b/stubs/bleedingEdge/ORM/QueryBuilder.stub
@@ -58,9 +58,9 @@ class QueryBuilder
 	}
 
 	/**
-	 * @param literal-string      $from
-	 * @param literal-string      $alias
-	 * @param literal-string|null $indexBy
+	 * @param literal-string|class-string $from
+	 * @param literal-string              $alias
+	 * @param literal-string|null         $indexBy
 	 *
 	 * @return $this
 	 */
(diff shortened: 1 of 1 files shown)


craigfrancis:
Thanks @VincentLanglet, I'm absolutely fine with that.

Oddly, I've not seen `::class` being used with these methods before, it makes sense though.

(and yes, a lot of time has passed, I still have a todo note reminding me daily, and its probably causing a bit of screen burn-in; unfortunately I don't have the free time to do further work atm).

ondrejmirtes:
Thank you.

PR #486, opened 2023-11-13
Use dynamic throw type extension on EntityManagerInterface::flush

VincentLanglet:


src/Type/Doctrine/EntityManagerInterfaceThrowTypeExtension.php
```diff
@@ -0,0 +1,49 @@
...
+
+	public function getThrowTypeFromMethodCall(MethodReflection $methodReflection, MethodCall $methodCall, Scope $scope): ?Type
+	{
+		$type = $scope->getType($methodCall->var);
+
+		if ((new ObjectType(EntityManagerInterface::class))->isSuperTypeOf($type)->yes()) {
+			return TypeCombinator::union(
+				...array_map(static function ($class): Type {
+					return new ObjectType($class);
+				}, self::SUPPORTED_METHOD[$methodReflection->getName()])
```
ondrejmirtes, line 42: This needs a test. You can create a new test case for the corresponding rule, and fetch the rule with `self::getContainer()->getByType(....`

ondrejmirtes (changes requested):

src/Type/Doctrine/EntityManagerInterfaceThrowTypeExtension.php
VincentLanglet: This seems covered by 
https://github.com/phpstan/phpstan-doctrine/blob/1.3.x/tests/Rules/Exceptions/data/unthrown-exception.php#L21-L25
to me.

I tried to comment the whole if in the `getThrowTypeFromMethodCall`, and tests were failing.

ondrejmirtes: If your `src/` does not change anything about test results then it's not needed.

Please show why your change is needed by writing a failing test first.

VincentLanglet: I added another test, which fail without the change

VincentLanglet: I added another test which fail without this change.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #488, opened 2023-11-14
QueryBuilderDqlRule: check also other methods if enabled

janedbal:
Without this, I'm getting this error even when type infering works perfectly thanks to other methods analysis. This makes me unable to enable `reportDynamicQueryBuilders: true`.


ondrejmirtes:
Thank you.

ondrejmirtes:
It's alright, I just tested on a real project, performance is the same and the number of "Could not analyse QueryBuilder with unknown beginning" goes down.

PR #489, opened 2023-11-16
QueryBuilder: allow dynamic args in methods not affecting result type

janedbal:
I believe it does not make sense to produce `mixed` just becase this extension does not understand e.g. some `->andWhere` call as this call does not contribute to the inferred result. 

This MR adds a list where we remain strict as before (select, from, indexBy etc) and the rest is allowed to have dynamic arguments.

### DIFF
--- a/src/Type/Doctrine/QueryBuilder/QueryBuilderGetQueryDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/QueryBuilder/QueryBuilderGetQueryDynamicReturnTypeExtension.php
@@ -32,6 +32,21 @@ use function strtolower;
 class QueryBuilderGetQueryDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtension
 {
 
+	/**
+	 * Those are critical methods where we need to understand arguments passed to them, the rest is allowed to be more dynamic
+	 * - this list reflects what is implemented in QueryResultTypeWalker
+	 */
+	const METHODS_AFFECTING_RESULT_TYPE = [
+		'add',
+		'select',
+		'addselect',
+		'from',
+		'join',
+		'innerjoin',
+		'leftjoin',
+		'indexby',
+	];
+
 	/** @var ObjectMetadataResolver */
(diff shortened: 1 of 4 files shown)


janedbal:
Our usecase is that we reuse some common conditions like this:

```php
$queryBuilder = $this->entityManager->createQueryBuilder()
            ->select('l')
            ->from(PickingLocation::class, 'l')
            ->andWhere('l.active = true')
            ->andWhere(WarehouseZoneRestrictor::getLocationZoneRestrictionsDQL('l', $warehouse, $warehouseZone)); // returns Andx
```

src/Type/Doctrine/QueryBuilder/QueryBuilderGetQueryDynamicReturnTypeExtension.php
```diff
@@ -32,6 +32,21 @@
 class QueryBuilderGetQueryDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtension
 {
 
+	/**
+	 * Those are critical methods where we need to understand arguments passed to them, the rest is allowed to be more dynamic
+	 * - this list reflects what is implemented in QueryResultTypeWalker
+	 */
+	private const METHODS_AFFECTING_RESULT_TYPE = [
+		'add',
```
ondrejmirtes, line 40: Somehow we must not forget to update this list when the type inference gets smarter and takes `andWhere` etc. into account.

ondrejmirtes:
Thank you.

ondrejmirtes:
FYI https://github.com/phpstan/phpstan-doctrine/commit/c94f43bd2f11c249c23b8830f54c28025468a08c

The original code was not every maintainable and buggy. For example `set` method affects result type and analysis.

janedbal:
> For example set method affects result type and analysis.

That one is more tricky and was currently producing `Could not analyse QueryBuilder with dynamic arguments.` all the time (any regular usage). This should fix it: https://github.com/phpstan/phpstan-doctrine/pull/498

PR #497, opened 2023-11-20
BaseExpressionDynamicReturnTypeExtension: support only add, addMultiple

### LINKED ISSUE #496: \Doctrine\ORM\Query\Expr\Base::count() doesn't handle dynamic arguments
I have been appreciating the recent PHPStan Doctrine query builder improvements. Thank you!

I found an issue with using `count()` for or and and expressions when arguments are added conditionally:

```php
$or = $qb->expr()->orX();

\PHPStan\dumpType($or->count()); // Dumped type: 0

$or->add('a.foo = 1');

\PHPStan\dumpType($or->count()); // Dumped type: 1

if (rand(0, 1)) {
    $or->add('a.bar = 2');
}

\PHPStan\dumpType($or->count()); // Dumped type: 2 (but should be int<1,2>)
```
It turns out that `if ($or->count()) { $qb->andWhere($or); }` can be replaced with `$qb->andWhere($or)` because the empty expression with be reduced, but using `$or->count()` to return `null` instead of executing a query causes this PHPStan error.

janedbal:
Resolves #496


ondrejmirtes:
Thank you.

PR #498, opened 2023-11-21
QueryBuilderGetQueryDynamicReturnTypeExtension: handle regular set usages

janedbal:


src/Type/Doctrine/QueryBuilder/QueryBuilderGetQueryDynamicReturnTypeExtension.php
```diff
@@ -152,6 +153,18 @@ public function getTypeFromMethodCall(
...
 				}
 
+				if ($lowerMethodName === 'set') {
+					try {
+						$args = $this->argumentsProcessor->processArgs($scope, $methodName, array_slice($calledMethodCall->getArgs(), 0, 1));
+					} catch (DynamicQueryBuilderArgumentException $e) {
+						return $defaultReturnType;
+					}
+					if (count($args) === 1) {
+						$queryBuilder->set($args[0], $args[0]);
```
ondrejmirtes, line 163: This branch does not seem to be tested

ondrejmirtes (changes requested):

src/Type/Doctrine/QueryBuilder/QueryBuilderGetQueryDynamicReturnTypeExtension.php
janedbal: It is. This is the main thing that enables the whole thing. Maybe you missed array_slice above.

janedbal: The idea here is to drop second argument (which is typically an expression) and do "dummy update" `alias.field = alias.field` instead.

ondrejmirtes: Oh alright, I know less and less about this extension every day :D

ondrejmirtes:
Thank you.

PR #503, opened 2023-12-04
Query: add test for some method calls

janedbal:


janedbal:
Hm, this is probably not a proper fix. This is more generic issue. Looks like it is related to all methods that are overwritten in Query and use `{@inheritdoc}`.

janedbal:
Changed to failing test till I find the issue.

janedbal:
Looks like related to https://github.com/phpstan/phpstan/issues/6462

janedbal:
This should fix it: https://github.com/phpstan/phpstan-src/pull/2803

ondrejmirtes:
Please test this after your merged PR https://github.com/phpstan/phpstan-src/pull/2803 is released :)

janedbal:
It works, do you want to bump phpstan version in composer just to have this merged?

ondrejmirtes:
Thank you!

PR #500, opened 2023-11-21
Infer QueryBuilderType for any method returning QB

janedbal:
Resolves https://github.com/phpstan/phpstan-doctrine/issues/499

PR #504, opened 2023-12-06
OtherMethodQueryBuilderParser: prevent broken type infering with partial analysis

janedbal:
Currently, when a file [is not among "analysed files"](https://github.com/phpstan/phpstan-src/blob/1.10.47/src/Parser/PathRoutingParser.php#L44) (which is any file not listed in partial analysis), every method body is [stripped by CleaningParser](https://github.com/phpstan/phpstan-src/blob/1.10.47/src/Parser/CleaningVisitor.php#L29). Due to that [OtherMethodQueryBuilderParser](https://github.com/phpstan/phpstan-doctrine/blob/1.3.53/src/Type/Doctrine/QueryBuilder/OtherMethodQueryBuilderParser.php#L95) cannot determine what is happening to the QueryBuilder there and results in `mixed`. This results in nasty:

- **partial analysis**: `mixed` returned
- **full analysis**: `list<Entity>` returned

Due to the nature of this feature (affecting only methods returning QueryBuilder) I think it is safe to remove the CleaningParser even with [this improvement](https://github.com/phpstan/phpstan-doctrine/pull/500).

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -72,6 +72,21 @@ conditionalTags:
 		phpstan.broker.methodsClassReflectionExtension: %doctrine.allCollectionsSelectable%
 
 services:
+	otherMethodAnalysisParser: # identical to defaultAnalysisParser, but uses inner PathRoutingParser without cleaning parser
+		class: PHPStan\Parser\CachedParser
+		arguments:
+			originalParser: @pathRoutingParser
+			cachedNodesByStringCountMax: %cache.nodesByStringCountMax%
+		autowired: false
+
+	pathRoutingOtherMethodAnalysisParser: # identical to pathRoutingParser, but no cleaning parser is used
+		class: PHPStan\Parser\PathRoutingParser
+		arguments:
+			currentPhpVersionRichParser: @currentPhpVersionRichParser
+			currentPhpVersionSimpleParser: @currentPhpVersionRichParser
+			php8Parser: @php8Parser
+		autowired: false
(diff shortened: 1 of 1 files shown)


extension.neon
```diff
@@ -72,6 +72,21 @@ conditionalTags:
...
 
 services:
+	otherMethodAnalysisParser: # identical to defaultAnalysisParser, but uses inner PathRoutingParser without cleaning parser
+		class: PHPStan\Parser\CachedParser
+		arguments:
+			originalParser: @pathRoutingParser
+			cachedNodesByStringCountMax: %cache.nodesByStringCountMax%
+		autowired: false
+
+	pathRoutingOtherMethodAnalysisParser: # identical to pathRoutingParser, but no cleaning parser is used
```
ondrejmirtes, line 82: I don't get it. Where is this one used?

ondrejmirtes (changes requested):

extension.neon
janedbal: That is obviously a typo, it should be used above. Fixed

ondrejmirtes (changes requested):
I've looked into this and thought about it. The problem is that a lot of files will now be cached twice in memory because here you're using a new instance of the cached parser.

Ideally you'd reuse the existing cache for already parsed files by the rich parser and just cache differently parsed files that are otherwise parsed by the simple parser.

PR #506, opened 2023-12-13
Autodetect driver setup for precise int/float/bool inference in expressions (stringified or not)

janedbal:
Currently, inferring expressions types (e.g. results of `AVG`, `SUM` etc) is imperfect as [it is currently implemented to serve all configurations](https://github.com/phpstan/phpstan-doctrine/blob/1.3.1/src/Type/Doctrine/Query/QueryResultTypeWalker.php#L841-L842) resulting in union types like `int|numeric-string`. This is painful to work with in real codebase with some specific configuration where you know it is always `int`.

This PR introduces **autodetection of driver & PHP version & connection setup to properly infer types in DQL expressions**:

- This PR **affects only expressions**
  - e.g. expression `AVG(e.cost)` is more precise now
  - e.g. direct field fetch `e.cost` behaves as before
- **Supported drivers** are: `mysqli, pdo_mysqli, pgsql, pdo_pgsql, sqlite3, pdo_sqlite`
  - Others are inferred as mixed
- Proper autodetection often needs **working database connection**
  - If connection fails, we return `mixed` or union with stringified type _(depending where it failed and if driver is known or not)_
  - If you have `bleedingEdge` enabled, connection failure is propagated and breaks your PHPStan execution
  - You dont need working connection if you are not using PDO and your Driver is not wrapped by middleware

EDIT: real connection no more needed since #586

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultTypeWalker.php
+++ b/src/Type/Doctrine/Query/QueryResultTypeWalker.php
@@ -13,18 +13,22 @@ use Doctrine\ORM\Query\AST\TypedExpression;
 use Doctrine\ORM\Query\Parser;
 use Doctrine\ORM\Query\ParserResult;
 use Doctrine\ORM\Query\SqlWalker;
+use PDO;
+use PDOException;
+use PHPStan\Doctrine\Driver\DriverDetector;
+use PHPStan\Php\PhpVersion;
 use PHPStan\ShouldNotHappenException;
+use PHPStan\TrinaryLogic;
 use PHPStan\Type\Accessory\AccessoryNumericStringType;
 use PHPStan\Type\BooleanType;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantFloatType;
 use PHPStan\Type\Constant\ConstantIntegerType;
-use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\ConstantTypeHelper;
 use PHPStan\Type\Doctrine\DescriptorNotRegisteredException;
 use PHPStan\Type\Doctrine\DescriptorRegistry;
+use PHPStan\Type\Doctrine\Descriptors\DoctrineTypeDriverAwareDescriptor;
 use PHPStan\Type\FloatType;
-use PHPStan\Type\GeneralizePrecision;
 use PHPStan\Type\IntegerRangeType;
 use PHPStan\Type\IntegerType;
 use PHPStan\Type\IntersectionType;
@@ -36,22 +40,26 @@ use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypeTraverser;
-use PHPStan\Type\TypeUtils;
 use PHPStan\Type\UnionType;
+use Throwable;
 use function array_key_exists;
 use function array_map;
+use function array_values;
 use function assert;
 use function class_exists;
 use function count;
-use function floatval;
 use function get_class;
 use function gettype;
-use function intval;
+use function in_array;
+use function is_int;
 use function is_numeric;
 use function is_object;
 use function is_string;
+use function method_exists;
 use function serialize;
 use function sprintf;
+use function stripos;
+use function strpos;
 use function strtolower;
 use function strtoupper;
 use function unserialize;
@@ -69,6 +77,10 @@ class QueryResultTypeWalker extends SqlWalker
 	private const HINT_TYPE_MAPPING = self::class . '::HINT_TYPE_MAPPING';
 
 	private const HINT_DESCRIPTOR_REGISTRY = self::class . '::HINT_DESCRIPTOR_REGISTRY';
+
+	private const HINT_PHP_VERSION = self::class . '::HINT_PHP_VERSION';
+
+	private const HINT_DRIVER_DETECTOR = self::class . '::HINT_DRIVER_DETECTOR';
 
 	/**
 	 * Counter for generating unique scalar result.
@@ -89,6 +101,12 @@ class QueryResultTypeWalker extends SqlWalker
 
 	/** @var EntityManagerInterface */
 	private $em;
+
+	/** @var PhpVersion */
+	private $phpVersion;
+
+	/** @var DriverDetector::*|null */
+	private $driverType;
 
 	/**
 	 * Map of all components/classes that appear in the DQL query.
@@ -112,15 +130,26 @@ class QueryResultTypeWalker extends SqlWalker
 	/** @var bool */
 	private $hasGroupByClause;
 
+	/** @var bool */
+	private $failOnInvalidConnection;
+
 	/**
 	 * @param Query<mixed> $query
 	 */
-	public static function walk(Query $query, QueryResultTypeBuilder $typeBuilder, DescriptorRegistry $descriptorRegistry): void
+	public static function walk(
+		Query $query,
+		QueryResultTypeBuilder $typeBuilder,
+		DescriptorRegistry $descriptorRegistry,
+		PhpVersion $phpVersion,
+		DriverDetector $driverDetector
+	): void
 	{
 		$query->setHint(Query::HINT_CUSTOM_OUTPUT_WALKER, self::class);
 		$query->setHint(Query::HINT_CUSTOM_TREE_WALKERS, [QueryAggregateFunctionDetectorTreeWalker::class]);
 		$query->setHint(self::HINT_TYPE_MAPPING, $typeBuilder);
 		$query->setHint(self::HINT_DESCRIPTOR_REGISTRY, $descriptorRegistry);
+		$query->setHint(self::HINT_PHP_VERSION, $phpVersion);
+		$query->setHint(self::HINT_DRIVER_DETECTOR, $driverDetector);
 
 		$parser = new Parser($query);
 		$parser->parse();
@@ -140,7 +169,6 @@ class QueryResultTypeWalker extends SqlWalker
 		$this->queryComponents = $queryComponents;
 		$this->nullableQueryComponents = [];
 		$this->hasAggregateFunction = $query->hasHint(QueryAggregateFunctionDetectorTreeWalker::HINT_HAS_AGGREGATE_FUNCTION);
-
 		$this->hasGroupByClause = false;
 
 		// The object is instantiated by Doctrine\ORM\Query\Parser, so receiving
@@ -172,6 +200,32 @@ class QueryResultTypeWalker extends SqlWalker
 		}
 
 		$this->descriptorRegistry = $descriptorRegistry;
+
+		$phpVersion = $this->query->getHint(self::HINT_PHP_VERSION);
+
+		if (!$phpVersion instanceof PhpVersion) { // @phpstan-ignore-line ignore bc promise
+			throw new ShouldNotHappenException(sprintf(
+				'Expected the query hint %s to contain a %s, but got a %s',
+				self::HINT_PHP_VERSION,
+				PhpVersion::class,
+				is_object($phpVersion) ? get_class($phpVersion) : gettype($phpVersion)
+			));
+		}
+
+		$this->phpVersion = $phpVersion;
+
+		$driverDetector = $this->query->getHint(self::HINT_DRIVER_DETECTOR);
+
+		if (!$driverDetector instanceof DriverDetector) {
+			throw new ShouldNotHappenException(sprintf(
+				'Expected the query hint %s to contain a %s, but got a %s',
+				self::HINT_DRIVER_DETECTOR,
+				DriverDetector::class,
+				is_object($driverDetector) ? get_class($driverDetector) : gettype($driverDetector)
+			));
+		}
+		$this->driverType = $driverDetector->detect($this->em->getConnection());
+		$this->failOnInvalidConnection = $driverDetector->failsOnInvalidConnection();
 
 		parent::__construct($query, $parserResult, $queryComponents);
 	}
@@ -228,6 +282,8 @@ class QueryResultTypeWalker extends SqlWalker
 		$dqlAlias = $pathExpr->identificationVariable;
 		$qComp = $this->queryComponents[$dqlAlias];
 		assert(array_key_exists('metadata', $qComp));
+
+		/** @var ClassMetadata<object> $class */
 		$class = $qComp['metadata'];
 
 		assert($fieldName !== null);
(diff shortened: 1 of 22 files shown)


src/Type/Doctrine/Query/QueryResultTypeWalker.php
```diff
@@ -1108,7 +1130,21 @@ public function walkLiteral($literal)
...
 				} else {
-					$type = new ConstantFloatType((float) $value);
+
+					if ($this->em->getConnection()->getDatabasePlatform() instanceof AbstractMySQLPlatform) {
+
+						// both PDO_mysql and mysqli hydrates 123.4 literals as string no matter the configuration (e.g. PDO::ATTR_STRINGIFY_FETCHES being false) and PHP version
+						// the only way to force float is to use 123.4e0 scientific notation
+						// https://dev.mysql.com/doc/refman/8.0/en/number-literals.html
+
+						if (strpos((string) $value, 'e') !== false || strpos((string) $value, 'E') !== false) {
```
JanTvrdik: stripos

src/Type/Doctrine/Query/QueryResultTypeWalker.php
```diff
@@ -1108,7 +1130,21 @@ public function walkLiteral($literal)
 				if (floatval(intval($value)) === floatval($value)) {
 					$type = new ConstantIntegerType((int) $value);
 				} else {
-					$type = new ConstantFloatType((float) $value);
+
+					if ($this->em->getConnection()->getDatabasePlatform() instanceof AbstractMySQLPlatform) {
+
+						// both PDO_mysql and mysqli hydrates 123.4 literals as string no matter the configuration (e.g. PDO::ATTR_STRINGIFY_FETCHES being false) and PHP version
+						// the only way to force float is to use 123.4e0 scientific notation
+						// https://dev.mysql.com/doc/refman/8.0/en/number-literals.html
```
JanTvrdik: Maybe this should mention that it's because 123.4 is a decimal literal, while 123.4e0 is a float literal.

janedbal: Added

ondrejmirtes:
> To match setup of modern PHP versions (8.1.25+) using most used driver (PDO), I added option not to stringify such expressions.

What about auto-detecting this from PhpVersion, and also from objectManagerLoader if one is provided?

src/Type/Doctrine/Query/QueryResultTypeWalker.php
```diff
@@ -812,43 +844,39 @@ public function walkSelectExpression($selectExpression)
 			$resultAlias = $selectExpression->fieldIdentificationVariable ?? $this->scalarResultCounter++;
 			$type = $this->unmarshalType($expr->dispatch($this));
 
-			if (class_exists(TypedExpression::class) && $expr instanceof TypedExpression) {
```
janedbal: This was dead block of code as `class_exists(TypedExpression::class)` is always false.

janedbal: https://github.com/phpstan/phpstan-src/pull/2838

src/Type/Doctrine/Query/QueryResultTypeWalker.php
```diff
@@ -1097,8 +1127,12 @@ public function walkLiteral($literal)
 				break;
 
 			case AST\Literal::BOOLEAN:
-				$value = strtolower($literal->value) === 'true' ? 1 : 0;
-				$type = new ConstantIntegerType($value);
```
janedbal: This was just wrong for postgres, that returns real booleans, but we were infering `1|0|'1'|'0'`

src/Type/Doctrine/Query/QueryResultTypeWalker.php
```diff
@@ -1423,4 +1469,132 @@ private function hasAggregateFunction(AST\SelectStatement $AST): bool
...
+	 * - pdo_mysql: https://github.com/php/php-src/commit/c18b1aea289e8ed6edb3f6e6a135018976a034c6
+	 * - pdo_sqlite: https://github.com/php/php-src/commit/438b025a28cda2935613af412fc13702883dd3a2
+	 * - pdo_pgsql: https://github.com/php/php-src/commit/737195c3ae6ac53b9501cfc39cc80fd462909c82
+	 *
+	 * @param IntegerType|FloatType|BooleanType $type
+	 */
+	private function shouldStringifyExpressions(Type $type): TrinaryLogic
+	{
+		$driver = $this->em->getConnection()->getDriver();
+		$nativeConnection = $this->getNativeConnection();
```
janedbal: Sadly, this causes connection initialization.

janedbal:
Note: ~~this is still WIP and I still plan to finalize this, but~~ implementing all the built-in SQL functions is non-trivial when considering all edge-cases. To be able to implement it properly, I created test repository to verify all differences: https://github.com/janedbal/php-database-drivers-fetch-test/

VincentLanglet:
Hi @janedbal, I need this PR to move on mine https://github.com/phpstan/phpstan-doctrine/pull/520.

What is missing here (maybe writing tests ?) ; do you need help ?

ondrejmirtes:
@VincentLanglet Jan already drowned dozens of hours to get this right, it's a very complex topic as you can see in here https://github.com/janedbal/php-database-drivers-fetch-test/. Don't rush this :)

VincentLanglet:
> @VincentLanglet Jan already drowned dozens of hours to get this right, it's a very complex topic as you can see in here [janedbal/php-database-drivers-fetch-test](https://github.com/janedbal/php-database-drivers-fetch-test/). Don't rush this :)

Sorry if my message was heard like this, wasn't my intention. I thought one of the issue with this big PR was to write tests for all the build-in method, and that maybe I could find and offer time to this.

Anyway, it will take the time it need to take ! :)

janedbal:
The green CI here is misleading, there is still ton of work awaiting (`walkFunction` is wrong, `DoctrineTypeDescriptor::getDatabaseInternalType` has no access to driver so it cannot return proper type, tests need to be MUCH better). I just didnt push my local WIP (maybe I should) :)

janedbal:
@VincentLanglet Pushed the WIP so that you can check the direction I'm leaning towards.

src/Type/Doctrine/Descriptors/DoctrineTypeDescriptor.php
```diff
@@ -13,10 +14,29 @@ interface DoctrineTypeDescriptor
...
+	 *
+	 * See: https://github.com/janedbal/php-database-drivers-fetch-test
+	 *
+	 *              mysql   sqlite  pdo_pgsql    pgsql
+	 * - decimal:  string    float     string   string
+	 * - float:    float     float     string    float
+	 * - bigint:   int         int        int      int
+	 * - bool:     int         int       bool     bool
+	 */
+	public function getDatabaseInternalType(Driver $driver): Type;
```
VincentLanglet: This change will be a BC break since the interface has `@api` annotation.

Not sure which strategy should be followed to avoid this...
Should another interface be introduced ?

janedbal: Yeah, another interface somehow.

VincentLanglet:
> @VincentLanglet Pushed the WIP so that you can check the direction I'm leaning towards.

Thanks, can I help you on this ?
If I understand correctly the goal is to uncomment the tests and make them pass or is there something else ?
Do you need more test to be written ?

I notice there was conflict with the 1.3.x branch, do you have time to solve them or do you want me to try ? :)

janedbal:
@VincentLanglet I've some stuff in backlog to do first before returning to this. But the outline what needs to be done:

- (1) Decide how to deal with [untested drivers](https://github.com/janedbal/php-database-drivers-fetch-test). For example how BOOL column is fetched. Do we want to:
  - Produce best guess? 
    - Currently only postgre uses `bool`, any other driver is just `1|0` (or `'1'|'0'` when stringified). So what do we produce for "other driver"? _Sidenote: I think bool is currently completely broken for postgre._
  - Return mixed?
  - Return benevolent-union of some reasonable value? What is reasonable value, those other drivers ever used?
- (2) Carefully finalize the implementation (mainly all the aggregate functions, which are tricky)
  - There are plenty of edge-cases to decide. 
    - For example `SQRT(negative)` leads either to NULL or a failure depending on a driver. But I assume we dont want to add NULL to the result just because of this. Or do we?
- (3) Test every supported driver
  - This requires completely new set of tests and new CI runs as it need to start **real database** to test our PHPStan types against (just like [here](https://github.com/phpstan/phpstan-doctrine/blob/4040cf0/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php#L227), but for much bigger matrix of PHP versions, drivers and their configs)
  - Basically, we need very similar CI run as used in https://github.com/janedbal/php-database-drivers-fetch-test/blob/master/test-all-php-versions.sh
- (4) Solve the [BC break](https://github.com/phpstan/phpstan-doctrine/pull/506#discussion_r1455970228) introduced

And maybe more stuff I already forgot. I think you can easily help with (3) as such test can be added in standalone MR and possibly merged beforehand. The rest is just digging in code.

VincentLanglet:
> @VincentLanglet I've some stuff in backlog to do first before returning to this. But the outline what needs to be done:
> 
> * (1) Decide how to deal with [untested drivers](https://github.com/janedbal/php-database-drivers-fetch-test). For example how BOOL column is fetched. Do we want to:
>   
>   * Produce best guess?
>     
>     * Currently only postgre uses `bool`, any other driver is just `1|0` (or `'1'|'0'` when stringified). So what do we produce for "other driver"? _Sidenote: I think bool is currently completely broken for postgre._
>   * Return mixed?
>   * Return benevolent-union of some reasonable value? What is reasonable value, those other drivers ever used?

I would take the little step by little step approach: untested drivers will returns the same type than before.
This way, when this PR is merged:
- tested drivers users will have a better type
- untested drivers won't have anything changed

And if a untested-drivers user complains about the type, he/we just have test this drivers and update the code to support it.

> * (2) Carefully finalize the implementation (mainly all the aggregate functions, which are tricky)
>   
>   * There are plenty of edge-cases to decide.
>     
>     * For example `SQRT(negative)` leads either to NULL or a failure depending on a driver. But I assume we dont want to add NULL to the result just because of this. Or do we?

What about starting with only the existing implemented aggregate functions ?
https://github.com/phpstan/phpstan-doctrine/blob/5ca0f53c60494fa3f2f4792e385b8233bda474fd/src/Type/Doctrine/Query/QueryResultTypeWalker.php#L931

It could be easier to first, make a PR with those aggregate functions and then adding other aggregate one by one ?

> * (4) Solve the [BC break](https://github.com/phpstan/phpstan-doctrine/pull/506#discussion_r1455970228) introduced

The options I see are
- Changing the signature to
```
public function getDatabaseInternalType(/** Driver $driver */): Type;
```
and passing the driver. In next major the param will be required. This is often done by Symfony.

- Creating an interface with a different method
```
interface DriverAwareDoctrineTypeDescriptor {
     public function getDatabaseInternalTypeFromDriver(Driver $driver): Type;
}
```
and implementing the interface ; then the check 
```
$typeDescriptor instanceof DriverAwareDoctrineTypeDescriptor
    ?  $typeDescriptor->getDatabaseInternalTypeFromDriver($driver);
    :  $typeDescriptor->getDatabaseInternalType();
```
could be made.

I think the decision is more on @ondrejmirtes about how he wants to proceed in order to evolve such method signature.

VincentLanglet:
Awesome, thanks a lot for this @janedbal

ruudk:
Thanks @janedbal for your continued work on making this smarter and smarter 💙

janedbal:
After discussion with @ondrejmirtes we decided to separate some stuff to other PRs:

- https://github.com/phpstan/phpstan-doctrine/pull/577
- https://github.com/phpstan/phpstan-doctrine/pull/578
- https://github.com/phpstan/phpstan-doctrine/pull/579
- https://github.com/phpstan/phpstan-doctrine/pull/580
- https://github.com/phpstan/phpstan-doctrine/pull/581
- https://github.com/phpstan/phpstan-doctrine/pull/582
- https://github.com/phpstan/phpstan-doctrine/pull/583
- https://github.com/phpstan/phpstan-doctrine/pull/584

 To minimize conflicts upon rebase, I squashed this MR. Here is a link to old commits if I ever need to check the real history: https://github.com/phpstan/phpstan-doctrine/commits/8c333fe

tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
```diff
@@ -75,6 +74,10 @@ public static function getAdditionalConfigFiles(): array
 
 	public static function setUpBeforeClass(): void
 	{
+		if (PHP_VERSION_ID < 80100) {
+			self::markTestSkipped('Test uses asserts without stringification, those work only with PHP >= 8.1. Stringification is tested in platform test.');
```
ondrejmirtes: So this entire test is skipped below PHP 8.1? I don't understand the reason and I'd prefer if the test wasn't skipped.

janedbal:
@ruudk, @arnaud-lb, @VincentLanglet Would you mind testing this in your codebases? It would be very helpful to have this verified by more real-world codebases. Thank you!

```json
{
    "require-dev": {
        "phpstan/phpstan-doctrine": "dev-option-not-to-stringify-numeric-types"
    },
    "repositories": [
        {
            "type": "git",
            "url": "git@github.com:janedbal/phpstan-doctrine.git"
        }
    ]
}
```

tests/Platform/QueryResultTypeWalkerFetchTypeMatrixTest.php
```diff
@@ -62,362 +102,4161 @@ public static function getAdditionalConfigFiles(): array
...
+		$pdoPgsqlExpectedResult,
+		$pgsqlExpectedResult,
+		$mssqlExpectedResult,
+		string $stringify
+	): void
+	{
+		$dataset = (string) $this->dataName();
+		$phpVersion = PHP_VERSION_ID;
+
+		$this->performDriverTest('pdo_mysql', self::CONFIG_DEFAULT, $data, $dqlTemplate, $dataset, $phpVersion, $mysqlExpectedType, $mysqlExpectedResult, $stringify);
```
ondrejmirtes: Each test case method should have a single `performDriverTest()` call.

Either create a new test case method for different parameters (each one will have `provideCases` data provider), or keep a single test case method, but create a carthesian product with the current `provideCases` method but add different values for drivers etc.

ondrejmirtes (changes requested):

tests/Platform/QueryResultTypeWalkerFetchTypeMatrixTest.php
janedbal: Done.

ruudk:
I tested it out on our large application (182 repositories, 16k files) and it works. Unfortunately, it could only remove 3 ignored errors 😅. I guess we don't use that many expressions, or we used Assert.

arnaud-lb:
I won't be able to test on a meaningful code base, but maybe @Kmecnin can?

VincentLanglet:
> I tested it out on our large application (182 repositories, 16k files) and it works. Unfortunately, it could only remove 3 ignored errors 😅. I guess we don't use that many expressions, or we used Assert.

It may remove more ignored errors with https://github.com/phpstan/phpstan-doctrine/pull/520.

I also tried the branch @janedbal and 
- getting 0 new errors
- removed 0 ignored errors

It's mainly because we already typed our code with things like `int|string` or `numeric`.

But this PR will help us to change the `numeric` type by the right type from the database. :)

tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
janedbal: Adjusted the asserts, now it run everywhere.

ondrejmirtes (approved):

ondrejmirtes:
Thank you so much! Let's see how well this fares in real-world projects :)

janedbal:
> It's mainly because we already typed our code with things like int|string or numeric.

Yeah, because PHPStan generally allows type widening. 


In our case, we have:
- Custom rule for direct return of Doctrine type where we deny widening (https://gist.github.com/janedbal/08ec3d5f45dad8627d2cf6499b082512)
  - e.g. `return $query->getResult()`
- Custom rule that enforces inline vardoc when Doctrine type is assigned to a variable (vardoc allows only narrowing by default)
  - e.g. `$result = $query->getResult()`

Those two rules ensure that I fix the types at least near the source and I can verify that the changes I made here are valid all over our codebase.

Issue #501, opened 2023-11-28
Issue with `QueryBuilderGetQueryDynamicReturnTypeExtension` when a variable typed as an interface is used


Kocal:
Hi everyone, 

When upgrading from 1.3.50 to 1.3.53, we started to have the following issue:
```
 -- ----------------------------------------------------------------------------------------------------------------------------
     Error
 -- ----------------------------------------------------------------------------------------------------------------------------
     Internal error: Internal error: Class 'App\Common\Entity\Behavior\StatsOwnerInterface' does not exist while analysing file
     /Users/halliaume/workspace/international-website/src/Core/Handler/Analytics/StatsHandler.php
     Run PHPStan with -v option and post the stack trace to:
     https://github.com/phpstan/phpstan/issues/new?template=Bug_report.yaml
     Child process error (exit code 1):
 -- ----------------------------------------------------------------------------------------------------------------------------
```

`App\Common\Entity\Behavior\StatsOwnerInterface` is indeed not a class, but it exists as an interface.

When running PHPStan with `--debug`: 
```
Uncaught Doctrine\Persistence\Mapping\MappingException: Class 'App\Common\Entity\Behavior\StatsOwnerInterface' does not exist in /PATH/TO/PROJECT/vendor/doctrine/persistence/src/Persistence/Mapping/MappingException.php:80
#0 /PATH/TO/PROJECT/vendor/doctrine/persistence/src/Persistence/Mapping/RuntimeReflectionService.php(39): Doctrine\Persistence\Mapping\MappingException::nonExistingClass('App\\Common\\Enti...')
#1 /PATH/TO/PROJECT/vendor/doctrine/persistence/src/Persistence/Mapping/AbstractClassMetadataFactory.php(283): Doctrine\Persistence\Mapping\RuntimeReflectionService->getParentClasses('App\\Common\\Enti...')
#2 /PATH/TO/PROJECT/vendor/doctrine/persistence/src/Persistence/Mapping/AbstractClassMetadataFactory.php(318): Doctrine\Persistence\Mapping\AbstractClassMetadataFactory->getParentClasses('App\\Common\\Enti...')
#3 /PATH/TO/PROJECT/vendor/doctrine/persistence/src/Persistence/Mapping/AbstractClassMetadataFactory.php(207): Doctrine\Persistence\Mapping\AbstractClassMetadataFactory->loadMetadata('App\\Common\\Enti...')
[...]

PR #507, opened 2023-12-14
Add test for #501

Kocal:
Hi,

I wanted to provide a failing test case for #501, but in fact it's not failing at all 😕
Maybe the issue is coming from our side... 

However, it can be useful to merge this test as it covers another case when an phpdoc-typed interface can be passed to `->from()`.


janedbal:
I think this can target 1.3.x

ondrejmirtes:
Thank you.

janedbal:
The problem in this MR is that you used `get_class($this)` (passes) instead of direct `Vehicle::class` (fails).

Anyway, the behaviour is fixed in https://github.com/phpstan/phpstan-doctrine/pull/515

PR #510, opened 2024-01-05
Allow overriding DescriptorRegistry service

ruudk:
## What has been done

I want to be able to override the default `PHPStan\Type\Doctrine\DescriptorRegistry` service. But that's not possible.

The service does not have a name, and I cannot override it on my own configuration because:
> Service name must be a alpha-numeric string and not a number, 'PHPStan\Type\Doctrine\DescriptorRegistryFactory' given.

The solution is to rename `DescriptorRegistry` to `DefaultDescriptorRegistry` and introduce a new `DescriptorRegistry` interface. That one is used everywhere. 

Together with this, we name introduce a `doctrineTypeDescriptorRegistry` service name, that is injected where it's needed.

## Why do I want this?

In our codebase, we rely on the new [TypedFieldMapper](https://www.doctrine-project.org/projects/doctrine-orm/en/2.17/reference/typedfieldmapper.html) functionality. 

We have a setup that dynamically creates these types from our identity objects. They all point to a single `GenericStringType`. The reason this works, is that Doctrine's [TypeRegistry](https://github.com/doctrine/dbal/blob/6a793fb72948c9dec844de57de8cdbc16ff75bec/src/Types/TypeRegistry.php#L27-L29) creates instances per registered. That means that types can point to separate instances of the same type. 

It allows us to do things like this:
```php
class Entity {
    #[Column(name: 'locale')]
    private Locale $locale;
}
```

`Locale` maps to an instance of `GenericStringType`. That type is able to construct the `Locale` object like this:
```php
final class GenericStringType extends StringType
{
    public function convertToDatabaseValue($value, AbstractPlatform $platform) : ?string
    {
        if ($value === null) {
            return null;
        }

        Assert::isInstanceOf($value, Stringable::class);

        return (string) $value;
    }

    public function convertToPHPValue($value, AbstractPlatform $platform) : ?object
    {
        $className = self::getTypeRegistry()->lookupName($this);

        if ($value === null) {
            return null;
        }
        
        return $className::fromString($value);
    }
}
```

This works great, except that PHPStan doesn't understand this. It operates under the impression that there can only be 1 type per FQCN.

By overriding the `DescriptorRegistry` I can do something like this:
```php
class CustomDescriptorRegistry implements DescriptorRegistry
{
    /**
     * @var array<class-string<Type>, DoctrineTypeDescriptor>
     */
    private $descriptors = [];

    /**
     * @param DoctrineTypeDescriptor[] $descriptors
     */
    public function __construct(
        array $descriptors,
    ) {
        foreach ($descriptors as $descriptor) {
            $this->descriptors[$descriptor->getType()] = $descriptor;
        }
    }

    public function get(string $type) : DoctrineTypeDescriptor
    {
        $typesMap = Type::getTypesMap();
        if (isset($typesMap[$type]) === false) {
            throw new DescriptorNotRegisteredException();
        }

        /** @var class-string<Type> $typeClass */
        $typeClass = $typesMap[$type];

        return match ($typeClass) {
            GenericStringType::class => new GenericStringTypeDescriptor($type),
            GenericIntegerType::class => new GenericIntegerTypeDescriptor($type),
            default => $this->descriptors[$typeClass] ?? throw new DescriptorNotRegisteredException(),
        };
    }
}
```

I pass `$type` to the constructor of these Generic type objects.

```php
final class GenericStringTypeDescriptor implements DoctrineTypeDescriptor
{
    public function __construct(
        private string $type,
    ) {}

    public function getType() : string
    {
        return GenericStringType::class;
    }

    public function getWritableToPropertyType() : Type
    {
        return new ObjectType($this->type);
    }

    public function getWritableToDatabaseType() : Type
    {
        return new ObjectType($this->type);
    }

    public function getDatabaseInternalType() : Type
    {
        return new StringType();
    }
}
```

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -74,7 +74,8 @@ conditionalTags:
 services:
 	-
 		class: PHPStan\Type\Doctrine\DescriptorRegistryFactory
-	-
+
+	doctrineTypeDescriptorRegistry:
 		class: PHPStan\Type\Doctrine\DefaultDescriptorRegistry
 		factory: @PHPStan\Type\Doctrine\DescriptorRegistryFactory::createRegistry
 
@@ -115,6 +116,7 @@ services:
 		arguments:
 			queryBuilderClass: %doctrine.queryBuilderClass%
 			argumentsProcessor: @doctrineQueryBuilderArgumentsProcessor
+			descriptorRegistry: @doctrineTypeDescriptorRegistry
 		tags:
 			- phpstan.broker.dynamicMethodReturnTypeExtension
 	-
@@ -138,6 +140,7 @@ services:
 		class: PHPStan\Type\Doctrine\CreateQueryDynamicReturnTypeExtension
 		arguments:
 			objectMetadataResolver: @PHPStan\Type\Doctrine\ObjectMetadataResolver
+			descriptorRegistry: @doctrineTypeDescriptorRegistry
 		tags:
 			- phpstan.broker.dynamicMethodReturnTypeExtension
 	-
--- a/extension.neon
+++ b/extension.neon
@@ -75,7 +75,7 @@ services:
 	-
 		class: PHPStan\Type\Doctrine\DescriptorRegistryFactory
 	-
-		class: PHPStan\Type\Doctrine\DescriptorRegistry
+		class: PHPStan\Type\Doctrine\DefaultDescriptorRegistry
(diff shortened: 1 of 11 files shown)


ruudk:
Should this target 1.4 or 1.3?

ondrejmirtes:
Thank you! README update would also be nice :)

PR #513, opened 2024-01-09
Resolve methodcall args in QueryBuilderType with proper Scope

### LINKED ISSUE #512: self::class in entity is interpreted in wrong context
With this entity:

```
#[Entity]
class Person {
    #[Id]
    #[Column(type: "integer")]
    public int $id;

    public static function getQb(EntityManager $em): QueryBuilder {
        return $em->createQueryBuilder()
                  ->select('p')
                  ->from(self::class, 'p');
    }
}
```

And this code:

```
class Test {
    public static function test(EntityManager $em): void {
        Person::getQb($em)->getQuery()->getResult();
    }
}
```

I get the following error:

```
 -- --------------------------------------------------------------------------------------------------------------------
     Error
 -- --------------------------------------------------------------------------------------------------------------------
     Internal error: Internal error: Class "App\Test" is not a valid entity or mapped super class. while analysing file
     C:\Projekte\TestProjects\PhpstanDoctrine\src\Test.php
     Run PHPStan with -v option and post the stack trace to:
     https://github.com/phpstan/phpstan/issues/new?template=Bug_report.yaml
     Child process error (exit code 1):
 -- --------------------------------------------------------------------------------------------------------------------
```

If I replace `self::class` with `Person::class`, it works as expected. It seems that `self` is being understood to be `Test`, when in fact it refers to `Person`. I can provide a fuller reproduction example if desired. This seems to have gone broken in a recent release.

janedbal:
Resolves #512

### DIFF
--- a/src/Type/Doctrine/QueryBuilder/CreateQueryBuilderDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/QueryBuilder/CreateQueryBuilderDynamicReturnTypeExtension.php
@@ -48,7 +48,8 @@ class CreateQueryBuilderDynamicReturnTypeExtension implements DynamicMethodRetur
 		}
 
 		return new $class(
-			$this->queryBuilderClass ?? 'Doctrine\ORM\QueryBuilder'
+			$this->queryBuilderClass ?? 'Doctrine\ORM\QueryBuilder',
+			$scope
 		);
 	}
 
--- a/src/Type/Doctrine/QueryBuilder/QueryBuilderGetQueryDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/QueryBuilder/QueryBuilderGetQueryDynamicReturnTypeExtension.php
@@ -147,7 +147,7 @@ class QueryBuilderGetQueryDynamicReturnTypeExtension implements DynamicMethodRet
 
 				if ($lowerMethodName === 'set') {
 					try {
-						$args = $this->argumentsProcessor->processArgs($scope, $methodName, array_slice($calledMethodCall->getArgs(), 0, 1));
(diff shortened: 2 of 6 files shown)


src/Type/Doctrine/QueryBuilder/CreateQueryBuilderDynamicReturnTypeExtension.php
```diff
@@ -48,7 +48,8 @@ public function getTypeFromMethodCall(
 		}
 
 		return new $class(
-			$this->queryBuilderClass ?? 'Doctrine\ORM\QueryBuilder'
+			$this->queryBuilderClass ?? 'Doctrine\ORM\QueryBuilder',
+			$scope
```
ondrejmirtes, line 52: To pass Scope in a type is weird. Perhaps we should do `argumentsProcessor->processArgs` sooner and carry the final args in the Type? Or something different like that.

ondrejmirtes (changes requested):

src/Type/Doctrine/QueryBuilder/CreateQueryBuilderDynamicReturnTypeExtension.php
janedbal: That was my original thought, but this felt easier. Anyway, do you see any real issue with this approach?

janedbal: Btw to fix the failing tests, I probably need to store Scope per the method call inside.

ondrejmirtes: I feel like this is still going to be wrong in some cases, as the failing tests prove. One Scope per QueryBuilder is not enough as the scope changes when the QB gets built. Better process the args sooner.

PR #514, opened 2024-01-12
Support class-string<T> in $qb->from

janedbal:


janedbal:
Just tested with our big codebase where this approach is used quite often and it works fine. 

--------

Some limitations still remain, we met those two:
```php
    /**
     * @param class-string<T> $class
     * @return list<T>
     *
     * @template T of SomeEntity
     */
    private function getFooBar(
        string $class,
    ): array
    {
        return $this->entityManager->createQueryBuilder()
            ->select('a')
            ->from($class, 'a');
            ->getQuery()
            ->getResult();

          // Method Clazz::getFooBar() should return list<T of SomeEntity> but returns list<SomeEntity>.                                                                                    
          // 💡 Type SomeEntity is not always the same as T. It breaks the contract for some argument types, typically subtypes.
    }
```

And this is (obviously) not supported:

```php
    /**
     * @param class-string<MyEntity&MyInterface> $class
     */
    private function getFooBar(
        string $class,
    ): array
    {
        return $this->entityManager->createQueryBuilder()
            ->select('a')
            ->from($class, 'a');
            ->getQuery()
            ->getResult();
    }
```

ondrejmirtes:
Thank you!

ondrejmirtes:
I can't release this - it broke 1.4.x (when the class-string leads to an interface, it crashes Doctrine): https://github.com/phpstan/phpstan-doctrine/actions/runs/7520215871

Please submit a fix on top of 1.3.x.

PR #515, opened 2024-01-15
No failure when MappingException occurs

janedbal:
This resolves:

- https://github.com/phpstan/phpstan-doctrine/pull/514#issuecomment-1890994136
- https://github.com/phpstan/phpstan-doctrine/issues/501


ondrejmirtes:
Thank you.

Issue #517, opened 2024-01-16
Performance issues after 1.3.54


derrabus:
Upgrading this library from 1.3.53 to 1.3.54 and beyond caused serious performance issues for us.

When running PHPStan with 1.3.x-dev of the plugin on my laptop, PHPStan's progress bar is stuck at 98% for 5 minutes, but the process completes eventually. With 1.3.53, PHPStan completes in 20 seconds on an empty result cache. On our CI, the PHPStan job timed out regularly which hasn't been a problem before.

With git bisect, I could identify a9d0aafb749e94b93dac3cd84a7f1419714f1578 (#500, @janedbal) as the casue of this issue. If I revert this very commit on the 1.3.x branch, everything's back to normal. Our workaround for now is to pin phpstan-doctrine to the 1.3.53 release.

Unfortunately, I cannot publish the codebase I'm experiencing this issue on. I'm also a bit lost about how I could isolate a reproducer here, but I'm willing to answer any questions that might help someone investigating the problem.

ondrejmirtes:
Hi, this sounds like it gets stuck on a single file where the complexity explodes. If you run PHPStan with `-vvv --debug` you'll find that file. If you then simplify the file and other files this one calls (files with method declarations this file calls), you'll find the code pattern the extension gets stuck on.

ondrejmirtes:
Hey, I just reverted https://github.com/phpstan/phpstan-doctrine/commit/a9d0aafb749e94b93dac3cd84a7f1419714f1578 and going to release that. I'd still appreciate if you were to point down the performance hog in 1.3.54.

I also experienced slowdown in real-world project but not as serious.

janedbal:
Idea: I can imagine similar behaviour, if the codebase would contain some really huge chain of methodcalls which return QueryBuilder. The differece of behaviour caused by this MR is best described in #499.

The way it works is that it deeply jumps to every method that returns QueryBuilder. And if that one contains method call that returns QueryBuilder, it jumps even there. So some huge chain of those might slow down the analysis significantly.

janedbal:
Instead of reverting, I think it should be possible to limit the depth of this chain @ondrejmirtes 

With revert, nobody ever finds out where the problem resides and this significant improvement will never be present in this plugin imo

ondrejmirtes:
I'm usually for causing a bit of pain to force people to report problems so that the software can get better for everyone later.

[...]

PR #518, opened 2024-01-16
Infer QueryBuilderType for any method returning QueryBuilder: revive with cache

janedbal:
- Attempt to resolve https://github.com/phpstan/phpstan-doctrine/issues/517 by visiting each method only once.

### DIFF
--- a/src/Type/Doctrine/QueryBuilder/OtherMethodQueryBuilderParser.php
+++ b/src/Type/Doctrine/QueryBuilder/OtherMethodQueryBuilderParser.php
@@ -35,6 +35,9 @@ class OtherMethodQueryBuilderParser
 	/** @var Container */
 	private $container;
 
+	/** @var array<string, array<string, list<QueryBuilderType>>> */
+	private $cache = [];
+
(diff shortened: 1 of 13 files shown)


src/Type/Doctrine/QueryBuilder/OtherMethodQueryBuilderParser.php
```diff
@@ -51,18 +54,23 @@ public function findQueryBuilderTypesInCalledMethod(Scope $scope, MethodReflecti
...
 		}
 
+		$methodName = $methodReflection->getName();
 		$fileName = $methodReflection->getDeclaringClass()->getFileName();
 		if ($fileName === null) {
 			return [];
 		}
 
+		if (isset($this->cache[$fileName][$methodName])) {
+			return $this->cache[$fileName][$methodName];
```
janedbal, line 64: Note: if this gets proven to help, we should improve the cache key to support multiple classes with the same method name in the same file.

ondrejmirtes:
Thank you.

PR #538, opened 2024-02-08
Remove inheritdoc phpdoc

VincentLanglet:


ondrejmirtes (changes requested):
There shouldn't be any errors when analysing that file with PHPStan, not with ORM 2 nor 3.

VincentLanglet:
> There shouldn't be any errors when analysing that file with PHPStan, not with ORM 2 nor 3.

I pushed a new commit.

Current state:
- No error with PHPStan with ORM 2
- No introduced error with PHPStan on ORM 3. I still have some errors, but less than the original PR and all the error I have, you also have them in https://github.com/phpstan/phpstan-doctrine/actions/runs/7830700494/job/21365380864?pr=535

ondrejmirtes:
Awesome, thank you!

PR #535, opened 2024-02-08
Compatibility with Doctrine ORM 3.0 and DBAL 4.0

ondrejmirtes:

PR #539, opened 2024-02-08
BigInt support for DBAL 4

### LINKED ISSUE #484: Bigint are casted to int (if possible) for DBAL 4
With https://github.com/doctrine/dbal/pull/6177, the bigint will now be casted to `int` by doctrine.
So I assume the BigIntType will need to be updated for Doctrine 4.

I assume 
https://github.com/phpstan/phpstan-doctrine/blob/39d78ad7ca5386acf71f727b9ae5ae5979a5e1ea/src/Type/Doctrine/Descriptors/BigIntType.php#L21
need to be changed at least from numeric-string to numeric "if DBAL version is 4".

Or should it be `BenevolentUnion(int, numeric)` because most of users use them for small enough values (?).

VincentLanglet:
Trying to Closes #484

This will require https://github.com/phpstan/phpstan-doctrine/pull/535

I'm currently not sure how we will handle the PHP32 bit situation to say that if it's not unsigned, we're sure it's integer.
Currently the unsigned data is not present in the descriptor.

I see multiple solutions with drawbacks @ondrejmirtes, I tried one.

Since the DoctrineTypeDescriptor is `@api` I'm not sure if we should create a new interface or if a better solution exists.

Open to any suggestion.

### DIFF
--- a/src/Rules/Doctrine/ORM/EntityColumnRule.php
+++ b/src/Rules/Doctrine/ORM/EntityColumnRule.php
@@ -11,6 +11,7 @@ use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\Doctrine\DescriptorNotRegisteredException;
 use PHPStan\Type\Doctrine\DescriptorRegistry;
+use PHPStan\Type\Doctrine\Descriptors\OptionRelatedDoctrineTypeDescriptor;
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
I decided to disregard the 32-bit possibility, because it's 2024 and it's only about DBAL 4 which just came out: https://github.com/phpstan/phpstan-doctrine/pull/535/commits/719a8f40cb747dad1c37b8dc1ce4c60368c791a1

PR #541, opened 2024-02-09
Add a rule to check that arrays are hinted to doctrine

BackEndTea:
I created this rule for our own code base because we forgot to properly hint to arrays when refactoring code to use doctrine. I figured it could be useful to others, and this seemed like the most logical repository for it.

If this is something you want in here i can expand the tests and improve support for the `Statement` class for example.

### DIFF
--- a/src/Rules/Doctrine/DBAL/ArrayParameterTypeRule.php
+++ b/src/Rules/Doctrine/DBAL/ArrayParameterTypeRule.php
@@ -1,0 +1,149 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Doctrine\DBAL;
+
+use Doctrine\DBAL\Connection;
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\Type\ConstantScalarType;
+use PHPStan\Type\ObjectType;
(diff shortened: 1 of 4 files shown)


ondrejmirtes:
Can you please tell me more? What kind of bug happens if you don't hint the parameter type correctly?

VincentLanglet:
> Can you please tell me more? What kind of bug happens if you don't hint the parameter type correctly?

When looking at the executeStatement method there is a line
```
if ($this->needsArrayParameterConversion($params, $types)) {
    [$sql, $params, $types] = $this->expandArrayParameters($sql, $params, $types);
}
```
with
```
private function needsArrayParameterConversion(array $params, array $types): bool
    {
        if (is_string(key($params))) {
            return true;
        }

        foreach ($types as $type) {
            if (
                $type === ArrayParameterType::INTEGER
                || $type === ArrayParameterType::STRING
                || $type === ArrayParameterType::ASCII
                || $type === ArrayParameterType::BINARY
            ) {
                return true;
            }
        }

        return false;
    }
```

So I assume, if you don't typehint the parameter as array, the value is not expanded which cause a wrong query.

ondrejmirtes:
Sorry, I didn't want details from Doctrine core (I don't understand it). I wanted to see some userland code and the implications of hinting/not hinting the types.

BackEndTea:
For us, if we don't add those hints, we get a warning of `Array to string conversion` since it tries to turn the array of parameters into a string.

PR #542, opened 2024-02-10
fix AnnotationDriver check

janatjak:
Fixes problem with `doctrine/annotations:*` & `doctrine/orm:^3`  as dependencies at same time

### DIFF
--- a/src/Doctrine/Mapping/ClassMetadataFactory.php
+++ b/src/Doctrine/Mapping/ClassMetadataFactory.php
@@ -29,7 +29,7 @@ class ClassMetadataFactory extends \Doctrine\ORM\Mapping\ClassMetadataFactory
 	protected function initialize(): void
 	{
 		$drivers = [];
-		if (class_exists(AnnotationReader::class)) {
+		if (class_exists(AnnotationDriver::class) && class_exists(AnnotationReader::class)) {
 			$docParser = new DocParser();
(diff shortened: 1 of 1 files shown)


src/Doctrine/Mapping/ClassMetadataFactory.php
```diff
@@ -29,7 +29,7 @@ public function __construct(string $tmpDir)
 	protected function initialize(): void
 	{
 		$drivers = [];
-		if (class_exists(AnnotationReader::class)) {
+		if (class_exists(AnnotationDriver::class)) {
```
ondrejmirtes: Should probably ask for both?

ondrejmirtes (changes requested):

src/Doctrine/Mapping/ClassMetadataFactory.php
janatjak: `doctrine/annotations` is peer dependency for `doctrine/orm` since orm version `2.10.0`

but if you want to use AnnotationDriver, it's "necessary"

on the other hand, why not add a double check 👍 

give me sec

ondrejmirtes:
Please rebase this on top of 1.3.x so I can merge and release it immediately.

janatjak:
done

ondrejmirtes:
Thank you!

PR #543, opened 2024-02-11
Fix aggregate function on litteral

### LINKED ISSUE #460: SUM() aggregation returns wrong type
I have the following code:

```php
$qb = $this->em->createQueryBuilder()
               ->addSelect('tvs.title')
               ->addSelect('tvs.broadcastDate')
               ->addSelect('COUNT(tva.id) AS count')
               ->addSelect('SUM(CASE WHEN tva.freePeriod = TRUE THEN 1 ELSE 0 END) AS freePeriodCount')
               ->addSelect('SUM(CASE WHEN tva.subscriptions IS NOT NULL THEN 0 ELSE 1 END) AS userStreamingCount')
               ->from(TvActivity::class, 'tva')
               ->join('tva.tvShow', 'tvs')
               ->addGroupBy('tvs.id');
$dbRows = $qb->getQuery()->getResult();
\PHPStan\dumpType($dbRows);
```

The dumped type is:

```
list<array{
    title: string,
    broadcastDate: DateTimeImmutable,
    count: int<0, max>|numeric-string,
    freePeriodCount: 0|1|'0'|'1'|null,
    userStreamingCount: 0|1|'0'|'1'|null,
}>
```

The last two fields can be arbitrary numbers, but are reported as 0 and 1 only. If it helps I can create a full reproduction repo, but hopefully things are clear like this. It's also wrong when doing `SUM(1)`, so the exact entity definitions might not matter.

Using phpstan-doctrine 1.3.40.

VincentLanglet:
Closes #460

Since you introduced this in https://github.com/phpstan/phpstan-doctrine/pull/232, I'll be happy to have your review @arnaud-lb 

When doing a SUM, I think the type must be "ungeneralize" (cf #460 )

When doing a AVG,
- if the value is not constant, I think the type must be "ungeneralize" too
- also, I got a failure without `$type = TypeCombinator::union($type, $type->toFloat(), $type->toFloat()->toString());` because the AVG value computed in my tests was `1.0` or `'1.0'` (even for `AVG(1)) and not `1` or `'1'` so I think we need to add the float result (and the stringified float result).

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultTypeWalker.php
+++ b/src/Type/Doctrine/Query/QueryResultTypeWalker.php
@@ -944,11 +944,8 @@ class QueryResultTypeWalker extends SqlWalker
 					$this->walkSimpleArithmeticExpression($aggExpression->pathExpression)
 				);
 
-				if (count($type->getConstantScalarValues()) !== 1) {
-					$type = TypeUtils::generalizeType($type, GeneralizePrecision::lessSpecific());
-				}
-
-				$type = TypeCombinator::union($type, $type->toFloat(), $type->toFloat()->toString());
+				$type = TypeCombinator::union($type, $type->toFloat());
+				$type = TypeUtils::generalizeType($type, GeneralizePrecision::lessSpecific());
 
 				return $this->marshalType(TypeCombinator::addNull($type));
 
--- a/src/Type/Doctrine/Query/QueryResultTypeWalker.php
+++ b/src/Type/Doctrine/Query/QueryResultTypeWalker.php
@@ -933,11 +933,31 @@ class QueryResultTypeWalker extends SqlWalker
 		switch ($aggExpression->functionName) {
 			case 'MAX':
 			case 'MIN':
+				$type = $this->unmarshalType(
+					$this->walkSimpleArithmeticExpression($aggExpression->pathExpression)
+				);
+
+				return $this->marshalType(TypeCombinator::addNull($type));
+
 			case 'AVG':
+				$type = $this->unmarshalType(
+					$this->walkSimpleArithmeticExpression($aggExpression->pathExpression)
+				);
+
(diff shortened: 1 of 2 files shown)


VincentLanglet:
Seems like 
```
$type->toFloat()->toString()
```
doesn't do what I was looking for @ondrejmirtes, I may require your help to end this PR.

If I have a ConstantInteger(1), I can get a ConstantFloat(1.0) with the `toFloat` method.
What would be the recommended way to get a ConstantString('1.0') from this float now ?
The toString method is giving me a ConstantString('1')...

janedbal:
I'm using [generalize for those cases](https://github.com/phpstan/phpstan-doctrine/pull/506/files#diff-7fd356ee7494ac7e9610c9496e854e994cbcef6c4c56d533fc5fed88e4568a47R454) (expand QueryResultTypeWalker)

VincentLanglet:
> I'm using [generalize for those cases](https://github.com/phpstan/phpstan-doctrine/pull/506/files#diff-7fd356ee7494ac7e9610c9496e854e994cbcef6c4c56d533fc5fed88e4568a47R454) (expand QueryResultTypeWalker)

I used `generalize` for the SUM method.

But if we're doing `AVG(1)` if thought this was better to return `1|'1'|...` rather than `float|int|numeric-string|null` but after a good night, I think it's maybe a useless edge case to handle.

janedbal:
For AVG, it is edgecase, but the same issue for SUM is not.

arnaud-lb (approved):
LGTM!

ondrejmirtes:
Thank you.

VincentLanglet:
Your welcome

Since 1.3.x is not the base branch it didn't closed automatically the issue https://github.com/phpstan/phpstan-doctrine/issues/460 @ondrejmirtes

PR #546, opened 2024-02-13
Enforce literal string for connection execute methods

VincentLanglet:
In the same way that we enforce passing literal string to QueryBuilder methods in https://github.com/phpstan/phpstan-doctrine/blob/1.4.x/stubs/bleedingEdge/ORM/QueryBuilder.stub

I wonder if we shouldn't enforce literal string to for `Connection::execute*` methods which executes SQL directly.

This would help chaning query like
```
->executeQuery('SELECT foo where foo.bar = '.$nonLiteralValue);
```
to
```
->executeQuery('SELECT foo where foo.bar = :value', ['value' => $nonLiteralValue]);
```

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -19,7 +19,12 @@ parameters:
 			- Doctrine\ORM\Tools\Pagination\Paginator
 	stubFiles:
 		- stubs/Criteria.stub
+		- stubs/DBAL/Cache/CacheException.stub
+		- stubs/DBAL/Cache/QueryCacheProfile.stub
 		- stubs/DBAL/Exception/UniqueConstraintViolationException.stub
+		- stubs/DBAL/Types/Type.stub
+		- stubs/DBAL/Exception.stub
+		- stubs/DBAL/Result.stub
 		- stubs/DocumentManager.stub
 		- stubs/DocumentRepository.stub
 		- stubs/EntityManager.stub
--- a/src/Stubs/Doctrine/StubFilesExtensionLoader.php
+++ b/src/Stubs/Doctrine/StubFilesExtensionLoader.php
@@ -35,6 +35,7 @@ class StubFilesExtensionLoader implements StubFilesExtension
 		}
 
 		$files = [
+			$path . '/DBAL/Connection.stub',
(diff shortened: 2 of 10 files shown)


VincentLanglet:
Hi, WDYT about it @ondrejmirtes ?

ondrejmirtes:
I'm waiting for Craig Francis to react to this.

VincentLanglet:
Friendly ping @craigfrancis if you find some time

craigfrancis:
Hi @VincentLanglet, sorry for the delay.

I've been meaning to look at these, as you're right, the `$sql` param really needs be a `literal-string`.

But, partially due to laziness, partially down to lack of sleep, can you explain why there are several stub files? As in, `Connection.stub` makes sense, but why are the others there (I suspect there is a simple/obvious explanation, but my brain isn't working at the moment).

(note to self: [source](https://github.com/doctrine/dbal/blob/4.0.x/src/Connection.php); might also consider `prepare`... and at a later date the 13 functions that use `string $query, array $params` like `fetchAssociative`; but also the `$sql` in `convertExceptionDuringQuery` and `expandArrayParameters`).

ondrejmirtes:
The reason for that is probably the last orange box on this page https://phpstan.org/user-guide/stub-files

craigfrancis:
Thanks Ondřej, yep that makes perfect sense :-)

VincentLanglet:
> The reason for that is probably the last orange box on this page [phpstan.org/user-guide/stub-files](https://phpstan.org/user-guide/stub-files)

Yes it is. As soon as I introduced some class in the phpdoc, I needed the stub too.

craigfrancis:
Great, in which case it looks good to me.

At a later date I'll check and do a PR for `prepare()`, the 13 functions that use `string $query, array $params` (e.g. `fetchAssociative`); and maybe `convertExceptionDuringQuery` and `expandArrayParameters` ([source](https://github.com/doctrine/dbal/blob/4.0.x/src/Connection.php)).

ondrejmirtes:
Thank you.

PR #547, opened 2024-02-13
Use older gedmo version for testing doctrine 3

VincentLanglet:
Gedmo/doctrine-extensions added a conflict with orm 3 and dbal 4 in the latest version 3.15.
So we currently need to use a lower version like 3.14 when testing ORM 3 @ondrejmirtes.

The error can be seen on the PR https://github.com/phpstan/phpstan-doctrine/pull/546

Not sure if it's the best way to fix it, but it does the job.


ondrejmirtes:
Thank you.

Issue #545, opened 2024-02-13
Enforce `Connection::executeQuery` on SELECT and `Connection::executeStatement` on others queries


VincentLanglet:
When using two database, one with READ/WRITE rights and one replica with READ writes, Doctrine use `executeQuery/executeStatement` to know which database to use.
In this situation using `executeQuery` for UPDATE/DELETE/etc give an error about missing rights.

A thing which doesn't help to detect an error is the fact `executeQuery` works properly for UPDATE/DELETE/etc statement when you use only one database with READ/WRITE permission.

It would be great to report a Phpstan error if executeQuery is used on a non-select query and if executeStatement is used on a select query.

I'll try to write one for constant string, but I'm not sure it will be possible for query like `'SELECT'.$sql` since the type will be simplified to `string`.

stof:
I have a valid use case for `executeQuery` for `UPDATE` statements: `UPDATE ... RETURNING`, which has a result set. But then, such case requires either being in a transaction (which would have already enforced connecting to the primary database) or ensuring it manually.

So such rule should either be opt-in or be smart enough to detect such patterns.

VincentLanglet:
> I have a valid use case for `executeQuery` for `UPDATE` statements: `UPDATE ... RETURNING`, which has a result set. But then, such case requires either being in a transaction (which would have already enforced connecting to the primary database) or ensuring it manually.

I'm not sure if this should be considered as valid given the comment
https://github.com/doctrine/dbal/blob/c70bae9e69e0e543044d5e282a624d9b3cbea6fb/src/Connections/PrimaryReadReplicaConnection.php#L43-L44
but I suppose it's the only way to have returning statements...

I added a commit to avoid such false positive then https://github.com/phpstan/phpstan-doctrine/pull/548/commits/184002d010d49a1d211cb3c1833dbd694c479403

stof:
@VincentLanglet read the paragraph just before the one you linked (and also the other parts of this phpdoc)

VincentLanglet:
> @VincentLanglet read the paragraph just before the one you linked (and also the other parts of this phpdoc)

I might missing some point in the phpdoc, but I understood that
```
$conn->ensureConnectedToPrimary();
$conn->execQuery('Update ...');
```
is valid.

But unless you're using 
- A select query
- A returning query

it seems safer to use `executeStatement` no ?
It avoids extra-work like transaction/calling ensureConnectedToPrimary, so it avoid mistake.

[...]

PR #548, opened 2024-02-13
Enforce Select queries on Connection::executeQuery

VincentLanglet:
Related to https://github.com/phpstan/phpstan-doctrine/issues/545

If you're okay adding something like this here, I can do the same rule for executeStatement @ondrejmirtes.

### DIFF
--- a/src/Rules/Doctrine/ORM/ExecuteQueryRule.php
+++ b/src/Rules/Doctrine/ORM/ExecuteQueryRule.php
@@ -1,0 +1,75 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Doctrine\ORM;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\Type\ObjectType;
+use function count;
+use function sprintf;
+use function str_starts_with;
+use function strtolower;
+use function trim;
+
+/**
+ * @implements Rule<Node\Expr\MethodCall>
+ */
+class ExecuteQueryRule implements Rule
+{
+
+	public function getNodeType(): string
+	{
+		return Node\Expr\MethodCall::class;
+	}
+
+	public function processNode(Node $node, Scope $scope): array
+	{
+		if (!$node->name instanceof Node\Identifier) {
+			return [];
+		}
+
+		if (count($node->getArgs()) === 0) {
+			return [];
+		}
+
+		$methodName = $node->name->toLowerString();
+		if (
+			$methodName !== 'executequery'
+			&& $methodName !== 'executecachequery'
+		) {
+			return [];
+		}
+
+		$calledOnType = $scope->getType($node->var);
+		$connection = 'Doctrine\DBAL\Connection';
+		if (!(new ObjectType($connection))->isSuperTypeOf($calledOnType)->yes()) {
+			return [];
+		}
+
+		$queries = $scope->getType($node->getArgs()[0]->value)->getConstantStrings();
+		if (count($queries) === 0) {
+			return [];
+		}
+
+		foreach ($queries as $query) {
+			if (!$this->isSelectQuery($query->getValue())) {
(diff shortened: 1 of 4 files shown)


VincentLanglet:
Do you agree with adding such rule to `phpstan-doctrine` repository @ondrejmirtes or should I implement this in my own project ? The issue is described in https://github.com/phpstan/phpstan-doctrine/issues/545#issue-2132391594

VincentLanglet:
I rebased the branch.

Do you agree with adding such rule to `phpstan-doctrine` repository @ondrejmirtes or should I implement this in my own project ? The issue is described in https://github.com/phpstan/phpstan-doctrine/issues/545#issue-2132391594

src/Rules/Doctrine/ORM/ExecuteQueryRule.php
```diff
@@ -0,0 +1,85 @@
...
+		$sql = strtolower(trim($sql));
+
+		// `executeQuery` can be used for returning statements like
+		// `UPDATE ... RETURNING`, so we ignore such case to avoid false positive.
+		// @see https://github.com/phpstan/phpstan-doctrine/issues/545#issuecomment-2015059629
+		if (str_contains($sql, 'returning')) {
+			return true;
+		}
+
+		return str_starts_with($sql, 'select');
```
stof, line 82: this is not enough. With CTEs, a read query might not _start_ with the `SELECT` keyword.

src/Rules/Doctrine/ORM/ExecuteQueryRule.php
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	private function isSelectQuery(string $sql): bool
+	{
+		$sql = strtolower(trim($sql));
+
+		// `executeQuery` can be used for returning statements like
+		// `UPDATE ... RETURNING`, so we ignore such case to avoid false positive.
+		// @see https://github.com/phpstan/phpstan-doctrine/issues/545#issuecomment-2015059629
+		if (str_contains($sql, 'returning')) {
```
stof, line 78: because of this case, this method is more about detecting whether the query has a result set or no (queries with a result set are the one that require using `executeQuery` to be able to consume the result set).

src/Rules/Doctrine/ORM/ExecuteQueryRule.php
stof: Another case is `EXPLAIN ...`

VincentLanglet: Should I return 
```
str_contains($sql, 'returning') || str_contains($sql, 'select');
```
?

I feel like it will miss multiple statement, but I don't see another way to look for select queries easily.

src/Rules/Doctrine/ORM/ExecuteQueryRule.php
VincentLanglet: My first purpose was to help developers to detect when `executeQuery` is wrongly used and it will crash on a PrimaryReplicaConnection because it require access to primary database.

In our codebase we had a `executeQuery('Update foo');` which was working fine until we switched to multiple database with primary and replica and crash in production. Such rule would have helped to avoid this.

ondrejmirtes:
I think this rule isn't generally applicable. As you said, these patterns are fine with a single DB connection.

VincentLanglet:
> I think this rule isn't generally applicable. As you said, these patterns are fine with a single DB connection.

Ok, I'll keep it for my project then.
Thanks for your answer

janedbal:
@VincentLanglet We would use such rule, is it open-sourced somewhere else or should I copy it?

VincentLanglet:
> @VincentLanglet We would use such rule, is it open-sourced somewhere else or should I copy it?

You're free to copy it (and expose it in a package if you want) ; my project is for work so it's private...

janedbal:
> and expose it in a package if you want

Thanks! We have several generally usable Doctrine related rules, but I never found enough time+priority to open-source it. So maybe one day it may become part of that.

PR #549, opened 2024-02-17
WIP - Introduce tests on platforms

VincentLanglet:
Related to https://github.com/phpstan/phpstan-doctrine/pull/506#issuecomment-1944268880 @janedbal,
I try to introduce tests on different platform. Does this solution would work for you ?

Edit: Are you also ok with this solution @ondrejmirtes ?
I had in mind to provide `@group platform` on every platform tests, each with a special group like `@group sql`.
And adding all different platform in github actions.

### DIFF
--- a/.github/workflows/test-platforms.yml
+++ b/.github/workflows/test-platforms.yml
@@ -1,0 +1,58 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
+
+name: "Test on platforms"
+
+on:
+  pull_request:
+  push:
+    branches:
+      - "1.3.x"
+
+jobs:
+  test-mysql:
+    name: "Tests mysql"
+    runs-on: "ubuntu-latest"
+
+    env:
+      DATABASE_URL: mysql://root@127.0.0.1:3306/phpstan_doctrine
+
+    services:
+      mysql:
+        image: mysql:${{ matrix.mysql-version }}
+        options: >-
+          --health-cmd "mysqladmin ping --silent"
+          --name=mysqlcontainer
+          -e MYSQL_ALLOW_EMPTY_PASSWORD=yes
+          -e MYSQL_DATABASE=phpstan_doctrine
+        ports:
+          - 3306:3306
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version:
+          - "8.2"
+        mysql-version:
+          - '5.7'
+          - '8.0'
+
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v3
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "${{ matrix.php-version }}"
+          ini-file: development
+          extensions: "mongodb"
+
+      - name: "Install dependencies"
+        run: "composer update --no-interaction --no-progress"
+
+      - name: Change mysql sql_mode
+        run: "docker exec mysqlcontainer mysql -u root -e \"SET GLOBAL sql_mode = '';\""
+
+      - name: "Tests"
+        run: "vendor/bin/phpunit --group=mysql"
--- a/tests/Type/Doctrine/Query/MysqlQueryResultTypeWalkerTest.php
+++ b/tests/Type/Doctrine/Query/MysqlQueryResultTypeWalkerTest.php
@@ -8,6 +8,12 @@ namespace PHPStan\Type\Doctrine\Query;
  */
 class MysqlQueryResultTypeWalkerTest extends PDOQueryResultTypeWalkerTest
 {
+	public static function getAdditionalConfigFiles(): array
+	{
+		return [
+			__DIR__ . '/../data/QueryResult/config-mysql.neon',
+		];
+	}
 
 	protected static function getEntityManagerPath(): string
 	{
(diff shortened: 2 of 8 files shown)


VincentLanglet:
Related to https://github.com/phpstan/phpstan-doctrine/pull/506#issuecomment-1944268880 @janedbal,
I try to introduce tests on different platform. Does this solution would work for you ?

tests/Type/Doctrine/Query/MysqlQueryResultTypeWalkerTest.php
```diff
@@ -0,0 +1,17 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Doctrine\Query;
+
+/**
+ * @group platform
+ * @group mysql
+ */
+class MysqlQueryResultTypeWalkerTest extends PDOQueryResultTypeWalkerTest
```
VincentLanglet, line 9: Currently I extends PDOQueryResultTypeWalkerTest but later we'll work his own dataprovider

.github/workflows/test-platforms.yml
```diff
@@ -0,0 +1,58 @@
...
+jobs:
+  test-mysql:
+    name: "Tests mysql"
+    runs-on: "ubuntu-latest"
+
+    env:
+      DATABASE_URL: mysql://root@127.0.0.1:3306/phpstan_doctrine
+
+    services:
+      mysql:
```
janedbal, line 20: There will be more of those right?

.github/workflows/test-platforms.yml
```diff
@@ -0,0 +1,58 @@
...
+          - 3306:3306
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version:
+          - "8.2"
+        mysql-version:
+          - '5.7'
+          - '8.0'
```
janedbal, line 37: I think those does not matter, but since it costs nothing here, why not.

.github/workflows/test-platforms.yml
```diff
@@ -0,0 +1,58 @@
...
+          -e MYSQL_ALLOW_EMPTY_PASSWORD=yes
+          -e MYSQL_DATABASE=phpstan_doctrine
+        ports:
+          - 3306:3306
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version:
+          - "8.2"
```
janedbal, line 34: We definitelly need to check "before 8.1" and "after 8.1" as there were some major changes.

tests/Type/Doctrine/Query/QueryResultTypeWalkerTestCase.php
```diff
@@ -0,0 +1,349 @@
...
+		];
+
+		$id = 1;
+
+		foreach (self::combinations($dataOne) as $combination) {
+			[$intColumn, $stringColumn, $stringNullColumn] = $combination;
+			$one = new One();
+			$one->id = (string) $id++;
+			$one->intColumn = $intColumn;
+			$one->stringColumn = $stringColumn;
```
janedbal, line 108: We definitelly need entities suited for platform testing (having float, decimal, bool, int, bigint columns).

tests/Type/Doctrine/Query/QueryResultTypeWalkerTestCase.php
```diff
@@ -0,0 +1,349 @@
...
+				$many->simpleArrayColumn = ['foo'];
+				$many->one = $one;
+				$one->manies->add($many);
+				$em->persist($many);
+			}
+
+			$em->persist($one);
+		}
+
+		foreach (self::combinations($dataJoinedInheritance) as $combination) {
```
janedbal, line 140: I'm not sure we need this complexity. I ended with super simple test fetching a single expression and verifying that type in https://github.com/janedbal/php-database-drivers-fetch-test All we need here is to check that against the inferred PHPStan type.

And I feel this is the best approach to accomplish what we need as it can be [quite readable in the test itself](https://github.com/janedbal/php-database-drivers-fetch-test/blob/master/tests/PhpDatabaseDriverTest.php#L127)

janedbal (changes requested):

.github/workflows/test-platforms.yml
VincentLanglet: Yes the purpose is to add one run by platform

tests/Type/Doctrine/Query/QueryResultTypeWalkerTestCase.php
VincentLanglet: I just copied the existing `QueryResultTypeWalkerTest` with the whole dataprovider.

Do you think we should split it ?
- Only single expression for platforms tests
- single expression and complex expression for PDO tests in order to validate PHPStan for these

I thought this was better to keep the same data for every platform.

janedbal:
The matrix (at some level) should also contain:
- driver (there are more for mysql and pgsql and some of them [behave differently](https://github.com/janedbal/php-database-drivers-fetch-test/blob/7a53490/tests/PhpDatabaseDriverTest.php#L77))
- config (mainly stringification on/off for PDO drivers)

VincentLanglet:
> The matrix (at some level) should also contain:
> 
> * config (mainly stringification on/off for PDO drivers)

Since the config must be passed to the entity manager in a `entity-manager.php` passed to phpstan in config with
```
parameters:
	doctrine:
		objectManagerLoader: entity-manager.php
```
I don't see how it would be in the matrix.

Currently, the only solution I see is to create one entityManager for every platform/config and one test case for every entityManager...

PR #552, opened 2024-03-09
Fix branching querybuilder type inference

VincentLanglet:
Hi,

I got an issue with a conditional query like
```
$qb = $em->createQueryBuilder();
if ($bool) {
    $qb->select('m.intColumn');
} else {
    $qb->select('m.intColumn', 'm.stringNullColumn');
}
$result = $qb->from(Many::class, 'm')->getQuery()->getResult();
```
which was considering to always returning `array<array{intColumn: int}>` which is wrong.

After debugging, I thought the issue was coming from `TypeCombinator::union` cf https://github.com/phpstan/phpstan/issues/10694
After more debugging, I find out the behavior was different for covariant template or invariant template
https://phpstan.org/r/79c3c099-5d1d-4e36-9cf3-d6028d68415d

So I think the template for Query should not be covariant.

The covariant was added by you @ondrejmirtes in https://github.com/phpstan/phpstan-doctrine/commit/ec49b7ab2efc2c1443746b42a9440c0f1d21db2e ; do you remember why ?

Thanks

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
 	],
 	"require": {
 		"php": "^7.2 || ^8.0",
-		"phpstan/phpstan": "^1.10.48"
+		"phpstan/phpstan": "^1.10.63"
 	},
 	"conflict": {
 		"doctrine/collections": "<1.0",
--- a/src/Rules/Doctrine/ORM/QueryBuilderDqlRule.php
+++ b/src/Rules/Doctrine/ORM/QueryBuilderDqlRule.php
@@ -15,6 +15,7 @@ use PHPStan\Type\Doctrine\ObjectMetadataResolver;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\TypeUtils;
 use Throwable;
+use function array_values;
 use function count;
 use function sprintf;
 use function strpos;
@@ -117,7 +118,8 @@ class QueryBuilderDqlRule implements Rule
 					$message .= sprintf("\nDQL: %s", $dql->getValue());
 				}
 
-				$messages[] = RuleErrorBuilder::message($message)
+				// Use message as index to prevent duplicate
+				$messages[$message] = RuleErrorBuilder::message($message)
 					->identifier('doctrine.dql')
 					->build();
 			} catch (AssertionError $e) {
@@ -125,7 +127,7 @@ class QueryBuilderDqlRule implements Rule
 			}
 		}
 
-		return $messages;
+		return array_values($messages);
 	}
 
 }
--- a/src/Type/Doctrine/Query/QueryType.php
+++ b/src/Type/Doctrine/Query/QueryType.php
@@ -32,6 +32,11 @@ class QueryType extends GenericObjectType
 		return parent::equals($type);
 	}
 
+	public function getTypeWithoutSubtractedType(): Type
+	{
+		return $this;
+	}
+
 	public function isSuperTypeOf(Type $type): TrinaryLogic
 	{
 		if ($type instanceof self) {
--- a/src/Type/Doctrine/Query/QueryType.php
+++ b/src/Type/Doctrine/Query/QueryType.php
@@ -11,15 +11,25 @@ use PHPStan\Type\Type;
 class QueryType extends GenericObjectType
 {
 
+	/** @var Type */
+	private $indexType;
+
+	/** @var Type */
+	private $resultType;
+
 	/** @var string */
 	private $dql;
 
-	public function __construct(string $dql, ?Type $indexType = null, ?Type $resultType = null)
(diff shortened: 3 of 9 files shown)


ondrejmirtes:
No, the problem is in https://github.com/phpstan/phpstan-doctrine/blob/1.4.x/src/Type/Doctrine/QueryBuilder/QueryBuilderType.php - it should override the `isSuperTypeOf` method and implement it correctly.

VincentLanglet:
> No, the problem is in [`1.4.x`/src/Type/Doctrine/QueryBuilder/QueryBuilderType.php](https://github.com/phpstan/phpstan-doctrine/blob/1.4.x/src/Type/Doctrine/QueryBuilder/QueryBuilderType.php?rgh-link-date=2024-03-11T08%3A14%3A25Z) - it should override the `isSuperTypeOf` method and implement it correctly.

I'm not sure to understand why it's QueryBuilderType. I tried to debug more and it's not called.

But after more debugging, I find out that during the execution of `TypeCombinator::union(...$resultTypes);`,
that this call 
https://github.com/phpstan/phpstan-src/blob/952b313e94151d5e942e4f4577d42dd2c60d76b9/src/Type/TypeCombinator.php#L426
is changing the type from `QueryType` to `GenericObjectType` and then the issue occurs because of the call `GenericObjectType::isSuperTypeOf` when `QueryType::isSuperTypeOf` works fine.

As shown in my latest commit the `testConditionalAddSelect` pass with the override of `changeSubtractedType`.

This seems correct to me but some others tests are failing https://github.com/phpstan/phpstan-doctrine/actions/runs/8233591779/job/22513364391?pr=552:
- `QueryBuilderDqlRuleSlowTest::testRuleBranches`: one error was commented (https://github.com/phpstan/phpstan-doctrine/blob/9c4935fe27184dd0e6ae7848a1c063ca51c1414d/tests/Rules/Doctrine/ORM/QueryBuilderDqlRuleSlowTest.php#L112-L115), seems like now I get it back so I think it's for the good. But I also get one line 107 which seems ok te me.
- `QueryBuilderDqlRuleTest::testRuleBranches`: same
- `QueryBuilderDqlRuleTest::testBranchingPerformance`: got two times the same message instead of one. I assume a array_unique will be necessary in the Rule.
- `ReturnQueryBuilderExpressionTypeResolverExtensionTest::testFileAsserts` got the "weird" type `Doctrine\ORM\Query<null, QueryResult\Entities\Many>|Doctrine\ORM\Query<null, QueryResult\Entities\Many>`. I think it's because the two query have different DQL but same results. Maybe `QueryType::equals`/`QueryType::isSuperTypeOf` need to be changed now.

WDYT @ondrejmirtes, do you agree with this analyse/new direction ?

VincentLanglet:
The only failing test is the following one:
```php
$branchingQuery = $this->getBranchingQueryBuilder($em)->getQuery();
assertType('Doctrine\ORM\Query<null, QueryResult\Entities\Many>', $branchingQuery);
```

I get
```php
'Doctrine\ORM\Query<null, QueryResult\Entities\Many>|Doctrine\ORM\Query<null, QueryResult\Entities\Many>'
```
now.

This is kinda true because, if I call `getDQL()` on `$branchingQuery` we're getting an union type
```php
$branchingQueryDQL = $this->getBranchingQueryBuilder($em)->getQuery()->getDQL();
assertType('\'SELECT m FROM QueryResult\\Entities\\Many m\'|\'SELECT m FROM QueryResult\\Entities\\Many m WHERE m.intColumn = 1\'', $branchingQueryDQL);
```

Should we:
- Update the test to assertType `'Doctrine\ORM\Query<null, QueryResult\Entities\Many>|Doctrine\ORM\Query<null, QueryResult\Entities\Many>'` ?
- do a PR on phpstan-src to simplify display of type when possible (the internal PHPStan type could be `'Doctrine\ORM\Query<null, QueryResult\Entities\Many>|Doctrine\ORM\Query<null, QueryResult\Entities\Many>'` but the display one `'Doctrine\ORM\Query<null, QueryResult\Entities\Many>'`).
- something else ?

VincentLanglet:
Now https://github.com/phpstan/phpstan-src/pull/2973 is merged, I just need to wait for the next release

ondrejmirtes:
Released now https://github.com/phpstan/phpstan/releases/tag/1.10.63, please update it here

VincentLanglet:
> Released now [phpstan/phpstan@`1.10.63` (release)](https://github.com/phpstan/phpstan/releases/tag/1.10.63), please update it here

Done.
I still had to update this line https://github.com/phpstan/phpstan-doctrine/pull/552/files#diff-60cf06ba47535191938faa998f36489178d29d99d8b2cbf39fac343871bec885R29

ondrejmirtes:
Thank you.

PR #560, opened 2024-04-17
Synchronize `Collection::add` with doctrine/collections

VincentLanglet:
Hi,

`Collection::add` returns `void` and not `true` according to doctrine/collections
https://github.com/doctrine/collections/blob/9b9c38a3eff3c7477989129b096d21e2b1710ee0/src/Collection.php#L34-L42

### DIFF
--- a/src/Stubs/Doctrine/StubFilesExtensionLoader.php
+++ b/src/Stubs/Doctrine/StubFilesExtensionLoader.php
@@ -3,9 +3,11 @@
 namespace PHPStan\Stubs\Doctrine;
 
 use Composer\InstalledVersions;
+use OutOfBoundsException;
 use PHPStan\BetterReflection\Reflector\Exception\IdentifierNotFound;
 use PHPStan\BetterReflection\Reflector\Reflector;
 use PHPStan\PhpDoc\StubFilesExtension;
+use function class_exists;
 use function dirname;
 use function strpos;
 
@@ -61,9 +63,13 @@ class StubFilesExtensionLoader implements StubFilesExtension
 			$files[] = $stubsDir . '/ServiceEntityRepository.stub';
 		}
 
-		$collectionVersion = class_exists(InstalledVersions::class)
-			? InstalledVersions::getVersion('doctrine/collections')
-			: null;
+		try {
+			$collectionVersion = class_exists(InstalledVersions::class)
+				? InstalledVersions::getVersion('doctrine/collections')
+				: null;
+		} catch (OutOfBoundsException $e) {
+			$collectionVersion = null;
+		}
 		if ($collectionVersion !== null && strpos($collectionVersion, '1.') === 0) {
 			$files[] = $stubsDir . '/Collections/Collection1.stub';
 		} else {
--- a/src/Stubs/Doctrine/StubFilesExtensionLoader.php
+++ b/src/Stubs/Doctrine/StubFilesExtensionLoader.php
@@ -2,10 +2,12 @@
 
 namespace PHPStan\Stubs\Doctrine;
 
+use Composer\InstalledVersions;
 use PHPStan\BetterReflection\Reflector\Exception\IdentifierNotFound;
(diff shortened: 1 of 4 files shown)


janedbal:
What about older versions? I see [true in 1.8.0](https://github.com/doctrine/collections/blob/1.8.0/lib/Doctrine/Common/Collections/Collection.php#L38)

VincentLanglet:
> What about older versions? I see [true in 1.8.0](https://github.com/doctrine/collections/blob/1.8.0/lib/Doctrine/Common/Collections/Collection.php?rgh-link-date=2024-04-17T12%3A13%3A46Z#L38)

Oh, do we have to duplicate the stub then ?
or drop old version of doctrine/collection ? (https://github.com/doctrine/collections/releases/tag/1.8.0 was released in 2022)

ondrejmirtes:
We need conditional including of different stubs based on the version. In a StubFilesExtension.

VincentLanglet:
> We need conditional including of different stubs based on the version. In a StubFilesExtension.

Done

src/Stubs/Doctrine/StubFilesExtensionLoader.php
```diff
@@ -59,6 +61,13 @@ public function getFiles(): array
 			$files[] = $stubsDir . '/ServiceEntityRepository.stub';
 		}
 
+		$collectionVersion = InstalledVersions::getVersion('doctrine/dbal');
+		if ($collectionVersion !== null && strpos($collectionVersion, '1.') === 0) {
+			$files[] = $stubsDir . '/Collections/Collection1.stub';
```
janedbal, line 74: Isnt this swapped? This condition matches versions 1.x, but the stub uses `public function add($element):void`

janedbal: Ok, you fixed it in https://github.com/phpstan/phpstan-doctrine/pull/560/commits/cb1c2df60f86d690247b9f74a537638cf9ac6330

VincentLanglet: > Ok, you fixed it in [cb1c2df](https://github.com/phpstan/phpstan-doctrine/commit/cb1c2df60f86d690247b9f74a537638cf9ac6330)

Yes, ^^'

src/Stubs/Doctrine/StubFilesExtensionLoader.php
```diff
@@ -59,6 +61,13 @@ public function getFiles(): array
 			$files[] = $stubsDir . '/ServiceEntityRepository.stub';
 		}
 
+		$collectionVersion = InstalledVersions::getVersion('doctrine/dbal');
```
ondrejmirtes: Please check that `InstalledVersions` class exists.

ondrejmirtes (changes requested):

src/Stubs/Doctrine/StubFilesExtensionLoader.php
janedbal: Strictly speaking, we should even catch OutOfBoundsException in case dbal is not installed: https://getcomposer.org/doc/07-runtime.md#knowing-the-version-of-package-x

VincentLanglet: Oh damn, and the check should be about `'doctrine/collections'` not `'doctrine/dbal'`.

I fixed it and added a class_exists check and the try/catch of OutOfBoundsException.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

VincentLanglet:
I messed up and target 1.4.x instead of 1.3.x.

Should I reopen a PR on 1.3.x @ondrejmirtes ?

ondrejmirtes:
No, you'll have to wait for release alongside PHPStan 1.11 😊 shouldn't be too far off

PR #561, opened 2024-04-17
Fix compatibility with dbal 4

### LINKED ISSUE #559: Incompatible bleedingEdge Connection stub typehint with DBAL 4
Hi,

I hope I'm not creating duplicity here, but I've just run into an incompatibility with the bleeding-edge stub of `DBAL\Connection` added in https://github.com/phpstan/phpstan-doctrine/pull/546. This issue arises with DBAL 4, leading to false positives when an array of `ParameterType|ArrayParameterType` enums is passed as the `$types` argument of `Connection::execute*` methods.

E.g
```php
namespace Vendor\App;

use Doctrine\DBAL\Connection;
use Doctrine\DBAL\ParameterType;

final readonly class DbalFooFinder
{
    public function __construct(private Connection $connection)
    {
    }

    /** @return list<array{bar: string, baz: string}> */
    public function find(string $baz): array
    {
        /** @var list<array{bar: string, baz: string}> $result */
        $result = $this->connection
            ->executeQuery(
                'SELECT bar, baz FROM foo WHERE baz = :baz',
                ['baz' => $baz],
                ['baz' => ParameterType::STRING],
            )
            ->fetchAllAssociative();

        return $result;
    }
}
```

Results in

```
 ------ --------------------------------------------------------------------------------
  Line   src/DbalFooFinder.php
 ------ --------------------------------------------------------------------------------
  23     Parameter #3 $types of method Doctrine\DBAL\Connection::executeQuery() expects
         array<int|string, Doctrine\DBAL\Types\Type|int|string|null>, array<string,
         Doctrine\DBAL\ParameterType::STRING> given.
 ------ --------------------------------------------------------------------------------
```

Even though the DBAL 4's typehint for `$types` parameter is
```
/**
 * @psalm-type WrapperParameterType = string|Type|ParameterType|ArrayParameterType
 * @psalm-type WrapperParameterTypeArray = array<int<0, max>, WrapperParameterType>|array<string, WrapperParameterType>
 */

/**
 * @psalm-param WrapperParameterTypeArray $types
 */
```

VincentLanglet:
Should close #559

### DIFF
--- a/src/Stubs/Doctrine/StubFilesExtensionLoader.php
+++ b/src/Stubs/Doctrine/StubFilesExtensionLoader.php
@@ -7,7 +7,10 @@ use OutOfBoundsException;
 use PHPStan\BetterReflection\Reflector\Exception\IdentifierNotFound;
 use PHPStan\BetterReflection\Reflector\Reflector;
 use PHPStan\PhpDoc\StubFilesExtension;
+use function class_exists;
 use function dirname;
+use function file_exists;
+use function strpos;
 
 class StubFilesExtensionLoader implements StubFilesExtension
 {
--- a/src/Stubs/Doctrine/StubFilesExtensionLoader.php
+++ b/src/Stubs/Doctrine/StubFilesExtensionLoader.php
@@ -2,6 +2,8 @@
 
 namespace PHPStan\Stubs\Doctrine;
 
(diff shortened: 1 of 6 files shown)


ondrejmirtes:
Thank you.

PR #562, opened 2024-04-18
Add __benevolent return types to doctrine Result stub

kamil-zacek:


ondrejmirtes:
Thank you!

derrabus:
This PR makes the assumption that values in a result match one of the types `float`, `int`, `string` or `null` wile DBAL itself documents them as `mixed`. I don't know if narrowing values to that union solves an actual problem for anyone. Just note that if we at Doctrine would believe that we could narrow down the types in this manner, we would've done so already.

DBAL's wrapper layer has no control over the types of the values inside a result set. Values are returned as the driver returns them.

* Looking at the out-of-the-box drivers of DBAL, you missed at least `bool` which can be returned by Postgres databases.
* If I used a custom driver or a middleware that is capable of hydrating let's say `DateTime` objects, a result could also contain `DateTime` objects. `DateTime` is of course just an example: a driver could really return any type really.

ondrejmirtes:
Alright, reverted, thank you for the feedback!

PR #563, opened 2024-04-18
Rich query builder stub

kamil-zacek:


ondrejmirtes:
Closing since this will have the same problems as https://github.com/phpstan/phpstan-doctrine/pull/562#issuecomment-2063774640

PR #568, opened 2024-05-21
QueryBuilderDqlRule: mention DQL for BranchingQueryBuilderType

janedbal:
Currently, I spent quite some time debugging where certain DQL error is coming from and what helped me most was to see which DQL branch actually caused the issue. So I believe it might be handy even for others.


ondrejmirtes:
Thank you!

PR #571, opened 2024-05-29
QueryResultTypeWalker: fix TypedExpression handling; COUNT(x) is always int

janedbal:


src/Type/Doctrine/Query/QueryResultTypeWalker.php
```diff
@@ -816,28 +815,8 @@ public function walkSelectExpression($selectExpression): string
...
-						if ($type instanceof IntegerType || $type instanceof FloatType) {
-							return TypeCombinator::union($type->toString(), $type);
-						}
-						if ($type instanceof BooleanType) {
-							return TypeCombinator::union($type->toInteger()->toString(), $type);
-						}
-					}
-					return $enforcedType;
-				});
+			if ($expr instanceof TypedExpression) {
```
janedbal, line 821: This was dead block of code as `class_exists(TypedExpression::class)` is **always false**.

janedbal:
I tried extracting this from https://github.com/phpstan/phpstan-doctrine/pull/506, but it is just wrong. ~~The stringification happens even for TypedExpressions, so closing this.~~

PR #572, opened 2024-05-30
Platform test: compare inferred types with real SQL engine results

janedbal:
- This is a needed test before finalization of #506. 
- The intention here is to have **real comparison of inferred types with real results** for numerous database drivers, PHP versions, configurations and Doctrine versions.
- So far, it only asserts current behaviour and **fixes boolean inferring for Postgres**, but I think it will gain plenty of improvements once #506 will be tested by that.
- In order to develop&test locally the same matrix as in CI, I added docker-compose setup copying the CI setup


Makefile
```diff
@@ -3,7 +3,7 @@ check: lint cs tests phpstan
 
 .PHONY: tests
 tests:
-	php vendor/bin/phpunit
+	php vendor/bin/phpunit --exclude-group=platform
```
ondrejmirtes: I'd think it'd make sense to do it the opposite way - when someone runs `vendor/bin/phpunit` do not run platform tests, only run them when someone specifically types `--group platform`. I do the same thing here: https://github.com/phpstan/phpstan-src/blob/1310ce93bf4ef94b1fdba1b90f22ccf003599623/phpunit.xml#L34-L39

ondrejmirtes (changes requested):
This is really nice :)

Two questions:

1) Why not cover PHP 7.2+?
2) Why not cover PHP 8.2+?

janedbal:
> Why not cover PHP 7.2+? Why not cover PHP 8.2+?

The main point of platform test is to cover different PHP behaviours and that changed only in PHP 8.1.

---------

Take this as a starting point, I have a big list of needed improvements of this test. Using complete range of PHP versions can be added.

Makefile
janedbal: Adjusted.

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #573, opened 2024-06-07
add immediately-invoked-callable into EntityManager stub

mzk:
https://phpstan.org/writing-php-code/phpdocs-basics#callables

### DIFF
--- a/stubs/EntityManagerInterface.stub
+++ b/stubs/EntityManagerInterface.stub
@@ -70,7 +70,7 @@ interface EntityManagerInterface extends ObjectManager
 
 	/**
 	 * @param-immediately-invoked-callable $func
-	 * @param Closure(): T $func
+	 * @param callable(): T $func
 	 * @return T
 	 *
 	 * @template T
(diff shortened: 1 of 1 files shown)


stubs/EntityManagerInterface.stub
```diff
@@ -68,4 +68,13 @@ interface EntityManagerInterface extends ObjectManager
 	 */
 	public function getClassMetadata($className);
 
+	/**
+	 * @param-immediately-invoked-callable $func
+	 * @param Closure(): T $func
+	 * @return T
+	 *
+	 * @template T
```
ondrejmirtes, line 76: You have some build errors here. Every callable should be accepted, but you're just allowing Closure here. Change it to `callable(): T`.

ondrejmirtes (changes requested):

stubs/EntityManagerInterface.stub
ondrejmirtes: Also - this wouldn't be a PHPStan error, I'm just pointing out what I'd prefer here.

The errors are reported because you don't have `use Closure` at the top or `\Closure` in the PHPDoc :)

ondrejmirtes:
Thank you.

PR #574, opened 2024-06-07
Fix EntityManagerInterface wrapInTransaction stub

jbafford:
`wrapInTransaction`'s callable takes `self` as a parameter.

### DIFF
--- a/stubs/EntityManagerInterface.stub
+++ b/stubs/EntityManagerInterface.stub
@@ -70,7 +70,7 @@ interface EntityManagerInterface extends ObjectManager
 
 	/**
 	 * @param-immediately-invoked-callable $func
-	 * @param callable(): T $func
+	 * @param callable(self): T $func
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
More like `$this` but we should test what it does with subclasses.

jbafford:
I went with `self` because that's what the `EntityManagerInterface` actually declares in its `@psalm-param` typehint on that method.

However, I wonder if neither is strictly accurate, since the `EntityManagerDecorator` passes the callable directly through to its wrapped `EntityManagerInterface`, so calling `wrapInTransaction` on a decorator is going to get your function called with the wrapped EM, which from the caller's point of view, is neither `$this` nor `self`. Perhaps loosening the type to `EntityManagerInterface` would be more accurate. Do you want me to make that change?

ondrejmirtes:
No, it's okay, thank you!

PR #577, opened 2024-06-24
Platform test: test more php versions, orm3 and dbal4

janedbal:
Separated from https://github.com/phpstan/phpstan-doctrine/pull/506


ondrejmirtes:
Thank you!

PR #578, opened 2024-06-24
Introduce DoctrineTypeDriverAwareDescriptor & DriverDetector

janedbal:
Separated from https://github.com/phpstan/phpstan-doctrine/pull/506, `DoctrineTypeDriverAwareDescriptor` is not yet used.

### DIFF
--- a/src/Doctrine/Driver/DriverType.php
+++ b/src/Doctrine/Driver/DriverType.php
@@ -1,0 +1,167 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Doctrine\Driver;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\IBMDB2\Driver as IbmDb2Driver;
+use Doctrine\DBAL\Driver\Mysqli\Driver as MysqliDriver;
+use Doctrine\DBAL\Driver\OCI8\Driver as Oci8Driver;
+use Doctrine\DBAL\Driver\PDO\MySQL\Driver as PdoMysqlDriver;
+use Doctrine\DBAL\Driver\PDO\OCI\Driver as PdoOciDriver;
+use Doctrine\DBAL\Driver\PDO\PgSQL\Driver as PdoPgSQLDriver;
+use Doctrine\DBAL\Driver\PDO\SQLite\Driver as PdoSQLiteDriver;
+use Doctrine\DBAL\Driver\PDO\SQLSrv\Driver as PdoSqlSrvDriver;
+use Doctrine\DBAL\Driver\PgSQL\Driver as PgSQLDriver;
+use Doctrine\DBAL\Driver\SQLite3\Driver as SQLite3Driver;
+use Doctrine\DBAL\Driver\SQLSrv\Driver as SqlSrvDriver;
+use mysqli;
+use PDO;
+use SQLite3;
+use Throwable;
+use function get_resource_type;
+use function is_resource;
+use function method_exists;
+use function strpos;
+
+class DriverType
+{
+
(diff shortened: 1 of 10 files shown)


src/Type/Doctrine/Descriptors/FloatType.php
```diff
@@ -26,7 +39,38 @@ public function getWritableToDatabaseType(): Type
 
 	public function getDatabaseInternalType(): Type
 	{
-		return TypeCombinator::union(new \PHPStan\Type\FloatType(), new IntegerType());
```
janedbal, line 29: I tried hard finding at least single case where integer is returned by some driver (in #506), but was not able to find any. So I believe it was just wrong, int is probably never returned.

On the other hand, numeric-string is returned pretty often and was missing here.

src/Type/Doctrine/Descriptors/DecimalType.php
```diff
@@ -29,7 +40,38 @@ public function getWritableToDatabaseType(): Type
 
 	public function getDatabaseInternalType(): Type
 	{
-		return TypeCombinator::union(new FloatType(), new IntegerType());
+		return TypeCombinator::union(
+			new FloatType(),
+			new IntegerType(),
+			new IntersectionType([
```
janedbal: numeric-string was missing here, used by most non-sqlite drivers.

src/Type/Doctrine/Descriptors/FloatType.php
```diff
@@ -26,7 +39,38 @@ public function getWritableToDatabaseType(): Type
...
+				new AccessoryNumericStringType(),
+			]);
+		}
+
+		if (
+			$driverType === DriverType::SQLITE3
+			|| $driverType === DriverType::PDO_SQLITE
+			|| $driverType === DriverType::MYSQLI
+			|| $driverType === DriverType::PDO_MYSQL
+			|| $driverType === DriverType::PGSQL
```
ondrejmirtes: in_array would be nicer

src/Doctrine/Driver/DriverType.php
```diff
@@ -0,0 +1,167 @@
...
+	public const PDO_SQLSRV = 'pdo_sqlsrv';
+	public const PGSQL = 'pgsql';
+	public const SQLITE3 = 'sqlite3';
+	public const SQLSRV = 'sqlsrv';
+
+
+	/**
+	 * @return self::*|null
+	 */
+	public static function detect(Connection $connection, bool $failOnInvalidConnection): ?string
```
ondrejmirtes: This method should be non-static, DriverType should be injected into descriptors as constructor argument.

`bool $failOnInvalidConnection` should be constructor argument here, `Connection` can still be passed as parameter into `detect()`.

ondrejmirtes (changes requested):

src/Type/Doctrine/Descriptors/FloatType.php
janedbal: I'll separate this to another PR.

src/Type/Doctrine/Descriptors/DecimalType.php
janedbal: I'll separate this to another PR.

src/Type/Doctrine/Descriptors/FloatType.php
janedbal: Done

src/Doctrine/Driver/DriverType.php
janedbal: Done

ondrejmirtes:
Thank you!

PR #579, opened 2024-06-25
QueryResultTypeWalker: fix TypedExpression handling

janedbal:
Separated from #506

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultTypeWalker.php
+++ b/src/Type/Doctrine/Query/QueryResultTypeWalker.php
@@ -3,7 +3,8 @@
 namespace PHPStan\Type\Doctrine\Query;
 
 use BackedEnum;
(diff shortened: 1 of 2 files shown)


src/Type/Doctrine/Query/QueryResultTypeWalker.php
```diff
@@ -817,28 +818,15 @@ public function walkSelectExpression($selectExpression): string
 			$resultAlias = $selectExpression->fieldIdentificationVariable ?? $this->scalarResultCounter++;
 			$type = $this->unmarshalType($expr->dispatch($this));
 
-			if (class_exists(TypedExpression::class) && $expr instanceof TypedExpression) {
```
janedbal, line 820: This was dead block, class_exists over interface is always false.

ondrejmirtes:
Thank you.

PR #580, opened 2024-06-25
More precise type inference with unary plus and minus

janedbal:
Separated from #506

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultTypeWalker.php
+++ b/src/Type/Doctrine/Query/QueryResultTypeWalker.php
@@ -1261,7 +1261,19 @@ class QueryResultTypeWalker extends SqlWalker
 		$primary = $factor->arithmeticPrimary;
 
 		$type = $this->unmarshalType($this->walkArithmeticPrimary($primary));
-		$type = TypeUtils::generalizeType($type, GeneralizePrecision::lessSpecific());
+
+		if ($type instanceof ConstantIntegerType && $factor->sign === false) {
+			$type = new ConstantIntegerType($type->getValue() * -1);
+
+		} elseif ($type instanceof IntegerRangeType && $factor->sign === false) {
+			$type = IntegerRangeType::fromInterval(
+				$type->getMax() === null ? null : $type->getMax() * -1,
+				$type->getMin() === null ? null : $type->getMin() * -1
+			);
+
+		} elseif ($type instanceof ConstantFloatType && $factor->sign === false) {
+			$type = new ConstantFloatType($type->getValue() * -1);
+		}
 
 		return $this->marshalType($type);
 	}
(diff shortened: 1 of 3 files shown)


src/Type/Doctrine/Query/QueryResultTypeWalker.php
```diff
@@ -1261,7 +1261,19 @@ public function walkArithmeticFactor($factor): string
 		$primary = $factor->arithmeticPrimary;
 
 		$type = $this->unmarshalType($this->walkArithmeticPrimary($primary));
-		$type = TypeUtils::generalizeType($type, GeneralizePrecision::lessSpecific());
+
+		if ($type instanceof ConstantIntegerType && $factor->sign === false) {
+			$type = new ConstantIntegerType($type->getValue() * -1);
```
ondrejmirtes, line 1266: 1) What about sign=true?
2) Looks like this doesn't have tests, at least from the diff

src/Type/Doctrine/Query/QueryResultTypeWalker.php
```diff
@@ -1261,7 +1261,19 @@ public function walkArithmeticFactor($factor): string
...
+
+		} elseif ($type instanceof IntegerRangeType && $factor->sign === false) {
+			$type = IntegerRangeType::fromInterval(
+				$type->getMax() === null ? null : $type->getMax() * -1,
+				$type->getMin() === null ? null : $type->getMin() * -1
+			);
+
+		} elseif ($type instanceof ConstantFloatType && $factor->sign === false) {
+			$type = new ConstantFloatType($type->getValue() * -1);
```
ondrejmirtes, line 1275: Same questions as above

ondrejmirtes (changes requested):

src/Type/Doctrine/Query/QueryResultTypeWalker.php
janedbal: 1. Sign `true` is `+1`, sign `null` is `1`, those dont change value of constant type
2. Tests are present in origin: [#506 commit](https://github.com/phpstan/phpstan-doctrine/pull/506/commits/d279ca92f5e57c90cb097a2768a44a77c75faaa7#diff-8d4ae04116fed63a04e4f0e23773e255339db33cbb52a2f0bf637cbc426f5ec2). Mostly I didnt add testcases to the old test as that one is MUCH worse. But I can add simple one to this separated PR.

janedbal: The test I'm trying to write produces wrong results as those two are not merged yet:

- #581
- #578 

DQL from the test:

```
SELECT		-o.intColumn as minusInt,
		-o.floatColumn as minusFloat,
		-COUNT(o.intColumn) as minusIntRange
FROM		QueryResult\Entities\One o
```

janedbal: Test added, it will fail once rebased on those PRs.

ondrejmirtes:
Thank you.

PR #581, opened 2024-06-25
Proper aggregate function detection (even deeper in AST)

janedbal:
Separated from #506. 

This is needed for **proper nullability** detection (when no group by is used, single row is always returned). Previously, it was working only when SUM was in first level in select AST. But it can be anywhere inside the tree, so we need pre-walk the AST to detect it.

### DIFF
--- a/src/Type/Doctrine/Query/QueryAggregateFunctionDetectorTreeWalker.php
+++ b/src/Type/Doctrine/Query/QueryAggregateFunctionDetectorTreeWalker.php
@@ -1,0 +1,321 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Doctrine\Query;
+
+use Doctrine\ORM\Query;
+use Doctrine\ORM\Query\AST;
+use function is_string;
+
+class QueryAggregateFunctionDetectorTreeWalker extends Query\TreeWalkerAdapter
+{
+
+	public const HINT_HAS_AGGREGATE_FUNCTION = self::class . '::HINT_HAS_AGGREGATE_FUNCTION';
+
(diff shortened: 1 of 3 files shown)


src/Type/Doctrine/Query/QueryAggregateFunctionDetectorTreeWalker.php
```diff
@@ -0,0 +1,321 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Doctrine\Query;
+
+use Doctrine\ORM\Query;
+use Doctrine\ORM\Query\AST;
+use function is_string;
+
+class QueryAggregateFunctionDetectorTreeWalker extends Query\TreeWalkerAdapter
```
janedbal, line 9: The AST traversal logic is copied from Doctrine's SqlWalker. But since they do double-dispatch which is not possible here, those places are mostly replaced with generic doWalkNode.

I also prefixed all methods with `do` so that this works in dbal3 and dbal4 (where TreeWalker interface lost most of its methods).

tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
```diff
@@ -647,6 +647,24 @@ public function getTestData(): iterable
 			',
 		];
 
+		yield 'aggregate deeper in AST' => [
+			$this->constantArray([
+				[
+					new ConstantStringType('many'),
+					TypeCombinator::addNull(new ObjectType(Many::class)),
```
janedbal, line 654: This wont be nullable in the old implementation, which is wrong.

ondrejmirtes:
Thank you!

PR #582, opened 2024-06-25
ReflectionDescriptor: deduce database internal type based on parent

janedbal:
Separated from #506.

This is crucial for every codebase that utilizes custom types a lot. Previously any `AVG(t.customTypedField)` was mixed. Now, it can produce proper type.


src/Type/Doctrine/Descriptors/ReflectionDescriptor.php
```diff
@@ -57,6 +69,20 @@ public function getWritableToDatabaseType(): Type
 
 	public function getDatabaseInternalType(): Type
 	{
+		$registry = $this->container->getByType(DefaultDescriptorRegistry::class);
+		$parents = $this->reflectionProvider->getClass($this->type)->getParentClassesNames();
```
ondrejmirtes, line 77: Better have a hasClass() guard here

ondrejmirtes (changes requested):

src/Type/Doctrine/Descriptors/ReflectionDescriptor.php
```diff
@@ -57,6 +69,20 @@ public function getWritableToDatabaseType(): Type
...
+		foreach ($parents as $dbalTypeParentClass) {
+			try {
+				// this assumes that if somebody inherits from DecimalType,
+				// the real database type remains decimal and we can reuse its descriptor
+				return $registry->getByClassName($dbalTypeParentClass)->getDatabaseInternalType();
+
+			} catch (DescriptorNotRegisteredException $e) {
+				continue;
+			}
+		}
```
janedbal, line 88: I also have another idea how to implement this without the requirement of known parent. Then, you can extend even abstract `Doctrine\DBAL\Types\Type` and still have proper type (which is not working in the implementation here).

We use it our codebase, but it may feel a bit more hacky:

```php
public function getDatabaseInternalType(): Type {
        /** @var DbalType $dbalType */
        $dbalType = new $this->type();
        $schemaColumn = new SchemaColumn('dummy', $dbalType, []);
        $fullSqlType = strtolower($dbalType->getSQLDeclaration($schemaColumn->toArray(), new MySQL80Platform()));

        $typeName = Strings::replace($fullSqlType, '~^([a-z ]+).*?$~', '$1');

        return match ($typeName) {
            'tinyint',
            'smallint',
            'mediumint',
            'int',
            'bigint' => new IntegerType(),
            'decimal',
            'numeric' => new IntersectionType([new StringType(), new AccessoryNumericStringType()]),
            'double precision' => new FloatType(),
            'varchar',
            'char',
            'binary',
            'tinytext',
            'text',
            'longtext',
            'json',
            'date',
            'datetime',
            'time' => new StringType(),
            default => throw new LogicException("Unexpected type: $fullSqlType"),
        };
    }
```

If you feel that this more generic approach is suitable even for phpstan-doctrine, I can try implementing it in #506 (as ReflectionDescriptor actually need to implement `DoctrineTypeDriverAwareDescriptor`).

src/Type/Doctrine/Descriptors/ReflectionDescriptor.php
janedbal: Lets agree if you want https://github.com/phpstan/phpstan-doctrine/pull/582#discussion_r1652239416 or not first.

src/Type/Doctrine/Descriptors/ReflectionDescriptor.php
ondrejmirtes: What about combining both, to cover more use-cases?

janedbal: If both, I think we can merge this and add that extra layer using `getSQLDeclaration` on top of this later.

src/Type/Doctrine/Descriptors/ReflectionDescriptor.php
janedbal: Added

ondrejmirtes:
Thank you!

PR #583, opened 2024-06-25
Fix wrong coalesce type inference

janedbal:
Separated from #506

Coalesce is pretty nasty function: 

- It picks one of the types given in args and tries to cast all the others to that one (if that one comes to play due to nullability of others).
- The logic differs platform by platform  
  - e.g. `COALESCE(1, 'foo')` is `'1'` in MySQL and `1` in SQLite and SQL server
- The casting logic is not documented for MySQL at all.
- SQL server has this documented, but it would be pretty complex to implement it precisely:
  - [https://learn.microsoft.com/en-us/sql/t-sql/language-elements/coalesce-transact-sql?view=sql-server-ver16&redirectedfrom=MSDN#comparing[…]and-isnull](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/coalesce-transact-sql?view=sql-server-ver16&redirectedfrom=MSDN#comparing-coalesce-and-isnull)
    - pt 2: "COALESCE returns the data type of value with the highest precedence."
  - https://learn.microsoft.com/en-us/sql/t-sql/data-types/data-type-precedence-transact-sql?view=sql-server-ver16&redirectedfrom=MSDN
    - the precedence list

Therefore, we cannot use the constant types if we do not know the casting logic as it results in wrong types deduced.

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultTypeWalker.php
+++ b/src/Type/Doctrine/Query/QueryResultTypeWalker.php
@@ -14,6 +14,7 @@ use Doctrine\ORM\Query\Parser;
 use Doctrine\ORM\Query\ParserResult;
 use Doctrine\ORM\Query\SqlWalker;
 use PHPStan\ShouldNotHappenException;
+use PHPStan\Type\Accessory\AccessoryNumericStringType;
 use PHPStan\Type\BooleanType;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantFloatType;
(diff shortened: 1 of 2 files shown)


src/Type/Doctrine/Query/QueryResultTypeWalker.php
```diff
@@ -1399,6 +1476,11 @@ private function resolveDatabaseInternalType(string $typeName, ?string $enumType
 		return $type;
 	}
 
+	private function canBeNull(Type $type): bool
+	{
+		return !$type->accepts(new NullType(), true)->no();
```
ondrejmirtes: `isSuperTypeOf` is better

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #584, opened 2024-06-25
QueryResultTypeWalker: fix nullability checks over unknown type

janedbal:
Separated from https://github.com/phpstan/phpstan-doctrine/pull/506

Without this fix, the new testcase fails to add nullability and the inferred type does not accept the real result (invalid type was inferred). More `mixed` tests with all possible AST nodes is in the origin PR.

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultTypeWalker.php
+++ b/src/Type/Doctrine/Query/QueryResultTypeWalker.php
@@ -376,7 +376,7 @@ class QueryResultTypeWalker extends SqlWalker
 					new FloatType()
 				);
 
-				if (TypeCombinator::containsNull($exprType)) {
+				if ($this->canBeNull($exprType)) {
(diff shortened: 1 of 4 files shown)


src/Type/Doctrine/Query/QueryResultTypeWalker.php
```diff
@@ -1399,6 +1399,11 @@ private function resolveDatabaseInternalType(string $typeName, ?string $enumType
 		return $type;
 	}
 
+	private function canBeNull(Type $type): bool
+	{
+		return !$type->accepts(new NullType(), true)->no();
```
ondrejmirtes: isSuperTypeOf makes more sense

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

PR #585, opened 2024-06-25
Fix unary minus test

janedbal:
Fix mistake in #580, it was not testing all the newly added condition branches.


ondrejmirtes:
Yes, this makes sense now :) When I saw code about negative numbers and I didn't see any negative numbers in the tests, it was suspicious.

PR #586, opened 2024-06-26
Avoid real connection for type inference

janedbal:
Instead of reading real connection properties, we just read the params used for connection initialization: `driver` & `driverClass` & `driverOptions` ([docs](https://www.doctrine-project.org/projects/doctrine-dbal/en/4.0/reference/configuration.html))


ondrejmirtes:
Thank you.

PR #588, opened 2024-06-27
Revert result inference for scalar method

VincentLanglet:
As recommended here: https://github.com/phpstan/phpstan-doctrine/pull/520#issuecomment-2194065918


ondrejmirtes:
Please explain in your own words why this has to be reverted.

janedbal:
I believe it is summarized here: https://github.com/phpstan/phpstan-doctrine/pull/520#issuecomment-2191843600

ondrejmirtes:
I'm on a phone and I can't load the comment.

I'd like a brief explanation and also some tests that fail when this is not reverted.

VincentLanglet:
> Please explain in your own words why this has to be reverted.

When we're computing the template values of `Query<Tkey, TValue>`, `TValue` represent the result of the Query for the hydration mode `HYDRATE_OBJECT` but this hydration mode does not behave always like others.

For instance, `SELECT o.bigIntColumn from ...` will return
- An array of array of one string with HYDRATE_OBJECT
- An array of array of one string with HYDRATE_ARRAY
- An array of array of one int with HYDRATE_SCALAR
- An int with HYDRATE_SINGLE_SCALAR
- An array of int with HYDRATE_SCALAR_COLUM

I didn't revert HYDRATE_ARRAY, dunno if @janedbal has a counter example for this case.

I would say that this dev is working something like 90% of the time but such false-positive is non acceptable so we should keep mixed until it's fixed.

For non-regression test, we need to directly test hydration mode on different driver so I made https://github.com/phpstan/phpstan-doctrine/pull/587 but I think @janedbal had something else in mind.

ondrejmirtes:
Please add a test case that would fail if the changes weren't reverted.

ondrejmirtes:
Thank you.

PR #591, opened 2024-06-28
QueryResultTypeWalkerTest: make dataset cases clickable in PHPStorm

janedbal:
This was not working due to imperfect PHPStorm detection. Now you can easily run single dataset case.

image

### DIFF
--- a/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
+++ b/tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
@@ -254,9 +254,6 @@ final class QueryResultTypeWalkerTest extends PHPStanTestCase
 	 */
 	public function getTestData(): iterable
 	{
-		$ormVersion = InstalledVersions::getVersion('doctrine/orm');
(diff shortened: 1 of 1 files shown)


tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
```diff
@@ -1325,26 +1246,6 @@ public function getTestData(): iterable
 			',
 		];
 
-		/*yield 'sqrt function' => [
```
ondrejmirtes, line 1328: I don't see this commented out test added again. It's commented out because it fails on some versions, but it's still valuable to re-add it at some point in the future.

ondrejmirtes (changes requested):

tests/Type/Doctrine/Query/QueryResultTypeWalkerTest.php
janedbal: It is tested in platform test, so I removed it here.

ondrejmirtes (approved):

ondrejmirtes:
Thank you.

PR #598, opened 2024-07-11
Fix detection of aggregate functions inside custom functions

janedbal:
We need to check even subnodes of custom functions. Since we cannot double-dispatch from `TreeWalker`, we need node-visitor-like approach that scans all subnodes. Doctrine itself is using only public props, but 3rd party libs [are not](https://github.com/beberlei/DoctrineExtensions/blob/master/src/Query/Mysql/Cast.php#L27), so that is why we do `(array) $node` (`get_mangled_object_vars` is available since PHP 7.4, so we cannot use that).


ondrejmirtes:
Thank you.

PR #599, opened 2024-07-12
Readme: how to type custom DQL functions

janedbal:
I recently tweeted about this: https://x.com/janedbal/status/1810953617123213359

But now realised this was undocumented.


janedbal:
`TypedExpression` is not documented even in Doctrine, so I sent PR even there: https://github.com/doctrine/orm/pull/11550

PR #600, opened 2024-07-12
Fix subselect aggregate functions

zacharylund:
/cc @janedbal 

[6339dff](https://github.com/phpstan/phpstan-doctrine/commit/6339dffef7c3168ebf705b7737e0022d0038d78f) caused PHPStan to think several of my queries with subselects should be nullable:

SELECT t1.col_int, (SELECT COUNT(t2.col_int) FROM Entity t2) FROM Entity t1

This used to return [int, mixed] but now returns [int|null, mixed].

Writing PHPStan code is fairly new to me, but I've attempted to fix this by not checking subselects. I've also added a test that fails before the change and succeeds after.


janedbal:
The fix looks valid to me, thanks!

ondrejmirtes:
Thank you!

PR #604, opened 2024-08-20
EntityRepository::count() returns positive int

### LINKED ISSUE #603: EntityRepository::count should return positive int
Adding this method to the `EntityRepository.stub` solves a few issues where positive integer are expected.

Afaik this method can never return a negative interger.
```
	/**
	 * @param array<string, mixed> $criteria
	 *
	 * @return int<0, max>
	 */
	public function count(array $criteria);
```

Never opened a PR here, so please advice how to proceed.
Also I am puzzled whether `return` or `phpstan-return` or `psalm-return` is the correct way.

kevinpapst:
Fixes #603

The method signature changed slightly between v2 and v3:

https://github.com/doctrine/orm/blob/2.17.x/lib/Doctrine/ORM/EntityRepository.php#L253
https://github.com/doctrine/orm/blob/3.2.x/src/EntityRepository.php#L131

How to handle that? The original code also mentions a `psalm-param`, see
https://github.com/doctrine/orm/blob/2.17.x/lib/Doctrine/ORM/EntityRepository.php#L247

### DIFF
--- a/stubs/EntityRepository.stub
+++ b/stubs/EntityRepository.stub
@@ -71,4 +71,11 @@ class EntityRepository implements ObjectRepository
 	 */
 	public function createQueryBuilder($alias, $indexBy = null);
 
+	/**
+	 * @param array<string, mixed> $criteria
+	 *
+	 * @return int<0, max>
+	 */
+	public function count(array $criteria);
+
 }


ondrejmirtes:
This is fine, only PHPDocs are read from the stubs, this will work. Thank you!

kevinpapst:
Thank you @ondrejmirtes 👍

Issue #509, opened 2023-12-26
ORM\Column with enumType but missing type causes internal error


jspetrak:
``Consider having a BackendEnum class FooEnum. The entity class has following property declaration:

```
#[ORM\Column(name: 'type', nullable: false, enumType: FooEnum::class)]
private FooEnum type;
```

When running PHPStan analysis, it fails with following stacktrace.

```
Internal error: Internal error: Call to a member function getName() on null while analysing file

     Post the following stack trace to https://github.com/phpstan/phpstan/issues/new?template=Bug_report.yaml:

     ##

     vendor\doctrine\orm\lib\Doctrine\ORM\Mapping\DefaultTypedFieldMapper.php(65)

     #0

     vendor\doctrine\orm\lib\Doctrine\ORM\Mapping\ClassMetadataInfo.php(1594):

     Doctrine\ORM\Mapping\DefaultTypedFieldMapper->validateAndComplete()

     #1

     vendor\doctrine\orm\lib\Doctrine\ORM\Mapping\ClassMetadataInfo.php(1644):

     Doctrine\ORM\Mapping\ClassMetadataInfo->validateAndCompleteTypedFieldMapping()

     #2

     vendor\doctrine\orm\lib\Doctrine\ORM\Mapping\ClassMetadataInfo.php(2723):

     Doctrine\ORM\Mapping\ClassMetadataInfo->validateAndCompleteFieldMapping()

     #3

     vendor\doctrine\orm\lib\Doctrine\ORM\Mapping\Driver\AttributeDriver.php(369):

     Doctrine\ORM\Mapping\ClassMetadataInfo->mapField()

     #4

     vendor\phpstan\phpstan-doctrine\src\Doctrine\Mapping\MappingDriverChain.php(31):

     Doctrine\ORM\Mapping\Driver\AttributeDriver->loadMetadataForClass()

     #5

     vendor\doctrine\orm\lib\Doctrine\ORM\Mapping\ClassMetadataFactory.php(149):

     PHPStan\Doctrine\Mapping\MappingDriverChain->loadMetadataForClass()

     #6

     vendor\doctrine\persistence\src\Persistence\Mapping\AbstractClassMetadataFactory.php(343):

     Doctrine\ORM\Mapping\ClassMetadataFactory->doLoadMetadata()

     #7

     vendor\doctrine\persistence\src\Persistence\Mapping\AbstractClassMetadataFactory.php(225):

[...]

Issue #353, opened 2022-07-15
Wrongly usage Doctrine annotation results in PHPStan internal error


tomtomklima:
# Bug report

Run into PHPStan internal error when tried new feature of Doctrine - use native PHP enum as a entity database value (new in version 2.11 https://www.doctrine-project.org/2022/01/11/orm-2.11.html).

How to reproduce: in annotation for Doctrine entity use *wrongly* new construct `enumType=Status:class` - with only one column before class keyword.

### Code snippet that reproduces the problem

Do not have https://phpstan.org/try link, but here is the snippet (note single column instead of correct two):

```
    /**
     * @ORM\Column(type="string", length=32, enumType=Status:class)
     */
    private Status $status;
```

### Expected output

PHPStan should not throw internal error,  but correctly catch invalid construct.

```
Stack trace:
https://github.com/phpstan/phpstan/issues/new?template=Bug_report.md:      
     #0                                                                         
     /application/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/  
     AnnotationException.php(80):                                               
     Doctrine\Common\Annotations\AnnotationException::semanticalError('Couldn'  
     t find c...')                                                              
     phpstan/phpstan#1                                                                         
     /application/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/  
     DocParser.php(1152):                                                       
     Doctrine\Common\Annotations\AnnotationException::semanticalErrorConstants  
     ('OrderStatus:cla...', 'property App\\En...')                              
     phpstan/phpstan#2                                                                         
     /application/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/  
     DocParser.php(1245): Doctrine\Common\Annotations\DocParser->Constant()     
     phpstan/phpstan#3                                                                         
     /application/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/  
     DocParser.php(1300): Doctrine\Common\Annotations\DocParser->PlainValue()   
     phpstan/phpstan#4                                                                         
     /application/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/  
     DocParser.php(1220):                                                       
     Doctrine\Common\Annotations\DocParser->FieldAssignment()                   
     phpstan/phpstan#5                                                                         
     /application/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/  
     DocParser.php(1063): Doctrine\Common\Annotations\DocParser->Value()        
[...]

Issue #320, opened 2022-04-20
Array of enums not supported


ThomasLandauer:
`doctrine/orm` 2.12.0 now supports arrays of enums, see https://github.com/doctrine/orm/pull/9497

This is the syntax:
```php
#[ORM\Column(type: Types::SIMPLE_ARRAY, length: 255, nullable: true, enumType: Foo::class)]
private array $foo = [];
```

phpstan is reporting:

> Property App\Entity\User::$foo type mapping mismatch: backing type string of enum App\Enum\Foo does not match database type array.
> Property App\Entity\User::$foo type mapping mismatch: database can contain App\Enum\Foo|null but property expects array.
> Property App\Entity\User::$foo type mapping mismatch: property can contain array but database expects App\Enum\Foo|null.

seb-jean:
I have the same problem

jeffreymoelands:
Same problem here on `doctrine/orm 2.14.1`

janklan:
It also applies to other types:

```
/**
 * @var ViewSettingsEnum[]
 */
#[ORM\Column(type: 'json', nullable: true, enumType: ViewSettingsEnum::class)]
private ?array $viewSettings = null;
```

results in 

```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   src/Entity/Employee.php                                                                                                                                                                                            
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  96     Property App\Entity\Employee::$viewSettings type mapping mismatch: backing type string of enum App\Enum\ViewSettingsEnum does not match database type array|bool|float|int|JsonSerializable|stdClass|string|null.  
  96     Property App\Entity\Employee::$viewSettings type mapping mismatch: database can contain App\Enum\ViewSettingsEnum|null but property expects array<App\Enum\ViewSettingsEnum>|null.                                 
  96     Property App\Entity\Employee::$viewSettings type mapping mismatch: property can contain array<App\Enum\ViewSettingsEnum>|null but database expects App\Enum\ViewSettingsEnum|null.                                 
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
```

amalricBzh:
Still present with doctrine/orm 3.1.3

siggidiel:
Still present on doctrine/orm 3.2.0

psprokofiev:
[...]

PR #605, opened 2024-08-23
Fix support for `enumType` on array fields

### LINKED ISSUE #478: The type integration does not properly support enumType on an array-based type
Currently, the resolution of doctrine types considers that any field that uses `enumType` will have a PHP type of `new ObjectType($enumType)` (potentially nullable depending on the nullability of the field).

This is a wrong assumption about the way this feature works in the Doctrine ORM. The `enumType` feature supports 2 cases:

- if the converted field value (based on the DBAL type of the field) is a scalar, it gets turned into an enum instance
- if the converted field value is an array, each value of the array gets turned into an enum instance

That second case is not supported by phpstan-doctrine, reporting the wrong type.
Both QueryResultTypeWalker and EntityColumnRule are impacted by that.

stof:
Closes https://github.com/phpstan/phpstan-doctrine/issues/478

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultTypeWalker.php
+++ b/src/Type/Doctrine/Query/QueryResultTypeWalker.php
@@ -19,6 +19,7 @@ use PHPStan\Php\PhpVersion;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\TrinaryLogic;
 use PHPStan\Type\Accessory\AccessoryNumericStringType;
+use PHPStan\Type\ArrayType;
 use PHPStan\Type\BooleanType;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\Constant\ConstantFloatType;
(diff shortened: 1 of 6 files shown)


ondrejmirtes:
Thank you!

ondrejmirtes:
Could you also look into this? https://github.com/phpstan/phpstan-doctrine/issues/320

And also https://github.com/phpstan/phpstan-doctrine/issues/509 + https://github.com/phpstan/phpstan-doctrine/issues/353

Thanks.

src/Type/Doctrine/Query/QueryResultTypeWalker.php
```diff
@@ -2028,7 +2041,7 @@ private function resolveDoctrineType(string $typeName, ?string $enumType = null,
 			$type = TypeCombinator::addNull($type);
 		}
 
-		return $type;
+			return $type;
```
janedbal, line 2044: Looks like some cs check is missing :)

stof: this was actually *introduced* by phpcbf

stof:
@ondrejmirtes I commented on those 3 issues (and 2 of them are not actually related to the support of enums)

Issue #608, opened 2024-09-12
Return of mixed in `$this->createQueryBuilder()->getQuery()->getResult()` when alias is `ev` or `event`


ruudk:
I have no idea what this is, or what causes it.

```php
 // This all works: list<Event>
dumpType($this->createQueryBuilder('e')->getQuery()->getResult());
dumpType($this->createQueryBuilder('my_name')->getQuery()->getResult());
dumpType($this->createQueryBuilder('oo')->getQuery()->getResult());
dumpType($this->createQueryBuilder('abcdef')->getQuery()->getResult());
dumpType($this->createQueryBuilder('user')->getQuery()->getResult());
dumpType($this->createQueryBuilder('av')->getQuery()->getResult());
dumpType($this->createQueryBuilder('eve')->getQuery()->getResult());
dumpType($this->createQueryBuilder('even')->getQuery()->getResult());

// This returns: mixed
dumpType($this->createQueryBuilder('ev')->getQuery()->getResult());
dumpType($this->createQueryBuilder('event')->getQuery()->getResult());
```

I also tried this on a different entity repository, for an entity named differently: Type. There, it gives the same problem.

So what is the cause for `ev` and `event` to behave weird?

janedbal:
Replicated: https://github.com/phpstan/phpstan-doctrine/pull/609

ruudk:
I was just using Xdebug to find the issue:

It's throwing here: 
https://github.com/phpstan/phpstan-doctrine/blob/be66017631e5289e8a3bd8973e59781642da40fd/src/Type/Doctrine/ArgumentsProcessor.php#L61-L66

ruudk:
It seems that `$className` is the alias `ev` now:
![Screenshot 2024-09-12 at 16 39 15@2x](https://github.com/user-attachments/assets/0e3527d4-eff6-4561-a173-cefe7bf445bd)

ruudk:
Interesting:
![Screenshot 2024-09-12 at 16 40 16@2x](https://github.com/user-attachments/assets/20df0c6c-12d1-4b41-ae6c-684b08ca2bfa)
It thinks this ConstantStringType is a ClassStringType

ruudk:
It seems that I have a class called `ev` and `event`. The last one makes sense, somewhere. But the first one? Interesting.

[...]

PR #609, opened 2024-09-12
Failing test for event alias inferred as mixed

janedbal:
Testcase for https://github.com/phpstan/phpstan-doctrine/issues/608


ruudk:
You're so fast ⚡  I added some debug info in https://github.com/phpstan/phpstan-doctrine/issues/608

janedbal:
Core issue fixed in https://github.com/phpstan/phpstan-doctrine/pull/717

PR #617, opened 2024-10-08
Idea how to infer most custom types without descriptors

janedbal:
I believe it would be much better to work out of the box even with custom types. 

Generating custom descriptors is painful and ensuring descriptors exist is disabled by default (`reportUnknownTypes`).

### DIFF
--- a/src/Type/Doctrine/Query/QueryResultTypeWalker.php
+++ b/src/Type/Doctrine/Query/QueryResultTypeWalker.php
@@ -3,8 +3,10 @@
 namespace PHPStan\Type\Doctrine\Query;
 
 use BackedEnum;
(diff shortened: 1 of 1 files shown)


src/Type/Doctrine/Query/QueryResultTypeWalker.php
```diff
@@ -2188,4 +2195,18 @@ private function simpleFloatify(Type $type): Type
 		});
 	}
 
+	private function autodetectDbalTypePhpValue(DbalType $dbalType): Type
```
janedbal, line 2198: Basically the same thing as ReflectionDescriptor does. So the deafult behaviour could be the same as registering ReflectionDescriptor for all registered types.

PR #622, opened 2024-10-29
Add phpdoc for transformation methods into other Collection stubs

### LINKED ISSUE #621: Invalid map in collection.
After update PHPStan to version 1.5.4, the map function of collection does not work anymore.

Example:

```php

use Doctrine\Common\Collections\ArrayCollection;
use Doctrine\Common\Collections\Collection;

class Item {

    private ?Parent  $parent = null;
    
    public function setParent(?Parent $parent): void
    {
        $this->parent = $parent;
    }
}

class Parent {

    /**
     * @var Collection<int, Item>
     */
    private Collection $items;

    public function __construct()
    {
        $this->items = new ArrayCollection();
    }

    public function __clone()
    {
        $this->items = $this->items->map(
            fn (Item $item): Item => (clone $item)->setParent($this)
        );
    }
}
```

The follwing error is raised:

```
------ -------------------------------------------------------------------------------------------------
  Line   src\Parent.php
 ------ -------------------------------------------------------------------------------------------------
  81     Property Parent::$items (Doctrine\Common\Collections\Collection<int, Item>)
         does not accept Doctrine\Common\Collections\ReadableCollection<int, Item>.
         🪪  assign.propertyType
 ------ ------------------------------------------------------------------------------------------------- 
```

Maybe, the `Collection` and `ArrayCollection` stubs must be also updated.

xificurk:
Fixes #621 

This is basically the same road as Doctrine itself took - repeat the modified phpdoc in the child interfaces/classes.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -43,7 +43,6 @@ parameters:
 		- stubs/Persistence/ObjectRepository.stub
 		- stubs/RepositoryFactory.stub
 		- stubs/Collections/ArrayCollection.stub
-		- stubs/Collections/ReadableCollection.stub
 		- stubs/Collections/Selectable.stub
 		- stubs/ORM/AbstractQuery.stub
 		- stubs/ORM/Exception/ORMException.stub
--- a/src/Stubs/Doctrine/StubFilesExtensionLoader.php
+++ b/src/Stubs/Doctrine/StubFilesExtensionLoader.php
@@ -64,8 +64,10 @@ class StubFilesExtensionLoader implements StubFilesExtension
 			$collectionVersion = null;
 		}
 		if ($collectionVersion !== null && strpos($collectionVersion, '1.') === 0) {
+			$files[] = $stubsDir . '/Collections/ReadableCollection1.stub';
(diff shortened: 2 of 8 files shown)


ondrejmirtes:
Please base it on top of 1.5.x (where the bug actually is and already released).

xificurk:
Sorry, base changed...

I don't understand the failure on 7.4 and 8.0, it seems related to doctrine/collections 1.8.

xificurk:
doctrine/collections 1.8 is the first version that got a extracted `ReadableCollection` interface, so `Collection1.stub` used for all 1.x versions is incorrectly inherting from `ReadableCollection` that exists only in 1.8.0.
But I have no idea why the test is failing on that one version where it is actually a correct setup.

xificurk:
Ah, I see, there is no `Collection::map()` in 1.8, it is inherited from `ReadableCollection`, so "incorrect" stub phpdoc is applied on this collections version.

Solved.

ondrejmirtes:
Thank you!

PR #623, opened 2024-11-07
Add api annotation to ObjectMetadataResolver::getClassMetadata

VincentLanglet:
Solve https://github.com/phpstan/phpstan/discussions/11975

Since this method is used a lot in phpstan-doctrine rules, it should be allowed for developer to implements customs rule with it too.

### DIFF
--- a/src/Type/Doctrine/ObjectMetadataResolver.php
+++ b/src/Type/Doctrine/ObjectMetadataResolver.php
@@ -106,6 +106,8 @@ final class ObjectMetadataResolver
 	}
 
 	/**
+	 * @api
+	 *
 	 * @template T of object
 	 * @param class-string<T> $className
 	 * @return ClassMetadata<T>|null


ondrejmirtes:
Thank you.

VincentLanglet:
Thanks @ondrejmirtes, do you think the api annotation can be added to the constructor too ? 
It would allows to write tests without the phpstan "error".:
```
Creating new PHPStan\Type\Doctrine\ObjectMetadataResolver is not covered by backward compatibility promise. The class might change in a minor PHPStan version.
```

https://github.com/phpstan/phpstan-doctrine/pull/625

ondrejmirtes:
Definitely not for the constructor. You should get the object from the container.

VincentLanglet:
> Definitely not for the constructor. You should get the object from the container.

Then how do I do for writing a Unit test like this one:
https://github.com/phpstan/phpstan-doctrine/blob/2.0.x/tests/Rules/Doctrine/ORM/RepositoryMethodCallRuleTest.php#L17

?

ondrejmirtes:
You write a neon config and put the path in getAdditionalConfigFiles.

VincentLanglet:
> You write a neon config and put the path in getAdditionalConfigFiles.

Thanks a lot, i found an example with https://github.com/phpstan/phpstan-doctrine/blob/2.0.x/tests/Rules/Doctrine/ORM/MagicRepositoryMethodCallRuleTest.php

`self::getContainer()->getByType()` works fine !

PR #631, opened 2024-11-20
Fixing tests

VincentLanglet:


composer.json
```diff
@@ -7,7 +7,7 @@
 	],
 	"require": {
 		"php": "^7.2 || ^8.0",
-		"phpstan/phpstan": "^1.12.6"
+		"phpstan/phpstan": "1.12.x-dev as 1.12.12"
```
ondrejmirtes: Just do `^1.12.12`

ondrejmirtes (changes requested):

ondrejmirtes:
Thank you.

VincentLanglet:
@ondrejmirtes This is doing more than fixing tests, it also improve the type inferred with lower/upper case. So it might worth a new release some times... (and maybe @janedbal you want to take a look)

PR #634, opened 2024-12-02
Mark all result methods in AbstractQuery as impure

paranoiq:
The main thing these methods do is a side effect. Methods in Collections are already marked as impure, so this should be nothing new.

This cannot be marked in local stub, because conflicting stub classes are a non-ingnorable error.


ondrejmirtes:
Thank you!

Issue #637, opened 2025-01-10
Symfony uuid


alcohol:
If I understand https://github.com/phpstan/phpstan-doctrine/issues/334, the `uuid` type from Symfony should be recognized if object manager is configured, which we have, eg;

```yaml
parameters:
  doctrine:
    objectManagerLoader: tools/phpstan/object-manager.php
```
However, the following annotations:

```php
    #[ORM\Id]
    #[ORM\Column(type: UuidType::NAME, length: 16, unique: true)]
    #[ORM\GeneratedValue(strategy: 'CUSTOM')]
    #[ORM\CustomIdGenerator(class: UuidGenerator::class)]
    private Uuid $id;
```

still yields:

```
 ------ -------------------------------------------------------------------------------------------------------------------------------------
  Line   src/Doctrine/Entity/Security/RefreshToken.php
 ------ -------------------------------------------------------------------------------------------------------------------------------------
  23     Property Doctrine\Entity\Security\RefreshToken::$id: Doctrine type "uuid" does not have any registered descriptor.
         🪪  doctrine.descriptorNotFound
 ------ -------------------------------------------------------------------------------------------------------------------------------------
```

Looking inside the compiled container I can find:

```php
            'doctrine.entity_managers' => [
                'default' => 'doctrine.orm.default_entity_manager',
            ],
            'doctrine.default_entity_manager' => 'default',
            'doctrine.dbal.connection_factory.types' => [
                'uuid' => [
                    'class' => 'Symfony\\Bridge\\Doctrine\\Types\\UuidType',
                ],
                'ulid' => [
                    'class' => 'Symfony\\Bridge\\Doctrine\\Types\\UlidType',
                ],
            ],
            'doctrine.connections' => [
                'default' => 'doctrine.dbal.default_connection',
            ],
            'doctrine.default_connection' => 'default',
```

alcohol:
Huh, odd. The following:

```php
<?php declare(strict_types=1);

require dirname(__DIR__, 2).'/vendor/autoload.php';

use Polyestershoppen\Kernel;

$kernel = new Kernel('local', true);
$kernel->boot();

$manager = $kernel->getContainer()->get('doctrine')->getManager();

dump(\Doctrine\DBAL\Types\Type::getTypesMap());

return $manager;
```

yields:
```
array:25 [
  "ascii_string" => "Doctrine\DBAL\Types\AsciiStringType"
[...]

PR #639, opened 2025-01-21
symfony uuid type descriptor

alcohol:
References #637

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,10 @@
 		"phpstan/phpstan-strict-rules": "^2.0",
 		"phpunit/phpunit": "^9.6.20",
 		"ramsey/uuid": "^4.2",
-		"symfony/cache": "^5.4"
+		"symfony/cache": "^5.4",
+		"symfony/doctrine-bridge": "^5.4",
+		"symfony/validator": "^5.4",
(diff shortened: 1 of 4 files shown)


alcohol:
Guess I have to add some stubs..?

ondrejmirtes:
No idea why it's failing. Try removing the dependencies from require-dev you just added.

alcohol:
I think it's failing cause a test uses a class from `symfony/validator` but that component is not in the `require-dev` list.

ondrejmirtes:
Also - the descriptor has to be registered, and also - add a test that shows what problem it's solving. Descriptors are already being tested in EntityColumnRuleTest.

alcohol:
That fixed most existing failures; except for deprecations on the very latest version.

alcohol:
@ondrejmirtes can I add the test inside the same class? Am wondering since both Ramsey and Symfony register a 'uuid' type and the one for Ramsey is already tested/used in that class.

ondrejmirtes:
Yeah, sure.

alcohol:
Yeah that didn't work; the error is for the Ramsey uuid type then, not Symfony.

PR #645, opened 2025-03-03
EntityRepository.stub fix nullability in findOneBy

### LINKED ISSUE #644: Deprecation in EntityRepository fails phpstan
Current phpstan doctrine error prevents moving forward to php8.4. No error in php8.3.

Line 45: Deprecated in PHP 8.4: Parameter #2 $orderBy (array) is implicitly nullable via default value null.
https://github.com/phpstan/phpstan-doctrine/blob/815d5ae40ffe3a7788df1cd557e30e474f3aba1c/stubs/EntityRepository.stub#L45

Image

Version info:
phpstan: 2.1.1
phpstan-doctrine: 2.0.1
php: 8.4.4

janedbal:
Fix https://github.com/phpstan/phpstan-doctrine/issues/644

### DIFF
--- a/stubs/EntityRepository.stub
+++ b/stubs/EntityRepository.stub
@@ -42,7 +42,7 @@ class EntityRepository implements ObjectRepository
 	 * @phpstan-param array<string, string>|null $orderBy
 	 * @phpstan-return TEntityClass|null
 	 */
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Thank you.

PR #647, opened 2025-03-12
fix build

staabm:


ondrejmirtes:
Please install phpstan-deprecation-rules here.

ondrejmirtes:
PHPStan build doesn't pass here.

staabm:
fixed the oversight, thanks.

I guess its correct that the tests are failling atm - looks unrelated to the PR, as the master branch also fails like that.

ondrejmirtes:
Thank you.

PR #648, opened 2025-03-13
Mark nullable EntityRepository methods as impure

### LINKED ISSUE #550: Repository methods considered pure
PHPStan 1.10.59, phpstan-doctrine 1.3.62, phpstan-phpunit 1.3.16, bleeding edge + strict rules

I have a piece of code in my tests that roughly looks like this:

```php
self::assertNull($userRepository->findOneBy(['email' => 'foo@example.com']));

// Calls to the class I'm testing that should lead to new database records.

$user = $userRepository->findOneBy(['email' => 'foo@example.com']);
self::assertInstanceOf(User::class, $staff2);
```

PHPStan emits an error:

> Call to static method PHPUnit\Framework\Assert::assertInstanceOf() with 'App\\Entity\\User' and null will always evaluate to false.

If I remove the first `assertNull()` call, the error is gone. Apparently, PHPStan treats the `findOneBy()` call as pure which is a mistake from my POV.

nCrazed:
Resolves #550

### DIFF
--- a/stubs/EntityRepository.stub
+++ b/stubs/EntityRepository.stub
@@ -20,6 +20,7 @@ class EntityRepository implements ObjectRepository
 	 * @phpstan-param int|null $lockMode
 	 * @phpstan-param int|null $lockVersion
 	 * @phpstan-return TEntityClass|null
+	 * @phpstan-impure
 	 */
 	public function find($id, $lockMode = null, $lockVersion = null);
 
@@ -41,6 +42,7 @@ class EntityRepository implements ObjectRepository
 	 * @phpstan-param array<string, mixed> $criteria The criteria.
 	 * @phpstan-param array<string, string>|null $orderBy
 	 * @phpstan-return TEntityClass|null
+	 * @phpstan-impure
 	 */
 	public function findOneBy(array $criteria, ?array $orderBy = null);
 


nCrazed:
At a glance, all of the failures seem to be due to missing Microsoft ODBC Driver which is completely unrelated to the changes in this PR

cs278 (commented):
Should `findAll()`, `findBy()`, `matching()`, `createQueryBuilder()` and `count()` also be marked as impure?

nCrazed:
They don't return a union type so them being pure isn't a bug in the context of type checking

Unless PHPStan has some functionality that relies on caching actual values returned (as opposed to just their type), I don't think it's worth being **that** pedantic about it.

cs278:
> Unless PHPStan has some functionality that relies on caching actual values returned (as opposed to just their type), I don't think it's worth being that pedantic about it.

PHPStan's type system is more descriptive than PHP's: https://phpstan.org/r/ab040ba5-b25a-465e-8c65-3560fa1b20b3

> Pure function always returns the same value if its inputs (object state and arguments) are the same, and has no side effects. Impure function has side effects and its return value might change even if the input does not.
> https://phpstan.org/blog/remembering-and-forgetting-returned-values

nCrazed:
This PR aims to address a specific bug that is actively breaking builds that, to the best of my knowledge, does not affect the methods you've highlighted.

Unless @ondrejmirtes chimes in with desire for those methods to also be changed in this PR I am leaving it as is.

Any larger efforts to address semantic correctness of the stubs in this projects can be taken care off in a separate PR

ondrejmirtes:
Thank you.

PR #650, opened 2025-03-24
Adapt for recent PHPStan

herndlm:
The switch improvements in https://github.com/phpstan/phpstan/releases/tag/2.1.10 caused to detect PHPStan dead code here

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
 	],
 	"require": {
 		"php": "^7.2 || ^8.0",
-		"phpstan/phpstan": "^1.12.12"
+		"phpstan/phpstan": "^1.12.23"
 	},
 	"conflict": {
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
This https://github.com/phpstan/phpstan-src/pull/3900 was merged into 1.12.x so there should be a PR here for 1.5.x which I will merge into 2.0.x. I believe the build fails there too.

herndlm:
ok, adapted for 1.5.x / PHPStan 1.12.23. looks like you have to fix the conflict in composer.json by requiring PHPStan ^2.1.10 when merging up then

ondrejmirtes:
Yeah, that's expected.

ondrejmirtes:
Thank you.

PR #656, opened 2025-04-24
Add missing descriptors for `SmallFloatType` and `EnumType`

### LINKED ISSUE #619: Doctrine DBAL 4.2 ships an Types::ENUM which should be preferred over implementing custom types for ENUM fields.
New developments concerning enum types are out now:

Doctrine DBAL 4.2 ships an Types::ENUM which should be preferred over implementing custom types for ENUM fields.

https://github.com/doctrine/orm/pull/11666
https://github.com/doctrine/orm/issues/11668

The following ways to use enums are yielding `Property App\Entity\Project::$status: Doctrine type "enum" does not have any registered descriptor.`

```
    #[ORM\Column(type: Types::ENUM)]
    private ?ProjectStatus $status = ProjectStatus::Open;
```

```
    #[ORM\Column(type: Types::ENUM, options: ['values' => ['open', 'locked', 'confirmed']])]
    private ?string $status = 'open';
```

janedbal:
- https://github.com/doctrine/dbal/pull/6536 (since dbal 4.2)
- https://github.com/doctrine/dbal/pull/6471 (since dbal 4.1)
- Resolves #619


phpstan-baseline-dbal-4.neon
```diff
@@ -0,0 +1,33 @@
...
+			path: src/Type/Doctrine/Descriptors/SmallFloatType.php
+
+		-
+			message: '#^Method PHPStan\\Type\\Doctrine\\Descriptors\\SmallFloatType\:\:getType\(\) should return class\-string\<Doctrine\\DBAL\\Types\\Type\> but returns string\.$#'
+			identifier: return.type
+			count: 1
+			path: src/Type/Doctrine/Descriptors/SmallFloatType.php
+
+		-
+			message: '#^Class Doctrine\\DBAL\\Types\\EnumType not found\.$#'
```
janedbal, line 28: At first, I tried adding new doctrine versions to composer.json, but it would require bigger effort to adjust this repo to support those everywhere, so I just referenced those not-yet-existing classes.

src/Type/Doctrine/Descriptors/SmallFloatType.php
```diff
@@ -0,0 +1,13 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Doctrine\Descriptors;
+
+class SmallFloatType extends FloatType
```
ondrejmirtes, line 5: No inheritance please

ondrejmirtes (changes requested):
+ tests describing what this actually fixes would be nice

src/Type/Doctrine/Descriptors/EnumType.php
```diff
@@ -0,0 +1,31 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Type\Doctrine\Descriptors;
+
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+
+class EnumType implements DoctrineTypeDescriptor
```
VincentLanglet, line 8: Am I wrong or without this type the data is considered as mixed ?

Now with this, it will report error `foo be 'A'|'B' but is string` on level 8 because the values option is not considered in the following code
```
#[ORM\Column(name: 'foo', type: Types::ENUM, options: ['values' => ['A', 'B'])]
```

janedbal: > Am I wrong or without this type the data is considered as mixed ?

You are correct.

src/Type/Doctrine/Descriptors/SmallFloatType.php
janedbal: Are you sure? I believe SmallFloatType IS FloatType in terms of type descriptor.

janedbal:
> tests describing what this actually fixes would be nice

It is pretty painful to test new doctrine here (due to all the compatibility hacks). But added.

VincentLanglet (commented):
The enum type will introduce new false-positive.

Before
```
/**
   * @Column(type="enum", options={"values"={"a", "b", "c"}})
   * @var 'a'|'b'|'c'
   *
   public $enum;
```

Was reported level 9 (mixed is not 'a'|'b'|'c')
and now will be reported on a lower level (string is not 'a'|'b'|'c')

Would it be possible to read the options ?

janedbal:
> Would it be possible to read the options ?

Yes, implemented.

VincentLanglet:
> > Would it be possible to read the options ?
> 
> Yes, implemented.

Great.

Could add a test for the EntityColumnRule too https://github.com/phpstan/phpstan-doctrine/blob/a1a9efb64708580a9d8b0d150340f7777d2b8aa0/tests/Rules/Doctrine/ORM/EntityColumnRuleTest.php#L35

(I assume we just need to add an enum field in the BrokenEntity https://github.com/phpstan/phpstan-doctrine/blob/a1a9efb64708580a9d8b0d150340f7777d2b8aa0/tests/Rules/Doctrine/ORM/data/MyBrokenEntity.php) 

I assume that
```
/**
   * @Column(type="enum", options={"values"={"a", "b", "c"}})
   * @var 'a'|'b'|'c'
   *
   public $enum;
```
shouldn't be reported but
```
/**
   * @Column(type="enum", options={"values"={"a", "b", "c"}})
   * @var string
   *
   public $enumWithUnprecisedPHPDoc;
```
should be reported

ondrejmirtes:
Thank you.

PR #665, opened 2025-06-13
PlatformTest: drop MSSQL database asserts

janedbal:
- Closes #657
- Since we still have `testUnknownDriver` test, we are not losing anything by this
  - This was there more-less prepared for easier implementation of sqlsrv support (currently not supported)

### DIFF
--- a/tests/Platform/docker/docker-setup.sh
+++ b/tests/Platform/docker/docker-setup.sh
@@ -2,7 +2,7 @@
 set -ex \
   && apt update \
   && apt install -y bash zip libpq-dev libsqlite3-dev \
-  && pecl install xdebug mongodb \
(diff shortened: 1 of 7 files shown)


tests/Platform/docker/docker-setup.sh
```diff
@@ -2,7 +2,7 @@
 set -ex \
   && apt update \
   && apt install -y bash zip libpq-dev libsqlite3-dev \
-  && pecl install xdebug mongodb \
+  && pecl install xdebug mongodb-1.19.4 \
```
janedbal, line 5: Newer versions are incompatible with PHP 8.0

ondrejmirtes:
Thank you!

Issue #659, opened 2025-04-25
Invalid mapping type to property type check for binary mapping since DBAL v.4


v-noskov:
Since DBAL v.4 binary mapping [converts](https://github.com/doctrine/dbal/blob/4.2.x/UPGRADE.md#bc-break-changes-to-handling-binary-fields) DB value to PHP string instead of resource. But PHPStan check produces an error in this case:
```
Property Entity\Foo::$bar type mapping          
         mismatch: database can contain resource|null but property expects     
         string|null.
```

ondrejmirtes:
What code reports this error?

v-noskov:
@ondrejmirtes I've added PR that demonstrates the issue. If you agree, I can try to fix it by changing \PHPStan\Type\Doctrine\Descriptors\BinaryType::getWritableToPropertyType.

PR #669, opened 2025-07-02
Invalid mapping type to property type check for binary mapping since DBAL v.4

v-noskov:
This PR demonstrates the bug described in #659


ondrejmirtes:
Yes, looking at the type implementation https://github.com/doctrine/dbal/blob/4.3.x/src/Types/BinaryType.php, phpstan-doctrine needs to understand that string is okay.

Issue #670, opened 2025-07-02
Invalid nullable property check for simple_array mapping type


v-noskov:
Doctrine DBAL always converts the database value to a PHP array, even when the field contains null. SimpleArrayType mapper converts the null value from DB to [an empty PHP array](https://github.com/doctrine/dbal/blob/4.2.x/src/Types/SimpleArrayType.php#L41).
However, when I put a not-nullable type hint for the simple_array mapping from the nullable DB field: 
```php
/**
 * @var list<string>
 * @ORM\Column(type="simple_array", nullable=true)
 */
private $simpleArrayFromNullableField;
```
PHPStan check produces an error:
`Property PHPStan\Rules\Doctrine\ORM\MyEntity::$simpleArrayFromNullableField type mapping mismatch: database can contain list<string>|null but property expects list<string>.`

v-noskov:
I see two possible ways to address the issue:

1. Always require a non-nullable property when using simple_array mapping. 
This approach aligns the property type with the behaviour of the [SimpleArrayType::convertToPHPValue()](https://github.com/doctrine/dbal/blob/4.2.x/src/Types/SimpleArrayType.php#L41) method, which always returns an array. It ensures type consistency by disallowing null at the PHP level.

2. Ignore property nullability when mapping a nullable database field with simple_array.
This approach gives developers the flexibility to define the property as nullable if they intend to store null. The [SimpleArrayType::convertToDatabaseValue()](https://github.com/doctrine/dbal/blob/4.2.x/src/Types/SimpleArrayType.php#L26) method will handle the null value correctly during persistence.

If maintainers agree with the issue and will choose the desired way, I can try to provide the fix.

PR #671, opened 2025-07-02
Invalid nullable property check for simple_array mapping type

v-noskov:
This PR demonstrates the bug described in #670


ondrejmirtes:
I agree, the type can't return null: https://github.com/doctrine/dbal/blob/d5a5a2165676e15a1805dc1b43d57c0bec7493bc/src/Types/SimpleArrayType.php#L43-L45

PR #678, opened 2025-09-03
Improve EntityColumnRule for enums

### LINKED ISSUE #677: Types::ENUM is wrongly inferred in entityColumnRule
The following code
```
    /**
     * @var self::LOGIN_METHOD_*
     */
    #[ORM\Column(name: 'login_method', type: Types::ENUM, options: ['default' => self::LOGIN_METHOD_BASIC_AUTH, 'values' => self::LOGIN_METHODS])]
    private string $loginMethod = self::LOGIN_METHOD_BASIC_AUTH;
```

now reports on level 8
```
Property App\Entity\User::$loginMethod type mapping mismatch: 
database can contain string but property expects 'BasicAuth'|'SAML'|'SSO'.

doctrine.columnType 
```

This is kinda a consequence of the PR https://github.com/phpstan/phpstan-doctrine/pull/656 (cc @janedbal)

I assume something similar to https://github.com/phpstan/phpstan-doctrine/pull/656/files#diff-7fd356ee7494ac7e9610c9496e854e994cbcef6c4c56d533fc5fed88e4568a47R2030 
should be done to the entityColumnRule.

VincentLanglet:
Closes https://github.com/phpstan/phpstan-doctrine/issues/677

### DIFF
--- a/src/Rules/Doctrine/ORM/EntityColumnRule.php
+++ b/src/Rules/Doctrine/ORM/EntityColumnRule.php
@@ -9,6 +9,7 @@ use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Type\ArrayType;
+use PHPStan\Type\Constant\ConstantStringType;
 use PHPStan\Type\Doctrine\DescriptorNotRegisteredException;
 use PHPStan\Type\Doctrine\DescriptorRegistry;
 use PHPStan\Type\Doctrine\ObjectMetadataResolver;
@@ -21,10 +22,14 @@ use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypehintHelper;
 use PHPStan\Type\TypeTraverser;
 use PHPStan\Type\TypeUtils;
+use PHPStan\Type\UnionType;
 use PHPStan\Type\VerbosityLevel;
 use Throwable;
+use function count;
 use function get_class;
 use function in_array;
+use function is_array;
+use function is_string;
 use function sprintf;
 
 /**
@@ -99,6 +104,26 @@ class EntityColumnRule implements Rule
 
 		$writableToPropertyType = $descriptor->getWritableToPropertyType();
 		$writableToDatabaseType = $descriptor->getWritableToDatabaseType();
+
+		if ($fieldMapping['type'] === 'enum') {
+			$values = $fieldMapping['options']['values'] ?? null;
+			if (is_array($values)) {
+				$enumTypes = [];
+				foreach ($values as $value) {
+					if (!is_string($value)) {
+						$enumTypes = [];
+						break;
+					}
+
+					$enumTypes[] = new ConstantStringType($value);
+				}
+
+				if (count($enumTypes) > 0) {
+					$writableToPropertyType = new UnionType($enumTypes);
+					$writableToDatabaseType = new UnionType($enumTypes);
+				}
+			}
+		}
 
 		$enumTypeString = $fieldMapping['enumType'] ?? null;
 		if ($enumTypeString !== null) {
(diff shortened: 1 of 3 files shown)


ondrejmirtes:
Thank you!

whataboutpereira:
```
    #[ORM\Column(type: Types::ENUM)]
    private CommentType $commentType;
```

```
    #[ORM\Column(type: Types::ENUM, enumType: CommentType::class)]
    private CommentType $commentType;
```

Now started triggering:

```
  Line   src\Entity\CommentNotification.php
 ------ -------------------------------------------------------------------------------------------   
  :19    Property App\Entity\CommentNotification::$commentType type mapping mismatch: backing type    
         string of enum App\Enum\CommentType does not match database type string.
         🪪  doctrine.enumType
```

Possibly because native enum is used in the database?

ondrejmirtes:
Yeah, should be fixed. Please keep the previous version until then.

whataboutpereira:
These are probably related:

```
        #[ORM\Column(type: Types::ENUM, options: ['values' => ['added', 'deleted', 'modified']])]
        private readonly string $action,
```

```
  :27    Property App\Entity\InventoryLog::$action type mapping mismatch: property can contain        
         string but database expects 'added'|'deleted'|'modified'.
         🪪  doctrine.columnType
```
Here the database field is an actual string.

VincentLanglet:
Yes indeed.

I wont be able to make a fix before tuesday or wednesday but I think the fix is easy if you wanna try @whataboutpereira 

The new code should be in a "elseif" after the if "enumTypeString !== null"

(And Im surprised there wasnt test for your situation)

VincentLanglet:
> These are probably related:
> 
> 
> 
> ```
> 
>         #[ORM\Column(type: Types::ENUM, options: ['values' => ['added', 'deleted', 'modified']])]
> 
>         private readonly string $action,
> 
> ```
> 
> 
> 
> ```
> 
>   :27    Property App\Entity\InventoryLog::$action type mapping mismatch: property can contain        
> 
>          string but database expects 'added'|'deleted'|'modified'.
> 
>          🪪  doctrine.columnType
> 
> ```
> 
> Here the database field is an actual string.

The database field is a string yes but I think for some driver Doctrine add an extra checks on those string values, for instance Mysql.

Also if you have listed values in your type, it's not a bad thing to ensure you are not accepting every possible string but only the listed values (or you're loosing the benefit of the enum type.
So for this error I dunno if @ondrejmirtes prefer:
- to not change anything since it's kinda a valid error
- to update my PR in order to just update the writableToPropertyType and let the writableToDatabaseType considered as string in order to have less impact with this PR

whataboutpereira:
> The database field is a string yes but I think for some driver Doctrine add an extra checks on those string values, for instance Mysql.
> 
> Also if you have listed values in your type, it's not a bad thing to ensure you are not accepting every possible string but only the listed values (or you're loosing the benefit of the enum type. So for this error I dunno if @ondrejmirtes prefer:
> 
> * to not change anything since it's kinda a valid error
> * to update my PR in order to just update the writableToPropertyType and let the writableToDatabaseType considered as string in order to have less impact with this PR

It seems this is a non-issue. Adding `/** @var 'added'|'deleted'|'modified' */` to the actual $action fixed these errors. I should really turn these into proper enums. :)

whataboutpereira:
> It seems this is a non-issue. Adding `/** @var 'added'|'deleted'|'modified' */` to the actual $action fixed these errors. I should really turn these into proper enums. :)

Funnily if I go back to 2.0.4 with these added `@var` tags, I get the same errors. So it's one or the other. :)

VincentLanglet:
> > It seems this is a non-issue. Adding `/** @var 'added'|'deleted'|'modified' */` to the actual $action fixed these errors. I should really turn these into proper enums. :)
> 
> 
> 
> Funnily if I go back to 2.0.4 with these added `@var` tags, I get the same errors. So it's one or the other. :)

Of course, the type wasnt precise enough before ; that was the purpose of the Pr

But it introduced a regression we need to fix first so you can eithrr
- use 2.0.4 without the var type
- use 2.0.5 with the var type and ignore the new error

And wait for 2.0.6 and use the var tag.

whataboutpereira:
> * use 2.0.5 with the var type and ignore the new error

I was contemplating just that. Looked at the rule code as well, but unsure how to fix it. :) Thanks!

VincentLanglet:
You can try https://github.com/phpstan/phpstan-doctrine/pull/681 @whataboutpereira

whataboutpereira:
> You can try #681 @whataboutpereira

Thanks! Works as intended now. :)

PR #681, opened 2025-09-09
Fix support for field with both enumType and type: enum

### LINKED ISSUE #679: False positive on enums with 2.0.6
After upgrading from 2.0.5 to 2.0.6, all my enum fields raise `doctrine.enumType` errors: 
> Property App\Foo::$bar type mapping mismatch: backing type string of enum App\Foo\Value does not match database type string.

VincentLanglet:
Closes https://github.com/phpstan/phpstan-doctrine/issues/679


ondrejmirtes:
Thank you!

Issue #674, opened 2025-07-25
Fields are not checked in orderBy in findBy/findOneBy functions


whataboutpereira:
While checking fields in the criteria of findBy and findOneBy functions works nicely, I notice mistyped field names in orderBy are being ignored.

`$this->findBy([], ['dates' => 'desc']);` will happily pass when the actual field is `date`.

Would it be hard to add field checking for orderBy in those repository functions?

Thanks!

VincentLanglet:
It's not that hard to add this I think ; do you want to do the PR @whataboutpereira ?

The rule is here https://github.com/phpstan/phpstan-doctrine/blob/6fe9e9dfe6bb2fe6149c794ec3bb46658e9e9b64/src/Rules/Doctrine/ORM/RepositoryMethodCallRule.php#L69

You can see that `argType` is the first param
https://github.com/phpstan/phpstan-doctrine/blob/6fe9e9dfe6bb2fe6149c794ec3bb46658e9e9b64/src/Rules/Doctrine/ORM/RepositoryMethodCallRule.php#L39

You'll need to do the same for the second param for methods with an order param

whataboutpereira:
Hello and thanks for the pointers! I'll definitely have a look at it.

whataboutpereira:
That worked. Unfortunately it fails to catch faulty passed on arguments like thus.

```
    public function findAll(array $orderBy = ['name' => 'ASC']): array
    {
        return $this->findBy([], $orderBy);
    }
```

ondrejmirtes:
Implemented https://github.com/phpstan/phpstan-doctrine/pull/683

PR #683, opened 2025-09-22
Check field names of orderBy parameter in findBy and findOneBy

whataboutpereira:
Check valid field names are used in orderBy parameter of findBy and findOneBy repository methods (issue #674).

Unfortunately this won't catch a nested call like this:

```
    public function findAll(array $orderBy = ['name' => 'ASC']): array
    {
        return $this->findBy([], $orderBy);
    }
```

### DIFF
--- a/src/Rules/Doctrine/ORM/RepositoryMethodCallRule.php
+++ b/src/Rules/Doctrine/ORM/RepositoryMethodCallRule.php
@@ -21,9 +21,15 @@ class RepositoryMethodCallRule implements Rule
 
 	private ObjectMetadataResolver $objectMetadataResolver;
 
-	public function __construct(ObjectMetadataResolver $objectMetadataResolver)
+	private bool $checkOrderByFields;
+
+	public function __construct(
+		ObjectMetadataResolver $objectMetadataResolver,
+		bool $checkOrderByFields = false
+	)
 	{
 		$this->objectMetadataResolver = $objectMetadataResolver;
+		$this->checkOrderByFields = $checkOrderByFields;
 	}
 
 	public function getNodeType(): string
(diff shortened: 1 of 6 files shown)


staabm:
@ondrejmirtes could you rebase this branch so we can see how/whether mutation testing is doing here?

ondrejmirtes:
Rebased

staabm:
thanks.

test succeeded, as this PR does not contain any `->yes()` or `->no()` invocation, which means no mutations are generated. mutation testing jobs are green as expected.

ondrejmirtes (changes requested):
Sorry for this late review. Given this adds a new error, this should only be happening with Bleeding Edge.

whataboutpereira:
> Sorry for this late review. Given this adds a new error, this should only be happening with Bleeding Edge.

Uh, sorry, forgot the ball is in my court. I'll look how adding it to Bleeding Edge goes.

whataboutpereira:
I tried to inject #[AutowiredParameter(ref: '%featureToggles.bleedingEdge%')] into the rule but that gives an error so I'm out of ideas now.

```
In Resolver.php line 495:
                                                                                                               
  Service 'rules.4' (type of PHPStan\Rules\Doctrine\ORM\RepositoryMethodCallRule): Parameter $bleedingEdge in  
   RepositoryMethodCallRule::__construct() has no class type or default value, so its value must be specified.
```

Also tried adding a doctrine.bleedingEdge to test, but the error stayed the same.

VincentLanglet (approved):

src/Rules/Doctrine/ORM/RepositoryMethodCallRule.php
```diff
@@ -87,6 +87,37 @@ public function processNode(Node $node, Scope $scope): array
...
 		}
 
+		$orderByType = count($node->getArgs()) > 1 ? $scope->getType($node->getArgs()[1]->value) : null;
+
+		if (
+			in_array($methodName, [
+				'findBy',
+				'findOneBy',
+			], true)
+			&& $orderByType !== null
```
VincentLanglet: You could eventually use early return instead
```
if (
			!in_array($methodName, [
				'findBy',
				'findOneBy',
			], true) {
      return $messages;
}
```
and so

VincentLanglet:
> I tried to inject #[AutowiredParameter(ref: '%featureToggles.bleedingEdge%')] into the rule but that gives an error so I'm out of ideas now.
> 
> ```
> In Resolver.php line 495:
>                                                                                                                
>   Service 'rules.4' (type of PHPStan\Rules\Doctrine\ORM\RepositoryMethodCallRule): Parameter $bleedingEdge in  
>    RepositoryMethodCallRule::__construct() has no class type or default value, so its value must be specified.
> ```
> 
> Also tried adding a doctrine.bleedingEdge to test, but the error stayed the same.

I think you can pass the bleedingEdge with the config
like
```
https://github.com/phpstan/phpstan-strict-rules/blob/80870ece60e57b2bc89d31bae545484f7689766f/rules.neon#L161-L164
```
you'll pass `%featureToggles.bleedingEdge%`

src/Rules/Doctrine/ORM/RepositoryMethodCallRule.php
whataboutpereira: Thanks for the pointers, got the bleedingEdge condition working now.

ondrejmirtes (approved):

ondrejmirtes:
Thank you!

PR #684, opened 2025-09-30
Fix PHPStan crashing with single value enum

VincentLanglet:
@ondrejmirtes I discovered a bug introduced by https://github.com/phpstan/phpstan-doctrine/pull/678

When an enum has only one value, I tried to do a Union of a single type and it crash with a ShouldNotHappenException.

I added a reproducer and a fix.

### DIFF
--- a/src/Rules/Doctrine/ORM/EntityColumnRule.php
+++ b/src/Rules/Doctrine/ORM/EntityColumnRule.php
@@ -22,7 +22,6 @@ use PHPStan\Type\TypeCombinator;
 use PHPStan\Type\TypehintHelper;
 use PHPStan\Type\TypeTraverser;
 use PHPStan\Type\TypeUtils;
-use PHPStan\Type\UnionType;
 use PHPStan\Type\VerbosityLevel;
 use Throwable;
 use function count;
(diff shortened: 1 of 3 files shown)


VincentLanglet:
Failures seems unrelated

src/Rules/Doctrine/ORM/EntityColumnRule.php
```diff
@@ -173,8 +172,8 @@ public function processNode(Node $node, Scope $scope): array
 				}
 
 				if (count($enumTypes) > 0) {
-					$writableToPropertyType = new UnionType($enumTypes);
-					$writableToDatabaseType = new UnionType($enumTypes);
+					$writableToPropertyType = TypeCombinator::union(...$enumTypes);
+					$writableToDatabaseType = TypeCombinator::union(...$enumTypes);
```
ondrejmirtes: Think of the machine and do the exact same union call only once :) Thank you.

ondrejmirtes (changes requested):

src/Rules/Doctrine/ORM/EntityColumnRule.php
VincentLanglet: Sure, if you can wait 5 min for the release, I'll do it now.

ondrejmirtes:
Thank you.

PR #685, opened 2025-10-03
Mark more method with phpstan-impure

VincentLanglet:
Similar to https://github.com/phpstan/phpstan-doctrine/pull/648 but with more methods
- find method exists in EntityManager/DocumentManager
- find/findOne exists in DocumentRepository
- findAll/findBy/count are also technically impure the same way find is


ondrejmirtes:
Thank you!

PR #686, opened 2025-10-04
Setup mutation testing

staabm:
requires 
- [green CI build](https://github.com/phpstan/phpstan-doctrine/pull/693)
- https://github.com/phpstan/phpstan-doctrine/pull/693
- https://github.com/phpstan/phpstan-doctrine/pull/694

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -236,5 +236,5 @@ jobs:
       - uses: "actions/upload-artifact@v4"
         with:
           # update-packages is not relevant for the download-artifact part, but we need it here to get unique artifact names across all jobs
-          name: "result-cache-${{ matrix.php-version }}${{ matrix.update-packages }}"
+          name: "result-cache-${{ matrix.php-version }}${{ matrix.update-packages && '-updated' || '' }}"
           path: "tmp/resultCache.php"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -241,5 +241,5 @@ jobs:
       - uses: "actions/upload-artifact@v4"
         with:
           # "update-packages" is not relevant for the download-artifact counterpart, but we need it here to get unique artifact names across all jobs
-          name: "result-cache-${{ matrix.php-version }}${{ matrix.update-packages && '-updated' || '' }}"
+          name: "result-cache-${{ matrix.php-version }}${{ matrix.update-packages && '-packages-updated' || '' }}"
           path: "tmp/resultCache.php"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -182,7 +182,7 @@ jobs:
         if: matrix.php-version == '8.2' || matrix.php-version == '8.3' || matrix.php-version == '8.4'
         run: |
           git fetch --depth=1 origin $GITHUB_BASE_REF
-          infection --git-diff-base=origin/$GITHUB_BASE_REF --git-diff-filter=AM --git-diff-lines --ignore-msi-with-no-mutations --min-msi=100 --min-covered-msi=100
+          infection --git-diff-base=origin/$GITHUB_BASE_REF --git-diff-lines --ignore-msi-with-no-mutations --min-msi=100 --min-covered-msi=100
 
   static-analysis:
     name: "PHPStan"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -169,7 +169,7 @@ jobs:
           php-version: "${{ matrix.php-version }}"
           ini-file: development
           extensions: pdo, mysqli, pgsql, pdo_mysql, pdo_pgsql, pdo_sqlite, mongodb
-          tools: infection
+          tools: infection:0.31.2
 
       - name: "Allow installing on PHP 8.4"
         if: matrix.php-version == '8.4'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -169,6 +169,7 @@ jobs:
           php-version: "${{ matrix.php-version }}"
           ini-file: development
           extensions: pdo, mysqli, pgsql, pdo_mysql, pdo_pgsql, pdo_sqlite, mongodb
+          tools: infection
 
       - name: "Allow installing on PHP 8.4"
         if: matrix.php-version == '8.4'
@@ -179,8 +180,7 @@ jobs:
 
       - name: "Run infection"
         if: matrix.php-version == '8.2' || matrix.php-version == '8.3' || matrix.php-version == '8.4'
-        run: make infection
-        #run: vendor/bin/infection --git-diff-filter=AM
+        run: vendor/bin/infection --git-diff-lines --ignore-msi-with-no-mutations --min-msi=100 --min-covered-msi=100
 
   static-analysis:
     name: "PHPStan"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -180,7 +180,9 @@ jobs:
 
       - name: "Run infection"
         if: matrix.php-version == '8.2' || matrix.php-version == '8.3' || matrix.php-version == '8.4'
-        run: vendor/bin/infection --git-diff-lines --ignore-msi-with-no-mutations --min-msi=100 --min-covered-msi=100
+        run: |
+          git fetch --depth=1 origin $GITHUB_BASE_REF
+          infection --git-diff-base=origin/$GITHUB_BASE_REF --git-diff-filter=AM --ignore-msi-with-no-mutations --min-msi=100 --min-covered-msi=100
 
   static-analysis:
     name: "PHPStan"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -235,5 +235,6 @@ jobs:
 
       - uses: "actions/upload-artifact@v4"
         with:
-          name: "result-cache-${{ matrix.php-version }}"
+          # update-packages is not relevant for the download-artifact part, but we need it here to get unique artifact names across all jobs
+          name: "result-cache-${{ matrix.php-version }}${{ matrix.update-packages }}"
           path: "tmp/resultCache.php"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -179,7 +179,6 @@ jobs:
         run: "composer install --no-interaction --no-progress"
 
       - name: "Run infection"
-        if: matrix.php-version == '8.2' || matrix.php-version == '8.3' || matrix.php-version == '8.4'
         run: |
           git fetch --depth=1 origin $GITHUB_BASE_REF
           infection --git-diff-base=origin/$GITHUB_BASE_REF --git-diff-lines --ignore-msi-with-no-mutations --min-msi=100 --min-covered-msi=100
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -178,6 +178,11 @@ jobs:
       - name: "Install dependencies"
         run: "composer install --no-interaction --no-progress"
 
+      - uses: "actions/download-artifact@v4"
+        with:
+          name: "result-cache"
+          path: "/tmp/"
+
       - name: "Run infection"
         run: |
           git fetch --depth=1 origin $GITHUB_BASE_REF
@@ -227,3 +232,8 @@ jobs:
 
       - name: "PHPStan"
         run: "make phpstan"
+
+      - uses: "actions/upload-artifact@v4"
+        with:
+          name: "result-cache"
+          path: "/tmp/resultCache.php"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -181,7 +181,7 @@ jobs:
       - uses: "actions/download-artifact@v4"
         with:
           name: "result-cache"
-          path: "/tmp/"
+          path: "${{ runner.temp }}"
 
       - name: "Run infection"
         run: |
@@ -236,4 +236,4 @@ jobs:
       - uses: "actions/upload-artifact@v4"
         with:
           name: "result-cache"
-          path: "/tmp/resultCache.php"
+          path: "${{ runner.temp }}/resultCache.php"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -145,6 +145,43 @@ jobs:
       - name: "Tests"
         run: "make tests"
 
+  mutating-testing:
+    name: "Tests"
+    runs-on: "ubuntu-latest"
+    needs: "tests"
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version:
+          - "8.2"
+          - "8.3"
+          - "8.4"
+
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v5
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "pcov"
+          php-version: "${{ matrix.php-version }}"
+          ini-file: development
+          extensions: pdo, mysqli, pgsql, pdo_mysql, pdo_pgsql, pdo_sqlite, mongodb
+
+      - name: "Allow installing on PHP 8.4"
+        if: matrix.php-version == '8.4'
+        run: "composer config platform.php 8.3.99"
+
+      - name: "Install dependencies"
+        run: "composer install --no-interaction --no-progress"
+
+      - name: "Run infection"
+        if: matrix.php-version == '8.2' || matrix.php-version == '8.3' || matrix.php-version == '8.4'
+        run: make infection
+        #run: vendor/bin/infection --git-diff-filter=AM
+
   static-analysis:
     name: "PHPStan"
     runs-on: "ubuntu-latest"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -146,7 +146,7 @@ jobs:
         run: "make tests"
 
   mutating-testing:
-    name: "Tests"
+    name: "Mutating Testing"
     runs-on: "ubuntu-latest"
     needs: "tests"
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -181,7 +181,7 @@ jobs:
       - uses: "actions/download-artifact@v4"
         with:
           name: "result-cache"
-          path: "${{ runner.temp }}"
+          path: "tmp/"
 
       - name: "Run infection"
         run: |
@@ -236,4 +236,4 @@ jobs:
       - uses: "actions/upload-artifact@v4"
         with:
           name: "result-cache"
-          path: "${{ runner.temp }}/resultCache.php"
+          path: "tmp/resultCache.php"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -182,7 +182,7 @@ jobs:
         if: matrix.php-version == '8.2' || matrix.php-version == '8.3' || matrix.php-version == '8.4'
         run: |
           git fetch --depth=1 origin $GITHUB_BASE_REF
-          infection --git-diff-base=origin/$GITHUB_BASE_REF --git-diff-filter=AM --ignore-msi-with-no-mutations --min-msi=100 --min-covered-msi=100
+          infection --git-diff-base=origin/$GITHUB_BASE_REF --git-diff-filter=AM --git-diff-lines --ignore-msi-with-no-mutations --min-msi=100 --min-covered-msi=100
 
   static-analysis:
     name: "PHPStan"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -148,7 +148,7 @@ jobs:
   mutation-testing:
     name: "Mutation Testing"
     runs-on: "ubuntu-latest"
-    needs: "tests"
+    needs: ["tests", "static-analysis"]
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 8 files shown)


tests/Infection/TrinaryLogicMutator.php
```diff
@@ -0,0 +1,66 @@
...
+                DIFF,
+		);
+	}
+
+	public function getName(): string
+	{
+		return 'TrinaryLogicMutator';
+	}
+
+	public function canMutate(Node $node): bool
```
staabm, line 39: atm this class replaces any call to a method named `yes`, or `no`.
for now it is not scoped to the `TrinaryLogic` class.

staabm:
current challenge: figure out a way to kill mutants across github action jobs - I will think about that

phpstan.neon
```diff
@@ -18,6 +18,7 @@ parameters:
 		- tests/*/data-attributes/*
 		- tests/*/data-php-*/*
 		- tests/Rules/Doctrine/ORM/entity-manager.php
+		- tests/Infection/
```
ondrejmirtes, line 28: Why exclude it?

.github/workflows/platform-test.yml
```diff
@@ -64,6 +65,20 @@ jobs:
...
+        with:
+          coverage: "pcov"
+          php-version: "${{ matrix.php-version }}"
+          ini-file: development
+          extensions: pdo, mysqli, pgsql, pdo_mysql, pdo_pgsql, pdo_sqlite, mongodb
+
+      - name: "Run infection"
+        if: matrix.update-packages == '' && (matrix.php-version == '8.2' || matrix.php-version == '8.3' || matrix.php-version == '8.4')
+        run: make infection
+        #run: vendor/bin/infection --git-diff-filter=AM
```
ondrejmirtes: I don't want to run it on "platform tests" but on the traditional tests in "build.yml" instead.

ondrejmirtes: Does this benefit from being run after `vendor/bin/phpunit` or should it be a completely separate job instead?

ondrejmirtes (changes requested):

.github/workflows/platform-test.yml
staabm: I put it here to reduce duplication. we can do it in a separate job, but running infection only makes sense when the test-suite was green before

ondrejmirtes: Alright, should be a separate job with `needs` to specify running after green tests.

phpstan.neon
staabm: because the class depends on infection/infection beeing declared as a dependency.

atm the dependency is installed "on-demand".

I think we will move this class elsewhere when using infection in more repositories

staabm: we need a recent infection version for our setup, and this does not work with PHP 7.4/8.0/8.1.

therefore I cannot declare it easily

ondrejmirtes: Yeah I can imagine a repository similar to https://github.com/phpstan/build-cs once we need this in multiple repositories 😊

ondrejmirtes: > therefore I cannot declare it easily

Yeah, I get it

.github/workflows/build.yml
```diff
@@ -145,6 +145,43 @@ jobs:
       - name: "Tests"
         run: "make tests"
 
+  mutating-testing:
+    name: "Mutating Testing"
```
ondrejmirtes: `Mutation` please

ondrejmirtes (changes requested):
The test is failing sporadically after introducing the random order. It's probably because of: https://github.com/phpstan/phpstan-doctrine/blob/5eaf37b87288474051469aee9f937fc9d862f330/tests/Rules/Doctrine/ORM/entity-manager.php#L34-L37 (which sometimes gets executed before this test and sometimes not).

I guess the test that executes this should clean up after itself.

Makefile
```diff
@@ -32,3 +32,8 @@ phpstan:
 .PHONY: phpstan-generate-baseline
 phpstan-generate-baseline:
 	php vendor/bin/phpstan analyse -l 8 -c phpstan.neon src tests -b phpstan-baseline.neon
+
+.PHONY: infection
+infection:
+	composer require --dev infection/infection -W --ignore-platform-req=ext-mongodb
+	vendor/bin/infection --logger-github=false
```
ondrejmirtes: Why turning off the logger-github?

ondrejmirtes: Also personally I wouldn't add this to `Makefile`. We'll need a more complex logic in the GHA job.

ondrejmirtes (changes requested):
Also I don't see PHPStan configured as mutant killer here.

ondrejmirtes:
Some comments posted by me disappeared. I linked the [--git-diff-filter](https://infection.github.io/guide/command-line-options.html?#git-diff-filter) / [--git-diff-base](https://infection.github.io/guide/command-line-options.html?#git-diff-base) / [--git-diff-lines](https://infection.github.io/guide/command-line-options.html?#git-diff-lines) options

ondrejmirtes (changes requested):
The PHAR distribution is the recommended one for installing Infection. I think we could just add `infection` to `tools` in setup-php action.

Makefile
staabm: added it to makefile to easy running infection locally (to reproduce CI problems)

staabm: > Why turning off the logger-github?

running infection in 3 php versions likely spams the 'files changed tab' with annotations.
I think we need a separate job which accumulates the separate job results and merges them togehter so we get a single final result with less false positives (mutation job on php 8.3 might error about a mutation which can only be killed in the php 8.1 test-suite)

staabm:
> Some comments posted by me disappeared. I linked the [--git-diff-filter](https://infection.github.io/guide/command-line-options.html?#git-diff-filter) / [--git-diff-base](https://infection.github.io/guide/command-line-options.html?#git-diff-base) / [--git-diff-lines](https://infection.github.io/guide/command-line-options.html?#git-diff-lines) options

yeah, I did not yet activate these options, because then there would be no results reported in this PR.

staabm:
> The PHAR distribution is the recommended one for installing Infection. I think we could just add `infection` to `tools` in setup-php action.

I think this only make sense in case we don't want run infection locally. the more we use GitHub Actions only stuff, the harder it is to run the very same setup on a local computer.

staabm:
> The test is failing sporadically after introducing the random order. It's probably because of:

I did not yet see a test, which started failling more often since random order was enabled.
which particular test do you mean?

the build on the master branch is broken long before random order with the same errors

Makefile
ondrejmirtes: > mutation job on php 8.3 might error about a mutation which can only be killed in the php 8.1 test-suite

I had this concern as well but I talked to Maks about this and given that Infection would only mutate covered lines then this shouldn't happen. If a test ran on PHP 8.3 covers a line and we mutate that line, it should fail the test on 8.3.

ondrejmirtes:
> I think this only make sense in case we don't want run infection locally. 

Yeah, I think we don't. Given the support for many PHP versions (and that we only typically have a single PHP version running locally), I think it's unlikely we'd run Infection locally.

Also I'm thinking we'll be passing GitHub Actions-specific env variables to the CLI options anyway.

Makefile
staabm: > it should fail the test on 8.3.

but the very same mutation will happen on 8.1/8.2 and no test will kill the mutation (in case it is a PHP 8.3 only test)

ondrejmirtes:
Why is the build green even when there are escaped mutants?

Makefile
ondrejmirtes: If this was really true Infection would be unusable for us. We'd have to be able to invoke each phase separately and modify the results before feeding them to the next phase. (Run tests on each PHP version, feed the data into Infection, let it mutate the sources, run tests on each PHP version again ourselves and so on.)

staabm:
> Why is the build green even when there are escaped mutants?

because we did not yet define a "min MSI", meaning there is no minimum threshold yet.
if we want the test to fail when at least a single escaped mutant occurs, we need min-msi=100

ondrejmirtes:
Yeah, I want that, alongside the Git-filtering options.

We can demonstrate the failing by adding some new change here in this PR temporarily.

Makefile
staabm: I think we could run infection on multiple php versions in parallel.. collect the mutations afterwards in a new followup github action and report only those as a problem, which were not killed in any of the previous jobs.
(as long as we do not do source transformations, the mutations should be the same on all 3 jobs)

alternatively we could run infection on a single php version only

staabm:
> Also I don't see PHPStan configured as mutant killer here.

I think, as long as we only mutate with `TrinaryLogicMutator` its pretty unlikely that the PHPStan killer will help us.
in addition the process is already slow and adding the PHPStan killer to the mix will make it a even slower

ondrejmirtes:
I really want it though. I don't want to write tests for things that would error in PHPStan. Also we'd have more incentives to make it faster 😊

.github/workflows/build.yml
```diff
@@ -145,6 +145,44 @@ jobs:
       - name: "Tests"
         run: "make tests"
 
+  mutation-testing:
+    name: "Mutation Testing"
+    runs-on: "ubuntu-latest"
+    needs: ["tests", "static-analysis"]
```
staabm, line 151: Phpunit killer requires a green test run
Phpstan killer requires a green phpstan run

ondrejmirtes:
Make sure to carry over PHPStan's result cache, probably with upload and download-artifact actions. It should speed up the build significantly.

tests/Infection/TrinaryLogicMutator.php
staabm: atm this mutator works for these classes in the phpstan-src codebase - which I think is not a bad thing 

grafik

.github/workflows/build.yml
```diff
@@ -145,6 +145,49 @@ jobs:
...
+        uses: actions/checkout@v5
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "pcov"
+          php-version: "${{ matrix.php-version }}"
+          ini-file: development
+          extensions: pdo, mysqli, pgsql, pdo_mysql, pdo_pgsql, pdo_sqlite, mongodb
+          tools: infection
```
maks-rafalko: question: should we here use `infection:0.31.2` something like this, to avoid automatic infection upgrades? I think this can lead to unexpected results if the version will be always the latest, as we may have potential BC breaks on Infection side while on 0.x

ondrejmirtes: Yes please 😊

staabm:
@maks-rafalko I wonder why infection treats this PHPStan result as a killed mutant?
it reads like PHPStan did not report an error at all, so I would expect a escaped mutant.

infection.log:
```
Escaped mutants:
================

Timed Out mutants:
==================

Skipped mutants:
================

Killed by Test Framework mutants:
=================================

Killed by Static Analysis mutants:
==================================

1) /home/runner/work/phpstan-doctrine/phpstan-doctrine/src/Rules/Doctrine/ORM/QueryBuilderDqlRule.php:74    [M] TrinaryLogicMutator [ID] e7038292ed4647a2b1851489639d8e24

@@ @@
         }
         // testing stuff
         $obj = (new ObjectType('Doctrine\ORM\QueryBuilder'))->isSuperTypeOf($calledOnType);
-        if ($obj->yes()) {
+        if (!$obj->no()) {
             $x = 1;
         } else {
             $x = 2;

$ '/home/runner/work/phpstan-doctrine/phpstan-doctrine/vendor/bin/phpstan' '--tmp-file=/tmp/infection/mutant.e7038292ed4647a2b1851489639d8e24.infection.php' '--instead-of=/home/runner/work/phpstan-doctrine/phpstan-doctrine/src/Rules/Doctrine/ORM/QueryBuilderDqlRule.php' '--configuration=/tmp/infection/phpstan.e7038292ed4647a2b1851489639d8e24.infection.neon' '--error-format=json' '--no-progress' '-vv' '--fail-without-result-cache'
  {"totals":{"errors":0,"file_errors":0},"files":{},"errors":[]}
  
  Result cache not used because extension file /home/runner/work/phpstan-doctrine/phpstan-doctrine/src/Rules/Doctrine/ORM/QueryBuilderDqlRule.php hash does not match.
  Result cache was not saved because of --tmp-file and --instead-of CLI options passed (editor mode).
  Elapsed time: 29 seconds
  Used memory: 234.5 MB


Errors mutants:
===============

Syntax Errors mutants:
======================

```

.github/workflows/build.yml
staabm: thanks - done.

maks-rafalko:
> @maks-rafalko I wonder why infection treats this PHPStan result as a killed mutant?

I would debug the exit code. Here is the logic how Infection determines if it's killed or escaped for PHPStan process output.

https://github.com/infection/infection/blob/466255c58cdc6dbe307012cc077b9088ff3cfaa5/src/StaticAnalysis/PHPStan/Mutant/PHPStanMutantExecutionResultFactory.php#L94

We discussed with @ondrejmirtes that checking non-zero exit code is not reliable, but AFAIK nothing has been implemented on PHPStan side to improve it, so we still check (non-)zero exit code.

staabm:
local debugging reveals: we get a exit code of `2`, because infection internally passes `--fail-without-result-cache` and the result cache is not used as the PHPStan output describes:

```
Result cache not used because extension file /Users/m.staab/dvl/phpstan-doctrine/src/Rules/Doctrine/ORM/QueryBuilderDqlRule.php hash does not match.
Result cache was not saved because of --tmp-file and --instead-of CLI options passed (editor mode).
```

ondrejmirtes:
Yeah, that's tricky. I suppose we should have a special exit code for when the result is green but result cache was not used (so that Infection does not use it to kill mutants).

Or maybe Infection can stop passing --fail-without-result-cache altogether.

staabm:
> Or maybe Infection can stop passing --fail-without-result-cache altogether.

I think thats the way to go. not everyone has the tools to use a result cache in CI.

maks-rafalko:
> Or maybe Infection can stop passing --fail-without-result-cache altogether.

I remember I've added it by @staabm's suggestion to avoid the cases when for some (unexpected) reason we miss the cache, so for me this option did make sense so far.

Can I ask you why do we have here

> Result cache not used because extension file /src/Rules/Doctrine/ORM/QueryBuilderDqlRule.php hash does not match.

Is it the issue that can be fixed here? Or is it expected?

staabm:
> Can I ask you why do we have here
> 
> > Result cache not used because extension file /src/Rules/Doctrine/ORM/QueryBuilderDqlRule.php hash does not match.
> 
> Is it the issue that can be fixed here? Or is it expected?

my guess is that this problem occurs because phpstan-doctrine is a phpstan extension repository (regular non phpstan extension projects would not have this case if I got it right)

staabm:
> I remember I've added it by @staabm's suggestion to avoid the cases when for some (unexpected) reason we miss the cache, so for me this option did make sense so far.

ohh I remember that.. while running infection, we actually have a primed cache (caused by the initial test-run). so we don't need CI tooling for the cache to work.

ondrejmirtes:
This can happen in any project with custom rules or other extensions. When Infection actually decides to mutate an extension file, it's inevitable the result cache won't be used.

maks-rafalko:
> This can happen in any project with custom rules or other extensions. When Infection actually decides to mutate an extension file, it's inevitable the result cache won't be used.

1. so do you say here we have a cache miss issue because Infection mutates the extension's file?
2. does it mean we need to remove passing `--fail-without-result-cache` by Infection?

I'm not sure I understand all the cases to make a decision.

ondrejmirtes:
I think what we're facing here is that we want to make sure the user primed the PHPStan cache before running Infection. But it can still be invalidated in these cases.

--fail-without-result-cache can sometimes fill this purpose but it's not perfect.

staabm:
@ondrejmirtes I think we are ready to land this.

in case you are fine with the current forced-error I would revert it and we can do the first steps with Infection on PHPStan.

As described above, we are likely seeing every mutation 3 times now (once for each PHP Version).
in case we don't want to put a new aggregation/evaluation job after mutations runs, we could start with a single PHP Version only.

src/Rules/Doctrine/ORM/QueryBuilderDqlRule.php
```diff
@@ -69,6 +69,15 @@ public function processNode(Node $node, Scope $scope): array
 			return [];
 		}
 
+		// testing stuff
```
staabm: TODO: revert testing change

.github/workflows/build.yml
```diff
@@ -145,6 +145,54 @@ jobs:
...
+
+      - name: "Run infection"
+        run: |
+          git fetch --depth=1 origin $GITHUB_BASE_REF
+          infection --git-diff-base=origin/$GITHUB_BASE_REF --git-diff-lines --ignore-msi-with-no-mutations --min-msi=100 --min-covered-msi=100 --log-verbosity=all --debug
+
+      - uses: "actions/upload-artifact@v4"
+        with:
+          name: "infection-log-${{ matrix.php-version }}"
+          path: "tmp/infection.log"
```
ondrejmirtes, line 195: We should always upload this, not only when Infection passes

ondrejmirtes (changes requested):
Am I right that all mutants would currently get killed by SA because of the `--fail--without-result-cache` option?

staabm:
> `--fail--without-result-cache`

nope, I [fixed it in infection](https://github.com/infection/infection/pull/2429) and updated this PR to the latest release

see CI results:

grafik

.github/workflows/build.yml
staabm: fixed

ondrejmirtes:
Awesome, thank you!

Do you want me to create `phpstan/build-infection` repository where the mutators could live? 😊

staabm:
> Do you want me to create `phpstan/build-infection` repository where the mutators could live?

sure. we need a plan how to scale it over all repos (or what steps next)

ondrejmirtes:
I imagine it could similarly to build-cs: https://github.com/phpstan/phpstan-doctrine/blob/9b118ac640b163a7cfe3ace4bc79130d2e2a468c/.github/workflows/build.yml#L61-L66

Does the infection "json5" config file format support includes? The only thing I'm looking to DRY is probably the mutators implementations and the `mutators` config key, everything else can be copy-pasted across repositories (because I suspect there will have to be some adjustments).

So I imagine:

1) Let's do the new repository
2) Let's try to reuse it in phpstan-doctrine
3) Then we can use it in more places, I would aggresively put it in phpstan-src probably first

.github/workflows/build.yml
romm: Hi there 👋 

I'm coming from https://github.com/CuyZ/Valinor/pull/721

@ondrejmirtes could you explain the reason why the artifacts should always be uploaded? Also, I'm not familiar with the artifact feature: when it's uploaded we can have access to the files directly?

ondrejmirtes: Because I want to see the logs regardless of whether Infection suceeds or fails.

> I'm not familiar with the artifact feature: when it's uploaded we can have access to the files directly?

Yes, on job summary. Artifacts are useful to pass files between jobs, but also for letting users download them. Example https://github.com/phpstan/phpstan-src/actions/runs/18351063855

romm: Great! Thanks for the info. Have a nice day! 😊

maks-rafalko: @ondrejmirtes I'm curios, in what way it is better than just displaying Infection logs to the `stdout`?

Exactly this Markus showed in this comment: https://github.com/phpstan/phpstan-doctrine/pull/686#issuecomment-3380729033

Any advantages uploaded file has comparing to `stdout`?

ondrejmirtes: There is far more information in infection.log than what's in the linked screenshot. The log shows the JSON output PHPStan printed when it finished.

ondrejmirtes: I had no idea it's possible to have the same thing on stdout that's currently uploaded in infection.log. If it's possible then I'd probably welcome a PR 😊

romm: 👀 waiting for an answer, to modify https://github.com/CuyZ/Valinor/pull/721 before merging 👀

maks-rafalko: If you really want to always see the most detailed information available for each build, then

- to see detailed logs, use `--log-verbosity=all --debug`. You already have it. ([explanation](https://infection.github.io/guide/how-to.html#How-to-debug-Infection))
- to display to `stdout` / `stderr` - `infection.json5` for each logger [supports the following](https://infection.github.io/guide/usage.html):
  - `php://stdout`
  - `php://stderr`

So, the final command would be:

```
infection --log-verbosity=all --debug --show-mutations=0
```

Here, `--show-mutations=0` is needed to not duplicate what is already printed to `stdout`.

Similar approach is used e.g. by [BetterReflection](https://github.com/Roave/BetterReflection/blob/6.66.x/infection.json.dist#L8), they always print the whole list of escaped mutations to `stderr`.

But you will have even more details thatnks to `--log-verbosity=all` and especially `--debug`

config needs to be updated:

```diff
- "text": "tmp/infection.log",
+ "text": "php://stdout",
```

romm: FYI @ondrejmirtes I just submitted https://github.com/infection/infection/pull/2438 — thought you could be interested.

PR #687, opened 2025-10-05
Run tests in random order

staabm:
preparation for infection


VincentLanglet:
Out of curiosity, what's the benefit for infection ?

staabm:
Infection optimzes performance by running only selected/filtered parts of the test-suite. 

For this to work reliably its important that tests define dependencies properly. 
Ordering and dependency problems of a test-suite surface best when running it in random order

ondrejmirtes:
Thank you.

PR #689, opened 2025-10-05
Add support for symfony uuid

### LINKED ISSUE #682: Doctrine type "uuid" does not have any registered descriptor using Symfony UUID
I've set up a working UUID like this as per Symfony documentation:

```
use Doctrine\ORM\Mapping as ORM;
use Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator;
use Symfony\Bridge\Doctrine\Types\UuidType;
use Symfony\Component\Uid\Uuid;

#[ORM\Entity]
class Test
{
    #[ORM\Id]
    #[ORM\Column(type: UuidType::NAME, unique: true)]
    #[ORM\GeneratedValue(strategy: 'CUSTOM')]
    #[ORM\CustomIdGenerator(class: UuidGenerator::class)]
    private ?Uuid $uuid;

    public function getUuid(): ?Uuid
    {
        return $this->uuid;
    }
}
```

However it yields:

```
Property App\Entity\Test::$uuid: Doctrine type "uuid" does not have any registered descriptor.
         🪪  doctrine.descriptorNotFound
```

I see descriptor support for ramsey/uuid, but how can I make it understand the symfony/uid type descriptor?

VincentLanglet:
Closes https://github.com/phpstan/phpstan-doctrine/issues/682

### DIFF
--- a/src/Type/Doctrine/Descriptors/Symfony/UlidTypeDescriptor.php
+++ b/src/Type/Doctrine/Descriptors/Symfony/UlidTypeDescriptor.php
@@ -2,44 +2,31 @@
 
 namespace PHPStan\Type\Doctrine\Descriptors\Symfony;
 
-use PHPStan\Rules\Doctrine\ORM\FakeTestingSymfonyUlidType;
-use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Doctrine\Descriptors\DoctrineTypeDescriptor;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
 use Symfony\Component\Uid\Ulid;
-use function in_array;
-use function sprintf;
 
 class UlidTypeDescriptor implements DoctrineTypeDescriptor
 {
 
-	private const SUPPORTED_UUID_TYPES = [
-		'Symfony\Bridge\Doctrine\Types\UlidType',
-		FakeTestingSymfonyUlidType::class,
-	];
-
+	/**
+	 * @var class-string<\Doctrine\DBAL\Types\Type>
+	 */
 	private string $uuidTypeName;
 
-	public function __construct(
-		string $uuidTypeName
-	)
+	/**
+	 * @param class-string<\Doctrine\DBAL\Types\Type> $uuidTypeName
+	 */
+	public function __construct(string $uuidTypeName)
 	{
-		if (!in_array($uuidTypeName, self::SUPPORTED_UUID_TYPES, true)) {
-			throw new ShouldNotHappenException(sprintf(
-				'Unexpected UUID column type "%s" provided',
-				$uuidTypeName,
-			));
-		}
-
 		$this->uuidTypeName = $uuidTypeName;
 	}
 
 	public function getType(): string
 	{
-		/** @var class-string<\Doctrine\DBAL\Types\Type> */
 		return $this->uuidTypeName;
 	}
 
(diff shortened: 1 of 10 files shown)


whataboutpereira:
Thank you, works nicely here.

src/Type/Doctrine/Descriptors/Symfony/UlidTypeDescriptor.php
```diff
@@ -0,0 +1,64 @@
...
+	];
+
+	private string $uuidTypeName;
+
+	public function __construct(
+		string $uuidTypeName
+	)
+	{
+		if (!in_array($uuidTypeName, self::SUPPORTED_UUID_TYPES, true)) {
+			throw new ShouldNotHappenException(sprintf(
```
ondrejmirtes: Why would you want to limit it like that? Let people do whatever they want.

src/Type/Doctrine/Descriptors/Symfony/UlidTypeDescriptor.php
```diff
@@ -0,0 +1,64 @@
...
+			));
+		}
+
+		$this->uuidTypeName = $uuidTypeName;
+	}
+
+	public function getType(): string
+	{
+		/** @var class-string<\Doctrine\DBAL\Types\Type> */
+		return $this->uuidTypeName;
```
ondrejmirtes, line 28: Instead of this `@var`, put the type above the property declaration.

ondrejmirtes (changes requested):
Same points apply to UuidTypeDescriptor.

ondrejmirtes:
Coding standard failures here

ondrejmirtes:
Thank you.

wjzijderveld:
How would you recommend dealing with the different subtypes? 
`type mapping mismatch: database can contain Symfony\Component\Uid\Uuid but property expects Symfony\Component\Uid\UuidV7`
In essence the error isn't wrong, the database could contain anything in theory, but it feels a bit weird having to specify a less specific typehint on my entity if I know its v7

VincentLanglet:
Is there a benefit to typing to Uuid7 rather than Uuid ?
Extra methods ?

Is it sure to be an Uuid7 ? Even if you migrate manually a value in database directly ?

So far I would say an easy way is to ignore it in the baseline.

wjzijderveld:
The benefits of uuidv7 are in (time) sortability and indexability. Hence we want to be strict about it, as it's important in our systems. But it's a bit of a grey area, we obviously validate its v7 in our domain code where we generate it, so there's an argument to be made that the persistence layer only cares about the fact its a UUID and not about the specific version. 

The argument about manually migrated data seems irrelevant to me, that's exactly what these doctrine  types are for, to be able to verify that during persistence/fetching. With the generic UUID type it also wouldn't guarantee that the database doesn't contain random strings.

VincentLanglet:
> The argument about manually migrated data seems irrelevant to me, that's exactly what these doctrine types are for, to be able to verify that during persistence/fetching. 

But the purpose of the EntityColumn rule from PHPStan is to avoid possible crash. PHPStan has right to say you could have manually inserted a non-uuid7 in the database and it will be loaded as another Uuid.

The same way if you type a nullable field with `private string $myNullableField` PHPStan report the possibly null error even if the doctrine type will always be a string. So it's more an issue about Symfony only exposing one doctrine type `uuid` rather explicit `uuid7` etc (which could be a good feature). 

And again if you think this error is unrelevant, the pattern is easy to ignore.

ondrejmirtes:
Unless the Doctrine type guarantees UUIDv7 will be assigned to the property, this reported error is valid.

PR #690, opened 2025-10-05
Fix BinaryType with dbal 4

### LINKED ISSUE #659: Invalid mapping type to property type check for binary mapping since DBAL v.4

VincentLanglet:
Closes https://github.com/phpstan/phpstan-doctrine/pull/669
Closes https://github.com/phpstan/phpstan-doctrine/issues/659

### DIFF
--- a/tests/Rules/Doctrine/ORM/EntityColumnRuleTest.php
+++ b/tests/Rules/Doctrine/ORM/EntityColumnRuleTest.php
@@ -503,6 +503,9 @@ class EntityColumnRuleTest extends RuleTestCase
 		$this->analyse([__DIR__ . '/data/bug-single-enum.php'], []);
 	}
 
+	/**
+	 * @dataProvider dataObjectManagerLoader
+	 */
 	public function testBug659(?string $objectManagerLoader): void
 	{
 		$this->allowNullablePropertyForRequiredField = false;
--- a/tests/Rules/Doctrine/ORM/EntityColumnRuleTest.php
+++ b/tests/Rules/Doctrine/ORM/EntityColumnRuleTest.php
@@ -28,6 +28,7 @@ use PHPStan\Type\Doctrine\Descriptors\StringType;
 use PHPStan\Type\Doctrine\ObjectMetadataResolver;
 use function array_unshift;
 use function class_exists;
+use function sprintf;
 use function strpos;
 use const PHP_VERSION_ID;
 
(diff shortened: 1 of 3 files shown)


tests/Rules/Doctrine/ORM/EntityColumnRuleTest.php
```diff
@@ -271,12 +271,26 @@ public function testSuperclass(?string $objectManagerLoader): void
...
-				17,
-			],
-		]);
+
+		$dbalVersion = InstalledVersions::getVersion('doctrine/dbal');
+		$hasDbal4 = $dbalVersion !== null && strpos($dbalVersion, '4.') === 0;
+		if ($hasDbal4) {
+			$errors = [
+				[
+					'Property PHPStan\Rules\Doctrine\ORM\MyBrokenSuperclass::$five type mapping mismatch: database can contain string but property expects int.',
```
ondrejmirtes: Instead of putting the entire `$errors` assignment in an if, do the assignment once and only change the part of the string that changes with `sprintf` and ternary operator.

ondrejmirtes (changes requested):

tests/Rules/Doctrine/ORM/EntityColumnRuleTest.php
```diff
@@ -478,4 +486,33 @@ public function testBug679(?string $objectManagerLoader): void
...
+	{
+		$this->allowNullablePropertyForRequiredField = false;
+		$this->objectManagerLoader = $objectManagerLoader;
+
+		$dbalVersion = InstalledVersions::getVersion('doctrine/dbal');
+		$hasDbal4 = $dbalVersion !== null && strpos($dbalVersion, '4.') === 0;
+		if ($hasDbal4) {
+			$errors = [
+				[
+					'Property PHPStan\Rules\Doctrine\ORM\MyEntity659::$binaryResource type mapping mismatch: database can contain string but property expects resource.',
```
ondrejmirtes, line 519: Do the same thing with sprintf here

ondrejmirtes (changes requested):

tests/Rules/Doctrine/ORM/EntityColumnRuleTest.php
VincentLanglet: It's not
- the same property
- the same type
- the same line

But I can still do it

ondrejmirtes: Please remove the last commit, I thought it's the same error and only the type is different. This change really doesn't make sense 😊

And please rebase the PR.

ondrejmirtes:
Thank you.

PR #692, opened 2025-10-06
Run tests using global type overrides in isolation

staabm:
see https://github.com/phpstan/phpstan-doctrine/pull/686#pullrequestreview-3304134426


ondrejmirtes:
Let me know if you mark this as ready (I don't think I have the workflow for this repository to post the comment.)

staabm:
I am not sure this PR fixes anything, but here we go.

ondrejmirtes:
Thank you!

PR #694, opened 2025-10-07
Move phpstan analyse paths into phpstan.neon

staabm:
deduplicate config, so we don't need another copy of it in the [infection setup PR](https://github.com/phpstan/phpstan-doctrine/pull/686).

in addition it eases configuring infection with phpstan, as it has such a misleading option handling.
at best we just have it in the config, so we don't need to fiddle with the additional options

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,8 +27,8 @@ cs-fix:
 
 .PHONY: phpstan
 phpstan:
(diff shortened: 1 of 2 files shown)


ondrejmirtes:
Thank you!

PR #696, opened 2025-10-10
Print infection output instead of uploading the log as artifact

staabm:
utilizes https://github.com/infection/infection/pull/2438


ondrejmirtes:
Thank you!

PR #699, opened 2025-10-18
Utilize build-infection

### LINKED ISSUE #7: Integrate into phpstan-doctrine

staabm:
closes https://github.com/phpstan/build-infection/issues/7


.github/workflows/build.yml
```diff
@@ -178,6 +178,9 @@ jobs:
       - name: "Install dependencies"
         run: "composer install --no-interaction --no-progress"
 
+      - name: "Install build-infection"
+        run: "composer require phpstan/build-infection --no-interaction --no-progress"
```
ondrejmirtes: I don't want this to be an actual Composer package with tags, I just want to `actions/checkout` this into a subdirectory.

ondrejmirtes (changes requested):

.github/workflows/build.yml
staabm: So how to make composer make aware of the class loading..? Using a local composer path repository config?

ondrejmirtes: What needs to load what?

staabm: Infection need to know where to find the mutator class

ondrejmirtes: 1) Checkout phpstan/build-infection
2) Run composer install in the build-infection directory
3) Infection can now find the custom mutator class?

staabm: fixed

infection.json5
```diff
@@ -10,6 +10,7 @@
   "logs": {
     "text": "tmp/infection.log"
   },
+  "bootstrap": "build-infection/vendor/autoload.php",
   "mutators": {
     "@default": false,
     "PHPStan\\Infection\\TrinaryLogicMutator": true
```
ondrejmirtes: I'd like DRY of the config so that the list of mutators is in build-infection and not in every repo, akin to phpcs.xml in build-cs.

ondrejmirtes (changes requested):

infection.json5
staabm: thats not possible atm. feature request is in https://github.com/infection/infection/issues/2453

ondrejmirtes: Yes, that would be a nice natice way. The easy way forward we can do now would be to get the list of mutators from build-infection and generate the final infection.json file here with a help of a PHP script, possibly using https://packagist.org/packages/colinodell/json5.

The script to do it shoud ship as part of build-infection.

staabm: couldn't we start as is? do you already have ideas for more mutators?
we do not yet experience with mutation testing, but already think about how we can scale it - even if we don't yet know how got it will work at all.

I wonder whether its worth the effort to build more complexity into it right now.

ondrejmirtes: Yes, I want to experiment by adding some built-in ones from Infection one by one.

I also want to do diff-based mutation testing which wouldn't be based on AST but by removing whole diff chunks from a PR.

ondrejmirtes: Easy iteration and acting on feedback is really what I always need to do my work 😊 That's why I need the DRY of the configs.

.github/workflows/build.yml
```diff
@@ -240,9 +266,3 @@ jobs:
 
       - name: "PHPStan"
         run: "make phpstan"
-
-      - uses: "actions/upload-artifact@v4"
```
staabm, line 244: instead of uploading/downloading artifacts for caching, I changed to use `actions/cache` because that way it also works on phpstan-src. special in phpstan-src is that the "static-analysis" and "test" jobs are spread across 2 workflows, which makes artifact handling otherwise complicated

.github/workflows/build.yml
```diff
@@ -178,10 +182,32 @@ jobs:
...
+          ref: "1.x"
+
+      - name: "Install build-infection dependencies"
+        working-directory: "build-infection"
+        run: "composer install --no-interaction --no-progress"
+
+      - name: "Configure mutators"
+        run: |
+          php build-infection/bin/infection-config.php \
+            --mutator-class='PHPStan\Infection\TrinaryLogicMutator' \
```
staabm: here we can configure the used mutators by just appending 
`--mutator-class='PHPStan\Infection\TrinaryLogicMutator' \`
as often as required

ondrejmirtes: I'd like the list to be just in build-infection. I don't expect different projects to need different mutators.

.github/workflows/build.yml
```diff
@@ -178,10 +182,27 @@ jobs:
...
+          repository: "phpstan/build-infection"
+          path: "build-infection"
+          ref: "1.x"
+
+      - name: "Install build-infection dependencies"
+        working-directory: "build-infection"
+        run: "composer install --no-interaction --no-progress"
+
+      - name: "Configure mutators"
+        run: cp build-infection/resources/infection.json5 .
```
staabm: ok, here we take the config 1:1 from build-infection

.github/workflows/build.yml
```diff
@@ -178,10 +182,27 @@ jobs:
...
+
+      - name: "Cache Result cache"
+        uses: actions/cache@v4
+        with:
+          path: ./tmp
+          key: "result-cache-v14-${{ matrix.php-version }}-${{ github.run_id }}"
```
ondrejmirtes: Why v14?

.github/workflows/build.yml
ondrejmirtes: Just FYI in phpstan-src I'll also want to add https://github.com/phpstan/phpstan-src/tree/2.1.x/build/PHPStan/Build because there are some tests for that in https://github.com/phpstan/phpstan-src/tree/2.1.x/tests/PHPStan/Build.

So if we're copying the whole config I'd still like to be able to modify it afterward.

ondrejmirtes (changes requested):

.github/workflows/build.yml
staabm: because its just a string and it doesn't matter

ondrejmirtes: Yeah but do v1 instead 😊

.github/workflows/build.yml
staabm: added a `--source-directory='another/path'` option

staabm: hmm I think whether it would make sense to move this `phpstan-src/build/PHPStan/Build` rules into a separate repo, as this might make sense to be re-used in 1st party extension repos?

ondrejmirtes: They're not relevant like that yet, mostly because phpstan-src can use PHP 8+ features thanks to downgrading and the extensions can't.

PR #701, opened 2025-10-24
fixes #700; add TResult usage to AbstractQuery.stub

### LINKED ISSUE #700: TResult is present but unused in AbstractQuery.stub
PR coming...

arderyp:
https://github.com/phpstan/phpstan-doctrine/issues/700

https://github.com/phpstan/phpstan/issues/10906#issuecomment-3429727875

### DIFF
--- a/stubs/ORM/AbstractQuery.stub
+++ b/stubs/ORM/AbstractQuery.stub
@@ -24,6 +24,7 @@ abstract class AbstractQuery
 	}
 
     /**
+     * @phpstan-return TResult[]
      * @phpstan-impure
      * @param string|AbstractQuery::HYDRATE_* $hydrationMode
      */
@@ -56,6 +57,7 @@ abstract class AbstractQuery
     }
 
     /**
+     * @phpstan-return ?TResult
      * @phpstan-impure
      * @param string|AbstractQuery::HYDRATE_*|null $hydrationMode
      * @throws NonUniqueResultException
@@ -65,6 +67,7 @@ abstract class AbstractQuery
     }
 
     /**
+     * @phpstan-return TResult
      * @phpstan-impure
      * @param string|AbstractQuery::HYDRATE_*|null $hydrationMode
      * @throws NonUniqueResultException


ondrejmirtes:
Please add a failing test showing what kind of problem this solves.

VincentLanglet (commented):
TResult is already used by
https://github.com/phpstan/phpstan-doctrine/blob/31a924b443a3c15a851c3a8bd0a9f672d3aedc35/src/Type/Doctrine/Query/QueryResultDynamicReturnTypeExtension.php
which rely on
https://github.com/phpstan/phpstan-doctrine/blob/31a924b443a3c15a851c3a8bd0a9f672d3aedc35/src/Type/Doctrine/HydrationModeReturnTypeResolver.php

And it gives better return type than the one added in the stub because they also rely on TKey

arderyp:
And yet it doesn't work in my environment. @ondrejmirtes Thought it was because I didn't have the config right, but I did have config according to docs as noted above.

So, maybe something is off in my environment, but I have looked in vain for months.  May be some issue that arrived when there is lots of abstraction, inheritance, and generics

If you have an idea let me know, but I've provided examples above

ondrejmirtes:
The OP's issue isn't "TResult is unused". His issue is "getOneOrNullResult returns mixed".

Which is by design because today we do not know the hydration mode.

arderyp:
Yeah that sounds right. My issue is mixed returns. The real ripening being discovered

VincentLanglet:
> If you have an idea let me know, but I've provided examples above

I saw your example https://github.com/phpstan/phpstan-doctrine/issues/700#issuecomment-3444007447

And I think it should work with
```
->getOneOrNullResult(AbstractQuery::HYDRATE_OBJECT);
```

As explained by ondrej, currently we annot be sure about `->getOneOrNullResult()`, because it rely on the default value
https://github.com/doctrine/orm/blob/4f36f0129a55aff44d474089c2c777f8544b1397/src/AbstractQuery.php#L106
but:
- You can use your own Query with a different value
- You can write something like
```
$qb->getQuery()->setHydrationMode(AbstractQuery::HYDRATE_SCALAR)->getOneOrNullResult();
```

I assume this would require a third template parameter on the Query stub...

arderyp:
This is very interesting.  I will test this next week and get back to you!

VincentLanglet:
> This is very interesting. I will test this next week and get back to you!

I'll try to improve this situation with https://github.com/phpstan/phpstan-doctrine/pull/702

arderyp:
holy crap, @VincentLanglet, that solved the headache I've been working around for many months now! How silly of me to not see that option, I am so grateful that you chimed in to point me in the right direction!! If #702 is workable, it seems like a great idea, as hydrating the object seems like a logical default assumption for this package.

Thanks again, you make my week.

PR #702, opened 2025-10-25
Introduce default hydration mode

VincentLanglet:


VincentLanglet:
Currently `Query<mixed>` seems to be considered as an alias for `Doctrine\ORM\Query<null, mixed>`.
How does this works @ondrejmirtes ?

I have some trouble to fix the PHPStan build now I have introduced a third template parameter
https://github.com/phpstan/phpstan-doctrine/actions/runs/18808996244/job/53667365401?pr=702

PR #703, opened 2025-10-26
Added `@covers`-class information to *RuleTests

staabm:
goal: speedup infection process

see https://github.com/phpstan/build-infection/pull/11#discussion_r2462958090


staabm:
we could add this annotations (or attributes equivalent on new phpversions) with a auto-fixable phpstan rule.

doing manually for now to see whether its worth it

ondrejmirtes:
What exactly `@covers` achieves in Infection? What if we fix a bug in NodeScopeResolver but write a test for a specific rule?

PR #705, opened 2025-10-29
Skip DocumentManagerTypeInferenceTest on PHP8+

staabm:
empty dataset is not supported, see https://github.com/phpstan/phpstan-phpunit/issues/251


ondrejmirtes:
Thank you!

PR #706, opened 2025-10-29
Update phpstan-phpunit to 2.0.8

staabm:
- removes no longer necessary return types
- fix wrongly named args in dataproviders

=> fixes required for a later PHPUnit 10 upgrade


tests/Platform/QueryResultTypeWalkerFetchTypeMatrixTest.php
```diff
@@ -630,3803 +630,3803 @@ public static function provideCases(): iterable
...
-			'sqlite' => new ConstantIntegerType(-1),
-			'pdo_pgsql' => new ConstantIntegerType(-1),
-			'pgsql' => new ConstantIntegerType(-1),
-			'mssql' => self::mixed(),
-			'mysqlResult' => -1,
-			'sqliteResult' => -1,
-			'pdoPgsqlResult' => -1,
-			'pgsqlResult' => -1,
-			'mssqlResult' => -1,
+			'dqlTemplate' => 'SELECT -1 FROM %s t',
```
staabm, line 633: I think the error which got triggered by the wrong arg names should be made phpunit11+ only

> Error: Missing parameter $dqlTemplate (string) in call to method PHPStan\Platform\QueryResultTypeWalkerFetchTypeMatrixTest::testPdoMysqlDefault().

staabm: ok, these error only showed up, as the phpstan-phpunit in the builds did not yet include the https://github.com/phpstan/phpstan-phpunit/commit/033f6906f7abc5acbe3c81b642ad9e363e1bb4a0 commit.

all good than

ondrejmirtes:
Thank you!

PR #717, opened 2026-01-28
Fix false positive for aliases matching PHP class names

podlesny-j:
When a QueryBuilder alias like 'event' coincidentally matches an
existing PHP class name (e.g. \Event from ext-event stubs),
isClassString() returns yes and isTransient() returns true, causing
a DynamicQueryBuilderArgumentException. This made the query type
resolve to mixed instead of the correct entity type.

The `isTransient` check in `ArgumentsProcessor::processArgs()` is now scoped to only the first
argument of from/join/innerJoin/leftJoin/rightJoin methods. For all other methods and argument
positions, class-strings that happen to match a PHP class name fall through to constant scalar
handling. This preserves the error for genuinely passing a non-entity class to from/join, while
fixing the false positive for aliases like 'event'.

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com> and janedbal

---  
based on https://github.com/phpstan/phpstan-doctrine/pull/609

### DIFF
--- a/src/Type/Doctrine/ArgumentsProcessor.php
+++ b/src/Type/Doctrine/ArgumentsProcessor.php
@@ -61,12 +61,13 @@ class ArgumentsProcessor
 			if ($value->isClassString()->yes() && count($value->getClassStringObjectType()->getObjectClassNames()) === 1) {
 				/** @var class-string $className */
 				$className = $value->getClassStringObjectType()->getObjectClassNames()[0];
-				if ($this->objectMetadataResolver->isTransient($className)) {
-					throw new DynamicQueryBuilderArgumentException();
+				if (!$this->objectMetadataResolver->isTransient($className)) {
+					$args[] = $className;
+					continue;
 				}
-
-				$args[] = $className;
-				continue;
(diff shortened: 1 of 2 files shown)


src/Type/Doctrine/ArgumentsProcessor.php
```diff
@@ -61,12 +62,14 @@ public function processArgs(
 			if ($value->isClassString()->yes() && count($value->getClassStringObjectType()->getObjectClassNames()) === 1) {
 				/** @var class-string $className */
 				$className = $value->getClassStringObjectType()->getObjectClassNames()[0];
-				if ($this->objectMetadataResolver->isTransient($className)) {
-					throw new DynamicQueryBuilderArgumentException();
+				$isEntityClassArgument = $argIndex === 0 && in_array($methodName, ['from', 'join', 'innerJoin', 'leftJoin', 'rightJoin'], true);
```
janedbal: There is no rightJoin method on QB

janedbal (approved):

ondrejmirtes:
Thank you!

PR #720, opened 2026-02-04
Add literal string types for insert, update and delete methods

### LINKED ISSUE #719: Literal string types missing on Connection::{update,insert,delete}
Using the literalString setting, this package reports non literal string usages in methods like `Connection::executeQuery`. But these types are missing for the update, insert and delete methods. 

These methods all use string concatenation to create the resulting SQL. The `$table` parameter should be a literal string, as well as the
array keys of the `$data` and`$criteria`.

BackEndTea:
Add literal string types for the insert, update and delete methods of the connection class.

I'm not quite sure how to test these changes, as i couldn't find any tests files that check the other literal string types.

Fixes #719

### DIFF
--- a/stubs/DBAL/Connection.stub
+++ b/stubs/DBAL/Connection.stub
@@ -74,4 +74,49 @@ class Connection
      */
     public function transactional(Closure $func);
 
+    /**
+     * Executes an SQL DELETE statement on a table.
+     *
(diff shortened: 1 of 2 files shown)


BackEndTea:
Test failures look unrelated to this PR

ondrejmirtes:
Thank you!

PR #723, opened 2026-02-11
Fix offsetAccess.invalidOffset

VincentLanglet:
Fixes

image


staabm (approved):
Since we already have a assert() a line below, I am fine with it. 

Usually I don't like dev-time only assertions

staabm (approved):

VincentLanglet:
> Since we already have a assert() a line below, I am fine with it.
> 
> Usually I don't like dev-time only assertions

Me too

PR #724, opened 2026-02-18
Skip EntityNotFinalRule when native lazy objects or lazy ghost objects are enabled

### LINKED ISSUE #667: Allow `EntityNotFinalRule` to be turned off when on Doctrine ORM 3.4 + PHP 8.4
With Doctrine ORM 3.4.0 [now](https://github.com/doctrine/orm/pull/11853) supporting native Lazy Objects introduced in PHP 8.4 there is no need for the EntityNotFinalRule anymore (introduced in https://github.com/phpstan/phpstan-doctrine/issues/202)

The new functionality needs to be manually enabled on the Doctrine ORM Configuration using `isNativeLazyObjectsEnabled`:
https://github.com/doctrine/orm/blob/3272e1c0af1fe1db037191848e556bc1d4b7cdbb/src/Configuration.php#L598 

When using Symfony, one needs to enable `enable_native_lazy_objects` on the Doctrine ORM configuration.

Would it be an idea to introduce a parameter that has a default value set to true, to toggle this rule? 

That way, people can disable the rule. 

Maybe when the feature is enabled by default in a next version of Doctrine, the rule can be removed completely. Or disabled by default, in a next major of this extension.

/cc @LeoVie  @ondrejmirtes

ondrejmirtes:
When Doctrine ORM is configured with lazy ghost objects (ORM 2.x) or
native lazy objects (ORM 3.4+), final entities no longer cause problems
because proxy subclasses are not needed. This detects the configuration
from the object manager and suppresses the error accordingly.

Fixes #667

https://claude.ai/code/session_01EHwDbWjcx526ofXQu1KxNz

### DIFF
--- a/compatibility/orm-3-baseline.php
+++ b/compatibility/orm-3-baseline.php
@@ -8,6 +8,8 @@ $ormVersion = InstalledVersions::getVersion('doctrine/orm');
 $hasOrm3 = $ormVersion !== null && strpos($ormVersion, '3.') === 0;
 if ($hasOrm3) {
 	$includes[] = __DIR__ . '/../phpstan-baseline-orm-3.neon';
+} else {
+	$includes[] = __DIR__ . '/../phpstan-baseline-orm-2.neon';
 }
 
 $config = [];
--- a/phpstan-baseline-orm-2.neon
+++ b/phpstan-baseline-orm-2.neon
@@ -1,0 +1,7 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Call to an undefined method Doctrine\\\\ORM\\\\Configuration\\:\\:isNativeLazyObjectsEnabled\\(\\)\\.$#"
+			identifier: method.notFound
+			count: 1
+			path: tests/Rules/Doctrine/ORM/EntityNotFinalRuleTest.php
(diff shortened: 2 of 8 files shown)


ruudk:
@ondrejmirtes The shared claude.ai link cannot be opened by anyone but you. Not sure if you are aware. Thanks for fixing it though!

ondrejmirtes:
Yeah, needs to be polished and finished though 😊 

My main motivation was that I wanted to mention this at https://github.com/phpstan/phpstan/blob/2.2.x/website/errors/doctrine.finalEntity.md / https://phpstan.org/error-identifiers/doctrine.finalEntity and realized it's not fixed yet.

ondrejmirtes:
@phpstan-bot Look at the CI bulld failures.
The pre-existing test for this rule needs to assert no errors based on the same conditions.
There can’t be local ignores but the errors need to be added to phpstan-baseline-orm-3.neon
The newly added test needs to be skipped if the right conditions are not met

And please skip the lazy ghost feature for ORM 2.x. Only keep the native lazy objects for new PHP and Doctrine3.x.

ondrejmirtes:
@phpstan-bot Please check the PHPStan build failures for each version. We will probably need to introduce a new conditional baseline file. There are failures for PHPStan on 7.4 through 8.4 because these run doctrine/orm 2.x.

PHPStan on 8.3 with Doctrine 3 has differenet failures which can be added to phpstan-baseline-orm-3.neon IMHO.

PR #739, opened 2026-03-13
Mark EntityManagerInterface::isOpen() as impure

### LINKED ISSUE #738: PHPStan assumes deterministic result for EntityManagerInterface::isOpen() within same scope
### Bug report

**Version**: PHPStan - PHP Static Analysis Tool 2.1.11

I noticed a situation where PHPStan assumes that the result of `EntityManagerInterface::isOpen()` remains the same within a scope.

However, in real applications the state of the EntityManager can change during execution.

For example, Doctrine may close the EntityManager after an exception during flush() or database interaction.

**Example**:
```php
public function run(EntityManagerInterface $em): void
{
    if ($em->isOpen() === false) {
        return;
    }

    $this->doSomething();

    if ($em->isOpen() === false) {
        $this->logger->warning('EntityManager closed');
    }
}
```

**Output**:
```bash
Strict comparison using === between true and false will always evaluate to false.  
         🪪  identical.alwaysFalse
```

**Phpstan config**:
```config
includes:
    - vendor/phpstan/phpstan-doctrine/extension.neon
    - vendor/phpstan/phpstan-symfony/extension.neon
    - vendor/phpstan/phpstan-strict-rules/rules.neon

parameters:
    level: 8
    strictRules:
        strictArrayFilter: false
    paths:
        - ../../src
    bootstrapFiles:
        - ../../Bootstrap/bootstrap.php
    doctrine:
        objectManagerLoader: phpstan-object-manager.php
    parallel:
        processTimeout: 1000.0
```

**Current workaround**:
```php
// @phpstan-ignore-next-line
if ($em->isOpen() === false) {
```

### Code snippet that reproduces the problem

_No response_

### Expected output

PHPStan should not assume that repeated calls to methods like `isOpen()` return the same value unless explicitly guaranteed.

Ideally there should be a way to mark such methods as **non-deterministic** or **state-dependent**.

**Proposed idea**:
```php
/**
 * @phpstan-nondeterministic
 */
public function isEntityManagerClosed(): bool
{
     return $this-em->isOpen() === false;
}
```

To avoid ignores in the code base.

### Did PHPStan help you today? Did it make you happy in any way?

_No response_

denis-korchagin95:
The result of isOpen() depends on the internal state of the EntityManager
and may change during execution (for example after an exception during flush()).

This change follows the existing approach used for other Doctrine methods
that depend on mutable internal state.

Fixes #738

### DIFF
--- a/stubs/EntityManager.stub
+++ b/stubs/EntityManager.stub
@@ -70,4 +70,9 @@ class EntityManager implements EntityManagerInterface
      */
     public function flush($entity = null);
 
+	/**
+	 * @return bool
+	 * @phpstan-impure
+	 */
+	public function isOpen();
 }
--- a/stubs/EntityManagerDecorator.stub
+++ b/stubs/EntityManagerDecorator.stub
@@ -61,4 +61,9 @@ class EntityManagerDecorator implements EntityManagerInterface
 	 */
 	public function copy($entity, $deep = false);
 
+	/**
+	 * @return bool
+	 * @phpstan-impure
+	 */
+	public function isOpen();
 }
--- a/stubs/EntityManagerInterface.stub
+++ b/stubs/EntityManagerInterface.stub
@@ -78,4 +78,9 @@ interface EntityManagerInterface extends ObjectManager
 	 */
 	public function wrapInTransaction(callable $func);
 
+	/**
+	 * @return bool
+	 * @phpstan-impure
+	 */
+	public function isOpen();
 }


stubs/EntityManager.stub
```diff
@@ -70,4 +70,11 @@ class EntityManager implements EntityManagerInterface
      */
     public function flush($entity = null);
 
+	/**
+	 * @return bool
+	 * @phpstan-impure
+	 *
+	 * @template T
```
VincentLanglet:
```suggestion
```

stubs/EntityManagerDecorator.stub
```diff
@@ -61,4 +61,11 @@ class EntityManagerDecorator implements EntityManagerInterface
 	 */
 	public function copy($entity, $deep = false);
 
+	/**
+	 * @return bool
+	 * @phpstan-impure
+	 *
+	 * @template T
```
VincentLanglet:
```suggestion
```

stubs/EntityManagerInterface.stub
```diff
@@ -78,4 +78,11 @@ interface EntityManagerInterface extends ObjectManager
 	 */
 	public function wrapInTransaction(callable $func);
 
+	/**
+	 * @return bool
+	 * @phpstan-impure
+	 *
+	 * @template T
```
VincentLanglet:
```suggestion
```

.gitignore
```diff
@@ -4,3 +4,4 @@
 /composer.lock
 /.env
 .phpunit.result.cache
+.idea
```
VincentLanglet: Please let this file untouched.

VincentLanglet (approved):

staabm:
since ondrej had the impression this fix might not work, I think we should back it by a test

ondrejmirtes:
It’s fine, it will work.

ondrejmirtes:
I was confused in the original issue 😊

ondrejmirtes:
But more work is necessary here - EntityManagerInterface also has the declaration.

VincentLanglet:
> EntityManagerInterface

This file is already updated in the PR.

There is 
- EntityManager
- EntityManagerDecorator
- EntityManagerInterface

ondrejmirtes:
Perfect, thank you. I was checking it out on my phone, wanted to provide as fast feedback as possible.

PR #740, opened 2026-04-14
Cleanup EntityRepositoryClassReflectionExtension

staabm:
looking into this right now, because we see CI errors related to this lines of code in https://github.com/phpstan/phpstan-src/pull/5449

### DIFF
--- a/src/Reflection/Doctrine/EntityRepositoryClassReflectionExtension.php
+++ b/src/Reflection/Doctrine/EntityRepositoryClassReflectionExtension.php
@@ -10,7 +10,6 @@ use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\Doctrine\ObjectMetadataResolver;
 use PHPStan\Type\IntegerType;
-use PHPStan\Type\NullType;
 use PHPStan\Type\TypeCombinator;
 use function lcfirst;
 use function str_replace;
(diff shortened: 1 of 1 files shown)


src/Reflection/Doctrine/EntityRepositoryClassReflectionExtension.php
```diff
@@ -107,7 +103,7 @@ public function getMethod(ClassReflection $classReflection, string $methodName):
 		} elseif (
 			strpos($methodName, 'findOneBy') === 0
 		) {
-			$returnType = TypeCombinator::union($entityClassType, new NullType());
+			$returnType = TypeCombinator::addNull($entityClassType);
```
staabm, line 106: drive-by cleanup

src/Reflection/Doctrine/EntityRepositoryClassReflectionExtension.php
```diff
@@ -88,10 +87,7 @@ public function getMethod(ClassReflection $classReflection, string $methodName):
 	{
 		$repositoryAncesor = $classReflection->getAncestorWithClassName(ObjectRepository::class);
 		if ($repositoryAncesor === null) {
-			$repositoryAncesor = $classReflection->getAncestorWithClassName(ObjectRepository::class);
-			if ($repositoryAncesor === null) {
-				throw new ShouldNotHappenException();
-			}
```
staabm, line 94: in the past this code was used to support `ObjectRepository` classes from different namespaces (across different major versions of doctrine), see https://github.com/phpstan/phpstan-doctrine/commit/a497a383bd970117b0680557ae12444d72d09095#diff-3cbc418c0bd3c05155f1b1f786e9a6c12efe64d7599b402dbd19299942061b74R47-R53

the remaining code references the very same class which is a leftover of this feature which no longer exists

VincentLanglet (approved):

PR #743, opened 2026-05-04
Fix return type on EntityRepository::matching() generics

VincentLanglet:
cc @staabm Don't release after this merge, I'll plan to do more PR.


tests/DoctrineIntegration/data/repositoryMatching-collection25.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace RepositoryMatching;
```
staabm: this namespaces overlapps with `tests/DoctrineIntegration/data/repositoryMatching-collection26.php`

staabm: it seems we are missing a CI check for unique symbol names (like we have on phpstan-src)

staabm (approved):

PR #744, opened 2026-05-05
Rely on same template name than doctrine

VincentLanglet:
Cf https://github.com/doctrine/DoctrineBundle/blob/9898f367da071d3d81084aa6007ab4d80437e62f/src/Repository/ServiceEntityRepository.php#L35
https://github.com/doctrine/persistence/blob/d138f3ab5f761055cab1054070377cfd3222e368/lib/Doctrine/Persistence/ObjectRepository.php#L10
https://github.com/doctrine/orm/blob/495cd06b9a630f9c38a21ceb249ed008edbe8414/lib/Doctrine/ORM/EntityRepository.php#L36
and so...

This avoid error like
https://github.com/phpstan/phpstan-doctrine/pull/742#issuecomment-4373485328

### DIFF
--- a/stubs/EntityRepository.stub
+++ b/stubs/EntityRepository.stub
@@ -8,26 +8,26 @@ use Doctrine\Common\Collections\Selectable;
 use Doctrine\Persistence\ObjectRepository;
 
 /**
- * @template TEntityClass of object
- * @implements ObjectRepository<TEntityClass>
+ * @template T of object
+ * @implements ObjectRepository<T>
  */
 class EntityRepository implements ObjectRepository
 {
 
-	/** @var class-string<TEntityClass> */
+	/** @var class-string<T> */
 	protected $_entityName;
 
 	/**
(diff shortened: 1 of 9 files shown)


stubs/EntityRepository.stub
```diff
@@ -8,26 +8,26 @@ use Doctrine\Common\Collections\Selectable;
 use Doctrine\Persistence\ObjectRepository;
 
 /**
- * @template TEntityClass of object
- * @implements ObjectRepository<TEntityClass>
+ * @template T of object
```
staabm, line 11: do we need to define a min-doctrine version (composer.json) in which this change is contained in ?

VincentLanglet: Why ? Doctrine never used TEntityClass as template name ; they directly used T

staabm: I see 👍

staabm (approved):

PR #745, opened 2026-05-06
name-collision-detector

staabm:
to prevent duplicate symbols like in https://github.com/phpstan/phpstan-doctrine/pull/743#discussion_r3186143456

inspired by https://github.com/phpstan/phpstan-src/commit/95cdbe577513286c36dcf513fe76f269e8a32125#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2

### DIFF
--- a/.github/workflows/lint-workflows.yml
+++ b/.github/workflows/lint-workflows.yml
@@ -119,17 +119,21 @@ jobs:
     timeout-minutes: 60
 
     steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.14.2
+        with:
+          egress-policy: audit
+
       - name: "Checkout"
-        uses: actions/checkout@v3
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
 
       - name: "Install PHP"
-        uses: "shivammathur/setup-php@v2"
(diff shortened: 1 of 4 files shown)


.github/workflows/lint-workflows.yml
```diff
@@ -111,3 +111,29 @@ jobs:
...
+
+  name-collision:
+    name: "Name Collision Detector"
+
+    runs-on: "ubuntu-latest"
+    timeout-minutes: 60
+
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.14.2
```
staabm, line 123: I am using the same `step-security/harden-runner` than any other workflow in this repo (and in phpstan-src).
I don't plan to update it with this PR (should be a separate PR)

.github/workflows/lint-workflows.yml
```diff
@@ -111,3 +111,29 @@ jobs:
...
+        with:
+          egress-policy: audit
+
+      - name: "Checkout"
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
```
VincentLanglet, line 128: This seems to be an old version, newer version are used for other workflow 
https://github.com/phpstan/phpstan-doctrine/blob/fb8f8746b2594d4668e739677ef1b43025c8b9ac/.github/workflows/build.yml#L39

VincentLanglet (approved):
I assume all the error reported by the Harden security step already exists for other workflow

PR #747, opened 2026-05-07
Remove Reflector from StubFilesExtensionLoader

xificurk:
Replace the expensive `reflectClass('...ServiceEntityRepository')` call with a cheap `InstalledVersions::getVersion('doctrine/doctrine-bundle')` + `version_compare` check. LazyServiceEntityRepository became ServiceEntityRepository's parent in DoctrineBundle 2.8.1 and was removed in 3.0.0, so the version threshold is exact.

Add StubFilesExtensionLoaderTest with ThrowingSourceLocator registered as `betterReflectionSourceLocator` in the test container — any future attempt to re-introduce reflectClass() in getFiles() will immediately throw.

--

Related:

https://github.com/phpstan/phpstan-src/pull/5577
https://github.com/phpstan/phpstan-src/pull/5538#issuecomment-4342781825

### DIFF
--- a/src/Stubs/Doctrine/StubFilesExtensionLoader.php
+++ b/src/Stubs/Doctrine/StubFilesExtensionLoader.php
@@ -4,24 +4,14 @@ namespace PHPStan\Stubs\Doctrine;
 
 use Composer\InstalledVersions;
 use OutOfBoundsException;
-use PHPStan\BetterReflection\Reflector\Exception\IdentifierNotFound;
-use PHPStan\BetterReflection\Reflector\Reflector;
 use PHPStan\PhpDoc\StubFilesExtension;
 use function class_exists;
 use function dirname;
 use function strpos;
+use function version_compare;
 
 class StubFilesExtensionLoader implements StubFilesExtension
 {
-
-	private Reflector $reflector;
-
-	public function __construct(
-		Reflector $reflector
-	)
-	{
-		$this->reflector = $reflector;
-	}
 
 	public function getFiles(): array
 	{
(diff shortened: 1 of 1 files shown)


src/Stubs/Doctrine/StubFilesExtensionLoader.php
```diff
@@ -73,6 +50,25 @@ public function getFiles(): array
...
+			$bundleVersion = InstalledVersions::getVersion('doctrine/doctrine-bundle');
+		} catch (OutOfBoundsException $e) {
+			return false;
+		}
+
+		if ($bundleVersion === null) {
+			return false;
+		}
+
+		return version_compare($bundleVersion, '2.8.1', '>=') && version_compare($bundleVersion, '3.0.0', '<');
```
VincentLanglet, line 69: Given the fact we're only testing doctrine/orm 16+ 
https://github.com/phpstan/phpstan-doctrine/blob/8af50e234613a72a4e90e2b09ecffe8e75b2a431/composer.json#L30

I was wondering about updating the conflict section
https://github.com/phpstan/phpstan-doctrine/blob/8af50e234613a72a4e90e2b09ecffe8e75b2a431/composer.json#L17

WDYT @staabm ?

This way this method would only become  `isInstalledVersion(doctrine/bundle, 2)`

staabm: > doctrine/orm 16+

you mean 2.16+ I guess?

> I was wondering about updating the conflict section

how would you change it? it seems the `"doctrine/orm": "<2.5", ` constraints exists for over 10 years in this extensiions

xificurk: Beware that this is a version check against `doctrine/doctrine-bundle`, not the core doctrine lib.

VincentLanglet: > > doctrine/orm 16+
> 
> you mean 2.16+ I guess?

Yes

> > I was wondering about updating the conflict section
> 
> how would you change it? it seems the `"doctrine/orm": "<2.5", ` constraints exists for over 10 years in this extensiions

The fact is that if we don't use `^2.5` in our require-dev section, we cannot be sure it works for this version.

So I feel like we should technically have
```
"conflict": {
		"doctrine/annotations": "<2.0",
		"doctrine/collections": "<1.6",
		"doctrine/common": "<2.7",
		"doctrine/dbal": "<3.3.8",
		"doctrine/lexer": "<2.0",
		"doctrine/mongodb-odm": "<2.4.3",
		"doctrine/orm": "<2.16",
		"doctrine/persistence": "<2.2.1"
	}
```
Or have test for older versions.... but maybe ondrej find this too strict.

VincentLanglet: > Beware that this is a version check against `doctrine/doctrine-bundle`, not the core doctrine lib.

Indeed @xificurk my bad this is not the same libs. Forget my comment.

BTW since doctrine-bundle is not in the require-dev section, I assume this dynamic stub is not tested at all. Not sure if we want too. (And that could be another PR topic)

staabm:
please remove the test again. we will cover it in https://github.com/phpstan/phpstan-src/pull/5617

xificurk:
Done, PR is fix only now.

staabm:
thank you

PR #756, opened 2026-05-22
Support query inference with multiple entity managers

### LINKED ISSUE #655: Repository return type errors when using multiple entity managers
php 8.3
phpstan 2

repository with reproduced issue: https://github.com/tedyno/phpstan-multiple-em-issue
phpstan ran with multiple object managers: https://github.com/tedyno/phpstan-multiple-em-issue/actions/runs/14328251812
phpstan an with single object manager: https://github.com/tedyno/phpstan-multiple-em-issue/actions/runs/14328261334

Hi, I'm having an issue to make phpstan work correctly when using multiple object managers. I followed a [comment](https://github.com/phpstan/phpstan-doctrine/issues/53#issuecomment-728037609) by @adrienbrault as I could not find any other way and there's a problem with recognizing correct return type.

```php
// src/Repository/UserRepository.php

    /**
     * @return User[]
     */
    public function get() : array
    {
        $query = $this->createQueryBuilder('c')
            ->getQuery();

        return $query->getResult();
    }
```

this simple query throws this when running phpstan
```bash
Method App\Repository\UserRepository::get() should return array<App\Entity\Main\User> but returns mixed.
```

When using object-manager.php from readme, the errors shows up only for EntityRepository of an Entity which is in the other manager indicating it resolved the return type correctly in UserRepository.

Could you advise how to proceed?

LubuSeb:
## Summary

Fixes return type inference for projects that use multiple Doctrine entity managers by allowing `objectManagerLoader` to return a Doctrine `ManagerRegistry`.

When the loader returns a registry, PHPStan Doctrine now resolves the object manager from the entity class found in the DQL before validating or inferring query result types. This keeps the existing single-manager loader behavior intact while supporting Symfony applications where each entity namespace is owned by a different manager.

## Details

- Add `ManagerRegistry` support to `ObjectMetadataResolver`.
- Resolve entity metadata and DQL/query-builder analysis through `getManagerForClass()` when possible.
- Add a regression fixture with two in-memory entity managers mapped to disjoint namespaces.
- Document returning the Doctrine registry from Symfony `objectManagerLoader` setups.

## Verification

- `vendor/bin/phpunit`
- Verified the repro from #655 by returning the Symfony Doctrine registry from `objectManagerLoader` and running PHPStan against `tedyno/phpstan-multiple-em-issue`; it reports no errors with this branch.

Fixes #655

### DIFF
--- a/src/Type/Doctrine/ObjectMetadataResolver.php
+++ b/src/Type/Doctrine/ObjectMetadataResolver.php
@@ -20,6 +20,7 @@ use function is_file;
 use function is_readable;
 use function method_exists;
 use function preg_match_all;
+use function preg_replace;
 use function reset;
 use function sprintf;
 use const PHP_VERSION_ID;
@@ -100,8 +101,13 @@ final class ObjectMetadataResolver
 			return $objectManagerLoaderResult;
 		}
 
-		preg_match_all('~\b(?:FROM|UPDATE)\s+([\\\\A-Za-z_][\\\\A-Za-z0-9_]*)(?:\s+|$)~i', $dql, $matches);
-		preg_match_all('~\bDELETE\s+(?:FROM\s+)?([\\\\A-Za-z_][\\\\A-Za-z0-9_]*)(?:\s+|$)~i', $dql, $deleteMatches);
+		$dqlWithoutStringLiterals = preg_replace("~'(?:''|[^'])*'~", "''", $dql);
+		if ($dqlWithoutStringLiterals === null) {
+			$dqlWithoutStringLiterals = $dql;
+		}
+
+		preg_match_all('~\b(?:FROM|UPDATE)\s+([\\\\A-Za-z_][\\\\A-Za-z0-9_]*)(?:\s+|$)~i', $dqlWithoutStringLiterals, $matches);
+		preg_match_all('~\bDELETE\s+(?:FROM\s+)?([\\\\A-Za-z_][\\\\A-Za-z0-9_]*)(?:\s+|$)~i', $dqlWithoutStringLiterals, $deleteMatches);
 		foreach (array_merge($matches[1], $deleteMatches[1]) as $className) {
 			if (!class_exists($className)) {
 				continue;
--- a/src/Type/Doctrine/ObjectMetadataResolver.php
+++ b/src/Type/Doctrine/ObjectMetadataResolver.php
@@ -14,10 +14,12 @@ use PHPStan\ShouldNotHappenException;
 use ReflectionException;
 use function array_merge;
 use function class_exists;
+use function count;
 use function is_file;
 use function is_readable;
 use function method_exists;
 use function preg_match_all;
+use function reset;
 use function sprintf;
 use const PHP_VERSION_ID;
 
@@ -68,7 +70,16 @@ final class ObjectMetadataResolver
 	public function getObjectManagerForClass(string $className): ?ObjectManager
 	{
 		$objectManagerLoaderResult = $this->getObjectManagerLoaderResult();
+		if ($objectManagerLoaderResult instanceof ObjectManager) {
+			return $objectManagerLoaderResult;
+		}
+
 		if ($objectManagerLoaderResult instanceof ManagerRegistry) {
+			$singleObjectManager = $this->getSingleObjectManager($objectManagerLoaderResult);
+			if ($singleObjectManager !== null) {
+				return $singleObjectManager;
+			}
+
 			$objectManager = $objectManagerLoaderResult->getManagerForClass($className);
 			if ($objectManager instanceof ObjectManager) {
 				return $objectManager;
@@ -76,6 +87,20 @@ final class ObjectMetadataResolver
 		}
 
 		return $this->getObjectManager();
+	}
+
+	public function getObjectManagerByName(string $name): ?ObjectManager
+	{
+		$objectManagerLoaderResult = $this->getObjectManagerLoaderResult();
+		if (!$objectManagerLoaderResult instanceof ManagerRegistry) {
+			return null;
+		}
+
+		try {
+			return $objectManagerLoaderResult->getManager($name);
+		} catch (\Throwable $e) {
+			return null;
+		}
 	}
 
 	public function getObjectManagerForDql(string $dql): ?ObjectManager
@@ -94,6 +119,17 @@ final class ObjectMetadataResolver
 		}
 
 		return $this->getObjectManager();
+	}
+
+	private function getSingleObjectManager(ManagerRegistry $managerRegistry): ?ObjectManager
+	{
+		$objectManagers = $managerRegistry->getManagers();
+		if (count($objectManagers) !== 1) {
+			return null;
+		}
+
+		$objectManager = reset($objectManagers);
+		return $objectManager;
 	}
 
 	/**
--- a/src/Type/Doctrine/ObjectMetadataResolver.php
+++ b/src/Type/Doctrine/ObjectMetadataResolver.php
@@ -54,15 +54,11 @@ final class ObjectMetadataResolver
 	public function getObjectManager(): ?ObjectManager
 	{
 		$objectManagerLoaderResult = $this->getObjectManagerLoaderResult();
-		if ($objectManagerLoaderResult instanceof ManagerRegistry) {
-			return $objectManagerLoaderResult->getManager();
-		}
-
-		if ($objectManagerLoaderResult instanceof ObjectManager) {
+		if (!$objectManagerLoaderResult instanceof ManagerRegistry) {
 			return $objectManagerLoaderResult;
 		}
 
-		return null;
+		return $objectManagerLoaderResult->getManager();
 	}
 
 	/**
@@ -71,20 +67,13 @@ final class ObjectMetadataResolver
 	public function getObjectManagerForClass(string $className): ?ObjectManager
 	{
 		$objectManagerLoaderResult = $this->getObjectManagerLoaderResult();
-		if ($objectManagerLoaderResult instanceof ObjectManager) {
+		if (!$objectManagerLoaderResult instanceof ManagerRegistry) {
 			return $objectManagerLoaderResult;
 		}
 
-		if ($objectManagerLoaderResult instanceof ManagerRegistry) {
-			$singleObjectManager = $this->getSingleObjectManager($objectManagerLoaderResult);
-			if ($singleObjectManager !== null) {
-				return $singleObjectManager;
-			}
-
-			$objectManager = $objectManagerLoaderResult->getManagerForClass($className);
-			if ($objectManager instanceof ObjectManager) {
-				return $objectManager;
-			}
+		$objectManager = $objectManagerLoaderResult->getManagerForClass($className);
+		if ($objectManager instanceof ObjectManager) {
+			return $objectManager;
 		}
 
 		return $this->getObjectManager();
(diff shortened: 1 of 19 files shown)


src/Type/Doctrine/ObjectMetadataResolver.php
```diff
@@ -47,23 +50,74 @@ public function hasObjectManagerLoader(): bool
...
+			$objectManager = $objectManagerLoaderResult->getManagerForClass($className);
+			if ($objectManager instanceof ObjectManager) {
+				return $objectManager;
+			}
+		}
+
+		return $this->getObjectManager();
+	}
+
+	public function getObjectManagerForDql(string $dql): ?ObjectManager
```
VincentLanglet, line 97: I see multiple issue with this method:
- I'm not sure it cover all the cases
- This complex logic shouldn't even be needed when when a single entity manager is used.

If the loader is an objectManager, you could return it directly.
We might even avoid this logic if a ManagerRegistry is provided but `getManagers` returns a single element array.

tests/Type/Doctrine/data/QueryResult/config-multiple-entity-managers.neon
```diff
@@ -0,0 +1,6 @@
+includes:
+	- ../../../../../extension.neon
+	- ../../../../../rules.neon
+parameters:
+	doctrine:
+		objectManagerLoader: entity-manager-multiple.php
```
VincentLanglet, line 6: I'm not a big fan of the fact now objectManagerLoader can either be an `objectManagerLoader` or a `objectManagerRegistryLoader`. Dunno if we should have
- a different option
- deprecating this option and always use objectManagerRegistryLoader
- or if it's fine for @ondrejmirtes

VincentLanglet:
Looking at https://github.com/phpstan/phpstan-doctrine/pull/757/changes#diff-e1bfea71addecdb2a3f73f91d95ed61b789458e5107d9efbbc39937d35b880fd some updates are also needed in `GetRepositoryDynamicReturnTypeExtension` no ?

tests/Type/Doctrine/data/QueryResult/config-multiple-entity-managers.neon
ondrejmirtes: I think it's fine.

LubuSeb:
Addressed the repository return-type path as well: the extension now reads the optional manager name argument and resolves the repository class through the matching object manager. I also fixed the lowest-dependency ObjectManager fixture compatibility issue and the coding-standard findings from the last run. CI is green on the latest head.

src/Type/Doctrine/ObjectMetadataResolver.php
```diff
@@ -47,23 +53,107 @@ public function hasObjectManagerLoader(): bool
...
+			$objectManager = $this->getObjectManagerForClass($className);
+			if ($objectManager !== null) {
+				return $objectManager;
+			}
+		}
+
+		return $this->getObjectManager();
+	}
+
+	private function getSingleObjectManager(ManagerRegistry $managerRegistry): ?ObjectManager
```
VincentLanglet: This method can be remove if we're doing the optimisation in the `getObjectManagerLoaderResult`.

Instead of 
```
$this->objectManagerLoaderResult = $this->loadObjectManager($this->objectManagerLoader);
```
you write
```
$result = $this->loadObjectManager($this->objectManagerLoader);
if ($result instanceof ManagerRegistry && count($result->getManager()) === 1) {
      $result = $result->getObjectManager();
}
$this->objectManagerLoaderResult = $result;
```

src/Type/Doctrine/ObjectMetadataResolver.php
```diff
@@ -47,23 +53,107 @@ public function hasObjectManagerLoader(): bool
...
+			return $objectManagerLoaderResult->getManager($name);
+		} catch (Throwable $e) {
+			return null;
+		}
+	}
+
+	public function getObjectManagerForDql(string $dql): ?ObjectManager
+	{
+		preg_match_all('~\b(?:FROM|UPDATE)\s+([\\\\A-Za-z_][\\\\A-Za-z0-9_]*)(?:\s+|$)~i', $dql, $matches);
+		preg_match_all('~\bDELETE\s+(?:FROM\s+)?([\\\\A-Za-z_][\\\\A-Za-z0-9_]*)(?:\s+|$)~i', $dql, $deleteMatches);
```
VincentLanglet: To avoid the preg_match_all/foreach logic. you should have the logic
```
$objectManagerResult = $this->getObjectManagerRsult();
if (!$objectManagerResult instanceof ManagerRegistry) {
      return $objectManagerResult;
}
```

src/Type/Doctrine/ObjectMetadataResolver.php
```diff
@@ -47,23 +53,107 @@ public function hasObjectManagerLoader(): bool
...
+		return null;
+	}
+
+	/**
+	 * @param class-string $className
+	 */
+	public function getObjectManagerForClass(string $className): ?ObjectManager
+	{
+		$objectManagerLoaderResult = $this->getObjectManagerLoaderResult();
+		if ($objectManagerLoaderResult instanceof ObjectManager) {
```
VincentLanglet: Could be directly
```
if (!$objectManagerLoaderResult instanceof ManagerRegistry) {
     return $objectManagerLoaderResult
}
```
which handle null and ObjectManager

LubuSeb:
Addressed the latest resolver simplification feedback in `3568a87`.

Changes made:
- `getObjectManagerLoaderResult()` now normalizes a `ManagerRegistry` with exactly one manager into that single `ObjectManager` when caching the loader result.
- `getObjectManager()` and `getObjectManagerForClass()` now return early when the cached loader result is not a registry.
- `getObjectManagerForDql()` now avoids the DQL class parsing path entirely for non-registry loader results.
- Removed the separate single-manager helper.

Validation:
- `parallel-lint` on `ObjectMetadataResolver.php` passed.
- Targeted PHPUnit for multiple-EM, named-manager repository, and ManagerRegistry inference passed: 8 tests / 8 assertions.
- Broader `tests/Type/Doctrine` passed: 279 tests / 8072 assertions.
- PHPStan analysis of `ObjectMetadataResolver.php` and `GetRepositoryDynamicReturnTypeExtension.php` passed.
- `phpcs` on `ObjectMetadataResolver.php` passed.
- GitHub CI is green on head `3568a87`.

src/Type/Doctrine/ObjectMetadataResolver.php
```diff
@@ -47,23 +53,99 @@ public function hasObjectManagerLoader(): bool
...
+
+	public function getObjectManagerForDql(string $dql): ?ObjectManager
+	{
+		$objectManagerLoaderResult = $this->getObjectManagerLoaderResult();
+		if (!$objectManagerLoaderResult instanceof ManagerRegistry) {
+			return $objectManagerLoaderResult;
+		}
+
+		preg_match_all('~\b(?:FROM|UPDATE)\s+([\\\\A-Za-z_][\\\\A-Za-z0-9_]*)(?:\s+|$)~i', $dql, $matches);
+		preg_match_all('~\bDELETE\s+(?:FROM\s+)?([\\\\A-Za-z_][\\\\A-Za-z0-9_]*)(?:\s+|$)~i', $dql, $deleteMatches);
```
VincentLanglet: Even if you don't use doctrine ; I saw you worked a lot with regex, so I'd like your opinion on this strategy @staabm

staabm: do I see this correctly, that we need this regex-magic, because we cannot get a AST for the DQL since we are in the process of creating a entity manager?

staabm: will the DQL always contain a full-qualified class-name  in the query?

staabm: I don't see a test which contains a DELETE statement.. is this line tested?

LubuSeb: Yes, that's the reason for the pre-parse step. At this point we need a candidate entity class before handing the DQL to Doctrine's EM-bound parser/metadata validation; otherwise the default manager can reject a tenant entity before we know which manager should parse it. The regex is intentionally a narrow heuristic to find an autoloadable root entity class, then the actual manager choice is delegated to `ManagerRegistry::getManagerForClass()`.

LubuSeb: No, not universally. This only supports the autoloadable class-string path. QueryBuilder/repository-generated DQL uses class metadata names, and direct `createQuery()` inference can resolve those FQCN-style class strings. If the DQL uses a short alias or anything not autoloadable, this falls back to the default configured manager rather than guessing.

LubuSeb: Good catch. I added focused coverage in `badaaae` using Doctrine's no-`FROM` form, `DELETE QueryResult\MultipleEntityManagers\Tenant\App a`, so it exercises the DELETE-specific extraction branch rather than the generic `FROM` branch.

Verification run in Docker with PHP 8.4:
- `vendor/bin/phpunit tests/Type/Doctrine/MultipleEntityManagersQueryTypeInferenceTest.php` -> 9 tests / 9 assertions
- `vendor/bin/phpunit tests/Type/Doctrine` -> 280 tests / 8073 assertions
- `git diff --check` -> clean

staabm: @stof do you forsee whether this regex-matching is "good enough" for most DQL to identify the used entity-class?
(I am no DQL expert and don't know which syntax people use/is supported)

in case we can think about more DQL syntax edge cases, this might help to add more tests and harden the regex pattern

LubuSeb:
Quick check-in on this. The latest head 0194cef expands the DQL coverage for multiple entity managers, and all CI checks are passing. I do not see any remaining unresolved review request on my side, but I am happy to adjust anything else if there is still a blocker before this can move forward.

VincentLanglet:
> Quick check-in on this. The latest head [0194cef](https://github.com/phpstan/phpstan-doctrine/commit/0194cef20457d69816f242613f50b6bbf25bcd46) expands the DQL coverage for multiple entity managers, and all CI checks are passing. I do not see any remaining unresolved review request on my side, but I am happy to adjust anything else if there is still a blocker before this can move forward.

I was waiting the answer from stof since staabm asked him in this comment
https://github.com/phpstan/phpstan-doctrine/pull/756#discussion_r3302974262

staabm:
//cc @stof

LubuSeb:
Pushed `309df6d` to address the DQL pre-parse concern from the review thread.

Changes:
- strip single-quoted DQL string literals before the lightweight entity-class scan, so class-looking text inside string literals cannot select the wrong manager;
- add resolver-level coverage for `AS` aliases, lowercase keywords, partial object selects, root entity before subquery, update/delete DQL, and escaped string-literal false positives;
- keep the existing FQCN-only behavior: unsupported entity-name/alias forms still fall back to the default manager rather than guessing.

Verification run locally on PHP 8.4:
- `vendor/bin/phpunit tests/Type/Doctrine/ObjectMetadataResolverMultipleEntityManagersTest.php`
- `vendor/bin/phpunit tests/Type/Doctrine/MultipleEntityManagersQueryTypeInferenceTest.php`
- `vendor/bin/phpunit tests/Type/Doctrine` -> 290 tests / 8083 assertions
- `vendor/bin/parallel-lint` on the modified files
- `vendor/bin/phpstan analyse` on the modified files

LubuSeb:
Pushed `ea64c8d` to make the new resolver fixture compatible with the lowest Doctrine Persistence versions used in CI. The fake `ObjectManager` now follows the older untyped interface style and includes the legacy `merge()` path, matching the compatibility pattern used by existing fixtures.

Local re-check:
- `vendor/bin/phpunit tests/Type/Doctrine/ObjectMetadataResolverMultipleEntityManagersTest.php`
- `vendor/bin/phpunit tests/Type/Doctrine` -> 290 tests / 8083 assertions
- `vendor/bin/parallel-lint` on modified files
- `vendor/bin/phpstan analyse` on the production resolver and new test class
- `git diff --check`

staabm (approved):

VincentLanglet (approved):

staabm:
lets move on. thanks @LubuSeb

LubuSeb:
Hey @peldax, just checking in again about the $300 bounty from #655. You mentioned it was going through the company process and that GitHub Sponsors would work. Is it still moving along, or do you need anything from me? Thanks!

LubuSeb:
hey @peldax, checking in again on the $300 bounty for #655. the fix has been merged and released since july 13, and you mentioned that github sponsors would work. could you give me an expected payout date or let me know if anything is still needed from me? thanks

PR #757, opened 2026-05-22
fix: infer repositories from named managers

ItsOtherMauridian:
## Summary

- supports `ManagerRegistry` object-manager loaders when resolving repository metadata
- uses the constant second argument of `ManagerRegistry::getRepository($class, $managerName)` to resolve metadata from the named manager
- adds a regression test where the same entity class has different custom repositories on the default and tenant managers

## Why

When an application has multiple managers that know about the same entity class, the repository type can depend on the explicit manager name passed to `ManagerRegistry::getRepository()`. Without considering that argument, PHPStan can infer the repository class from the default manager even when the call explicitly requests another manager.

## Local validation

Executed in Docker with Composer dependencies installed via:

```text
composer install --no-interaction --no-progress --prefer-dist --ignore-platform-req=ext-mongodb
```

Validation commands:

```text
php -l src/Type/Doctrine/ObjectMetadataResolver.php
php -l src/Type/Doctrine/GetRepositoryDynamicReturnTypeExtension.php
php -l tests/Type/Doctrine/NamedManagerRepositoryTypeInferenceTest.php
php -l tests/Type/Doctrine/data/namedManagerRepository.php
php -l tests/Type/Doctrine/data/named-manager-registry.php
php vendor/bin/phpunit --filter NamedManagerRepositoryTypeInferenceTest
php vendor/bin/phpunit tests/Type/Doctrine
git diff --check
```

Observed results:

```text
OK (2 tests, 2 assertions)
OK (271 tests, 8064 assertions)
```

## Disclosure

This PR was prepared with AI assistance and manually checked before submission.

### DIFF
--- a/src/Type/Doctrine/GetRepositoryDynamicReturnTypeExtension.php
+++ b/src/Type/Doctrine/GetRepositoryDynamicReturnTypeExtension.php
@@ -104,9 +104,10 @@ class GetRepositoryDynamicReturnTypeExtension implements DynamicMethodReturnType
 		}
 
 		$repositoryTypes = [];
+		$managerName = $this->getManagerName($scope, $methodCall->getArgs());
 		foreach ($objectNames as $objectName) {
 			try {
-				$repositoryClass = $this->getRepositoryClass($objectName, $defaultRepositoryClass);
+				$repositoryClass = $this->getRepositoryClass($objectName, $defaultRepositoryClass, $managerName);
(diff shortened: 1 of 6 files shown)


VincentLanglet:
Hi @ItsOtherMauridian, this feels kinda similar to the already opened by @LubuSeb PR
https://github.com/phpstan/phpstan-doctrine/pull/756 (with less place where the ManagerRegistry is used)

The main discussion https://github.com/phpstan/phpstan-doctrine/pull/756#discussion_r3293051278 still stands and need a decision first.

PR #762, opened 2026-05-28
Update ServiceEntityRepository stub

zacharylund:
Add new methods in doctrine/doctrine-bundle v3.

In doctrine/doctrine-bundle 2.x, PHPStan reports that Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository::findBy() returns list<Entity> but in 3.x, PHPStan reports that it returns array<int, Entity>.

### DIFF
--- a/stubs/ServiceEntityRepository.stub
+++ b/stubs/ServiceEntityRepository.stub
@@ -1,11 +1,73 @@
 <?php
 namespace Doctrine\Bundle\DoctrineBundle\Repository;
 
-use \Doctrine\ORM\EntityRepository;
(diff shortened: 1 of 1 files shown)


VincentLanglet:
> In doctrine/doctrine-bundle 2.x, PHPStan reports that Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository::findBy() returns list but in 3.x, PHPStan reports that it returns array<int, Entity>.

Does https://github.com/phpstan/phpstan-doctrine/pull/767 solve your issue ?

zacharylund:
Yes, and that appears to be a better solution. Thanks!

PR #767, opened 2026-06-10
Sync ObjectRepository stub types with Doctrine

herndlm:
Refs: https://github.com/phpstan/phpstan/discussions/14802

See https://github.com/doctrine/persistence/blob/4.2.x/src/ObjectRepository.php

no clue if this file is even being needed still, maybe for older versions of Doctrine?

looks like they changed it to a list just recently, see https://github.com/doctrine/persistence/commit/d2ab8e3e36f2989c9ee9e9ec0a916e84551c5cd6

there's also open https://github.com/phpstan/phpstan-doctrine/pull/762 which might be trying to fix the same thing

### DIFF
--- a/stubs/DocumentRepository.stub
+++ b/stubs/DocumentRepository.stub
@@ -21,17 +21,18 @@ class DocumentRepository implements ObjectRepository
 	public function find($id, $lockMode = null, $lockVersion = null);
 
 	/**
-	 * @phpstan-return array<int, TDocumentClass>
+	 * @phpstan-return list<TDocumentClass>
 	 * @phpstan-impure
 	 */
 	public function findAll();
 
 	/**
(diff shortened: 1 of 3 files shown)


herndlm:
saw those param changes and copied them over too. should I revert them to avoid potential annoying new reported issues popping up? 🤔

VincentLanglet (approved):

VincentLanglet:
> saw those param changes and copied them over too. should I revert them to avoid potential annoying new reported issues popping up? 🤔

I think it's ok to have the same than original class, like you did

staabm (approved):

herndlm:
hi guys, thx for merging this so fast, but the criteria type change was a bug/mistake, I get a huge amount of false positives now. let me quickly fix that.. sorry

PR #769, opened 2026-06-12
Avoid reporting always true switch

VincentLanglet:
This is as-safe as the previous way but without the always true case error I think


ondrejmirtes:
We shouldn't need to change this.

PR #776, opened 2026-07-15
Add a check for string-backed enum maximum length 

jwpage:
This adds a check for entity columns are mapped to a string-backed `enumType` to ensure that the maximum length of the enum values does not exceed the defined database column length.

(An additional future `enumType` check could be to ensure that integer-backed `enumType` columns with negative numbers are not mapped to an unsigned integer, but I will leave that for a different PR.)


jwpage:
Needs further consideration

PR #777, opened 2026-07-22
perf: Check QueryBuilder method names before class hierarchies

realFlowControl:
Reorder the `EntityRepository` and `EntityManager` `createQueryBuilder` guards so the cheap method-name comparison runs before `ClassReflection::is()`.

The resolver sees hundreds of thousands of calls, while these special cases apply only to `createQueryBuilder`. Avoiding hierarchy checks for every other method removes unnecessary reflection work without changing the accepted calls.

Benchmark: uncached Sylius analysis of 2,366 files with eight workers, five runs per variant. Median elapsed time fell from 28.91 s to 28.24 s (-2.3%) and aggregate user+sys CPU from 176.22 s to 173.44 s (-1.6%).

Benchmark run locally using PHPStan `2.2.x` and PHPStan-Doctrine `2.0.x` on MacOS with PHP 8.5.7


staabm (approved):

staabm:
thank you!

PR #778, opened 2026-07-22
perf: Cache QueryBuilder class classification

realFlowControl:
Cache whether each stable receiver class name is a `QueryBuilder` instead of repeating `ClassReflection::is()` for every call on that class.

Receiver classification is argument- and scope-independent, and the measured resolver performs more than 325,000 receiver reflection passes. Caching only this boolean avoids repeated hierarchy checks without caching scope-dependent method resolution.

This reduced median elapsed time from 28.45 s to 28.22 s (-0.8%) and aggregate user+sys CPU from 168.90 s to 167.37 s (-0.9%).

Benchmark run locally using PHPStan 2.2.x and PHPStan-Doctrine 2.0.x on MacOS with PHP 8.5.7

### DIFF
--- a/src/Type/Doctrine/QueryBuilder/ReturnQueryBuilderExpressionTypeResolverExtension.php
+++ b/src/Type/Doctrine/QueryBuilder/ReturnQueryBuilderExpressionTypeResolverExtension.php
@@ -25,6 +25,9 @@ class ReturnQueryBuilderExpressionTypeResolverExtension implements ExpressionTyp
 
 	private OtherMethodQueryBuilderParser $otherMethodQueryBuilderParser;
 
+	/** @var array<string, bool> */
+	private array $queryBuilderClasses = [];
+
 	public function __construct(
 		OtherMethodQueryBuilderParser $otherMethodQueryBuilderParser
 	)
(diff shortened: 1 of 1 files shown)


src/Type/Doctrine/QueryBuilder/ReturnQueryBuilderExpressionTypeResolverExtension.php
```diff
@@ -85,7 +88,8 @@ private function getMethodReflection(CallLike $call, Scope $scope): ?MethodRefle
 		$methodName = $call->name->name;
 
 		foreach ($callerType->getObjectClassReflections() as $callerClassReflection) {
-			if ($callerClassReflection->is(QueryBuilder::class)) {
+			$className = $callerClassReflection->getName();
+			if (($this->queryBuilderClasses[$className] ??= $callerClassReflection->is(QueryBuilder::class))) {
 				return null; // covered by QueryBuilderMethodDynamicReturnTypeExtension
```
staabm, line 93: OT: this comment made me curious. looking into `QueryBuilderMethodDynamicReturnTypeExtension` and 
made me realize that we have a config arround to declare what the `QueryBuilder` class is, since https://github.com/phpstan/phpstan-doctrine/commit/4d630c6755ff0b67d202f62571e21311831f0207

staabm (commented):
while the change is tiny, I wonder whether such a small benefit is worth caching?

how much overhead does this extension cost the static analysis run on your project?
maybe its worth building a list of method names using a collector, so we could prevent the `is` call alltogehter for most calls based on method name (whether thats worth the effort depends on hoch much impact this could have)?

VincentLanglet:
> while the change is tiny, I wonder whether such a small benefit is worth caching?

I agree, I dunno what's the strategy about such perf improvement

VincentLanglet (approved):

staabm:
needs rebase/conflict resolution. are the improvements still visible on the latest phpstan release?

(please remove merge commits)

PR #779, opened 2026-07-22
perf: Reuse the QueryBuilder ObjectType

realFlowControl:
Create the `QueryBuilder` `ObjectType` once in the resolver constructor and reuse it for return-type compatibility checks.

The resolver is called more than 600,000 times in the measured workload. The `ObjectType` is immutable for this use, so allocating the same value for each method candidate only adds allocation and CPU overhead.

Benchmark: uncached Sylius analysis of 2,366 files with eight workers, five runs per variant. Median elapsed time fell from 28.91 s to 28.34 s (-2.0%) and aggregate user+sys CPU from 176.22 s to 169.68 s (-3.7%).

Benchmark run locally using PHPStan 2.2.x and PHPStan-Doctrine 2.0.x on MacOS with PHP 8.5.7

### DIFF
--- a/src/Type/Doctrine/QueryBuilder/ReturnQueryBuilderExpressionTypeResolverExtension.php
+++ b/src/Type/Doctrine/QueryBuilder/ReturnQueryBuilderExpressionTypeResolverExtension.php
@@ -25,11 +25,14 @@ class ReturnQueryBuilderExpressionTypeResolverExtension implements ExpressionTyp
 
 	private OtherMethodQueryBuilderParser $otherMethodQueryBuilderParser;
 
+	private ObjectType $queryBuilderObjectType;
+
 	public function __construct(
 		OtherMethodQueryBuilderParser $otherMethodQueryBuilderParser
 	)
 	{
 		$this->otherMethodQueryBuilderParser = $otherMethodQueryBuilderParser;
+		$this->queryBuilderObjectType = new ObjectType(QueryBuilder::class);
 	}
 
 	public function getType(Expr $expr, Scope $scope): ?Type
@@ -50,7 +53,7 @@ class ReturnQueryBuilderExpressionTypeResolverExtension implements ExpressionTyp
 
 		$returnType = ParametersAcceptorSelector::selectFromArgs($scope, $expr->getArgs(), $methodReflection->getVariants())->getReturnType();
 
-		$returnsQueryBuilder = (new ObjectType(QueryBuilder::class))->isSuperTypeOf($returnType)->yes();
+		$returnsQueryBuilder = $this->queryBuilderObjectType->isSuperTypeOf($returnType)->yes();
 
 		if (!$returnsQueryBuilder) {
 			return null;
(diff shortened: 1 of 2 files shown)


realFlowControl:
I do not fully understand why infection still fails, something is off

staabm:
`Trinary->yes()` vs. `!Trinary->no()` mutations get killed by implementing a test which instructs the `Trinary->maybe()` case. I think in this case this means you need to test a union with multiple classes.

we don't necessarily need a green mutation testing build on each PR though

realFlowControl:
> we don't necessarily need a green mutation testing build on each PR though

which would involve adding this mutation to a baseline file? Otherwise I assume this would just make `phpstan:2.0.x` fail

staabm (approved):
I think its fine

src/Type/Doctrine/QueryBuilder/ReturnQueryBuilderExpressionTypeResolverExtension.php
```diff
@@ -25,11 +25,14 @@ class ReturnQueryBuilderExpressionTypeResolverExtension implements ExpressionTyp
...
 	private OtherMethodQueryBuilderParser $otherMethodQueryBuilderParser;
 
+	private ObjectType $queryBuilderObjectType;
+
 	public function __construct(
 		OtherMethodQueryBuilderParser $otherMethodQueryBuilderParser
 	)
 	{
 		$this->otherMethodQueryBuilderParser = $otherMethodQueryBuilderParser;
+		$this->queryBuilderObjectType = new ObjectType(QueryBuilder::class);
```
VincentLanglet, line 35: I never worked a lot of perf issue so I rely on you @staabm.

1) Wouldn't it be better to move this from the constructor and use
`$this->queryBuilder ??= new ObjectType(QueryBuilder::class)` at the place we need it ?

2) If we start caching ObjectType, shouldn't we start by the PHPStan repository ? I'm pretty sure we have lot of reused ObjectType and we might want to expose a dedicated factory/method which cache the instance for every instance.

In the same way we might want to cache also so ConstantType...

staabm: > 1. `$this->queryBuilder ??= new ObjectType(QueryBuilder::class)` at the place we need it ?

I had the same gut feeling, but than found other classes in which we do it in `__construct`

> 2\. If we start caching ObjectType, shouldn't we start by the PHPStan repository ? I'm pretty sure we have lot of reused ObjectType and we might want to expose a dedicated factory/method which cache the instance for every instance.

I think the caching we do here is only necessary when its invoked very very often. as long as we don't have evidence, I would not add another level of indirection

> In the same way we might want to cache also so ConstantType...

not sure what you mean exactly. maybe I missunderstood the above

VincentLanglet (approved):

staabm:
thank you @realFlowControl

PR #789, opened 2026-08-26
Ask the node-callback scope directly instead of toMutatingScope()

SanderMuller:
phpstan-src deprecated `Scope::toMutatingScope()` (`4cc5e5e51`, kept as an alias in `a7a9bb794`), so `make phpstan` here fails on the `extension-tests / phpstan-doctrine` lane of **every** phpstan-src pull request:

```
Call to deprecated method toMutatingScope() of interface PHPStan\Analyser\Scope
 [ERROR] Found 1 error
```

The deprecation message says: *"The scope answers every ask directly - call the methods on the scope itself, or `toWalkScope()` for the engine-facing walk scope."* Since `OtherMethodQueryBuilderParser` only wants the type of a `return` expression inside a node callback, asking the scope directly is the whole change:

```diff
-$exprType = $scope->toMutatingScope()->getType($node->expr);
+$exprType = $scope->getType($node->expr);
```

The advantage over `toWalkScope()` is that this works on the PHPStan already required here, so `"phpstan/phpstan": "^2.2.2"` stays as it is and nothing has to wait for a release. `toWalkScope()` does not exist in any release yet - 2.2.9 predates the rename - so that route needs `^2.2.10` first, as @ondrejmirtes noted in phpstan/phpstan#15110.

## Verification (against the installed 2.2.9)

- The line is covered: instrumenting it shows **11 hits** across the test suite.
- **489 tests, 8334 assertions, 26 skipped** - identical before and after.
- `vendor/bin/phpstan analyse -c phpstan.neon`: `[OK] No errors`.

One caveat I could not close locally: the extension-tests lane runs against a compiled phar of phpstan-src, which I cannot build here, so I have not executed this against dev phpstan-src - only reasoned from the deprecation's own guidance plus the covered tests on 2.2.9. If you would rather take the route @ondrejmirtes suggested (`^2.2.10` plus `toWalkScope()`), say so and I will switch it.

Refs phpstan/phpstan#15110

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
 	"keywords": ["static analysis"],
 	"require": {
 		"php": "^7.4 || ^8.0",
-		"phpstan/phpstan": "^2.2.2"
+		"phpstan/phpstan": "^2.2.10"
 	},
 	"conflict": {
 		"doctrine/collections": "<1.0",
--- a/src/Type/Doctrine/QueryBuilder/OtherMethodQueryBuilderParser.php
+++ b/src/Type/Doctrine/QueryBuilder/OtherMethodQueryBuilderParser.php
@@ -98,7 +98,7 @@ class OtherMethodQueryBuilderParser
 				return;
 			}
 
-			$exprType = $scope->toMutatingScope()->getType($node->expr);
+			$exprType = $scope->getType($node->expr);
 
 			TypeTraverser::map($exprType, static function (Type $type, callable $traverse) use (&$queryBuilderTypes): Type {
 				if ($type instanceof UnionType || $type instanceof IntersectionType) {


ondrejmirtes:
You have to require ^2.2.10 at the same time 😊

SanderMuller:
Done - `^2.2.10` is in the same commit now. Understood on the reason: `getType()` on the callback scope only answers from the walk's stored results from 2.2.10 on, so my green suite on 2.2.9 proved the shape compiles, not that the answer is right there.

Note the install will not resolve until 2.2.10 is tagged, so CI here stays red until then. `composer.lock` is not tracked, so nothing else needs updating.

ondrejmirtes:
Thank you!

